PR #69, opened 2026-05-07
feat: Add LangChain middleware integration

dagardner-nv:
#### Overview
* Adds a LangChain integration package located in `python/nemo_flow/integrations/langchain`
* Adds a `nemo_flow.utils.run_sync` method, I figured other integrations will likely need this.
* Defined `nemo_flow` as known-first party in the ruff settings, this triggered several formatting changes beyond the scope of this PR
* Work-around for wasm-pack GitHub repo migration causing the package to fail to be installed with `taiki-e/install-action`

- Closes #

## Summary by CodeRabbit

* **New Features**
  * LangChain integration: middleware and callbacks to route model and tool calls through agents
  * Sync-run utility to execute async coroutines from sync code while preserving context

* **Documentation**
  * Integration guide and usage examples added; integrations overview restructured

* **Tests**
  * New unit and integration tests for LangChain integration and the sync-run utility

* **Chores**
  * CI updated to run Python integration tests, cache integration deps, build integration wheels, and ignore editor/third-party paths


dagardner-nv:
@coderabbitai review

integrations/langchain/src/langchain_nemo_flow/middleware.py
```diff
@@ -0,0 +1,183 @@
...
+        return await nemo_flow.llm.stream_execute(
+            model_name,
+            request,
+            func,
+            collector,
+            finalizer,
+            model_name=model_name,
+            codec=codec,
+            response_codec=response_codec,
+        )
```
willkill07: Should these be private?

integrations/langchain/hatch_fetch_version.py
willkill07, line 1: Do we need this? when we _do_ publish we always have a git tag.

integrations/langchain/uv.lock
willkill07, line 1: I am debating if we want to use extras to enable integration support -- e.g.

`nemo-flow[langchain]` rather than a separate package. Multiple `uv.lock` files aren't really fasible.

Then we could have a cononical `nemo_flow.integrations.<vendor>` 🤷🏻

I'm not sure what the best approach is here.

python/nemo_flow/utils.py
```diff
@@ -0,0 +1,44 @@
...
+    ctx = contextvars.copy_context()
+    try:
+        scope_stack = nemo_flow.get_scope_stack()
+
+        def _run_with_scope_stack() -> Any:
+            nemo_flow.set_thread_scope_stack(scope_stack)
+            return asyncio.run(coro)
+
+        with ThreadPoolExecutor(max_workers=1) as pool:
+            return pool.submit(ctx.run, _run_with_scope_stack).result()
```
willkill07: Is there any way to hoist this such that we do not need to recreate a ThreadPoolExecutor with each `run_sync` ?

integrations/langchain/pyproject.toml
```diff
@@ -0,0 +1,74 @@
...
+]
+
+dependencies = [
+    # Minimum version to support they hooks we are using
+    "langchain>=1.2.9,<2.0.0",
+    # Allow the langchain to dictate the version of langchain-core, not strictly needed
+    # but we do import from langchain_core
+    "langchain-core",
+    "nemo-flow>=0.2.0,<0.3.0",
+]
```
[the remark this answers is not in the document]
dagardner-nv: Huh? This isn't an optional dependency, it is a direct dependency and is in the `dependencies` array on line 34 of `integrations/langchain/pyproject.toml`

integrations/langchain/uv.lock
dagardner-nv: The big advantage is that we could run a single `pytest` process to cover both the python extension and the integrations.

The down-side would be if these integrations end up introducing conflicting dependencies.

willkill07 (changes requested):
Thanks for the PR. The overall direction is good: this uses LangChain’s public `AgentMiddleware` boundaries instead of patching LangChain for the new package, it keeps the integration observability-focused, and the callback tests do a nice job covering basic scope lifecycle and parent/child scope propagation.

I do think a few changes are needed before merge:

1. **Use the root lockfile / uv workspace model for the integration dependencies.** This PR adds `integrations/langchain/uv.lock` and updates CI to cache both that file and the root `uv.lock`. For this repo, please keep dependency resolution centralized: add `integrations/langchain` as a uv workspace member, make the package resolve through the root workspace/root `uv.lock`, remove the nested lockfile, and update the `just`/CI recipes accordingly. uv’s workspace docs call out this exact model: each member can have its own `pyproject.toml`, while the workspace shares a single lockfile. Reference: https://docs.astral.sh/uv/concepts/projects/workspaces/

2. **Fix `run_sync` before relying on it from the middleware.** `uv run pytest python/tests/test_utils.py` fails locally with `RuntimeError: cannot reuse already awaited coroutine`. The broad `except Exception: pass` around the scope-stack path can swallow a failure after the coroutine has already been submitted/awaited, then the fallback resubmits the same consumed coroutine. Please restructure this so only expected scope-stack lookup/setup failures fall back before the coroutine is executed, and worker execution failures propagate normally. The fallback should also run under the captured context rather than dropping `contextvars`.

3. **Add focused coverage for the tool-call wrappers and the context/scope bridge.** The integration implements sync and async model wrappers plus sync and async tool wrappers, but tests currently cover only model wrapping and callback lifecycle. LangChain documents `wrap_model_call` and `wrap_tool_call` as the wrap-style boundaries around model and tool calls, so the tool path should be tested too: args rewrite, `ToolMessage`/`Command` passthrough, async path, and behavior when NeMo Flow changes tool args. Reference: https://docs.langchain.com/oss/python/langchain/middleware/custom

4. **Keep the PR scope tight.** `_llm_stream_execute` is currently unused and there is no streaming hook/test path for it, so I would remove it until it is wired through real public LangChain behavior.

5. **Small hygiene items worth fixing while you are here.** Make the internal `langchain_run_id` metadata key win over caller-supplied metadata, make the serialized-id fallback safe for an empty `id` list, fix the nonstandard `# ty: ignore[...]` comments in tests

Validation I ran:

- `just test-python-integration` passes: 14 passed, 1 upstream LangGraph deprecation warning.
- `uv run pytest python/tests/test_utils.py` fails with the `run_sync` coroutine reuse issue above.
- `uv run ruff check python/nemo_flow/utils.py python/tests/test_utils.py integrations/langchain/src/langchain_nemo_flow integrations/langchain/tests` passes.
- `uv lock --check` passes in both the root and `integrations/langchain`, but the nested lockfile structure is still the concern.
- `git diff --check c02d24cc31c26b656c5934daa9596a69124eeb8b...HEAD` fails due trailing whitespace in the LangChain patch file.

I also cross-checked the LangChain public middleware references: the use of `wrap_model_call`/`awrap_model_call` and `wrap_tool_call`/`awrap_tool_call` is the right integration boundary for this style of observability middleware; the main remaining issues are dependency/workspace shape, the sync bridge bug, and missing coverage for the tool path.

integrations/langchain/tests/test_middleware.py
```diff
@@ -0,0 +1,121 @@
...
+"""Tests for the LangChain NeMo Flow middleware."""
+
+from __future__ import annotations
+
+import asyncio
+from typing import Any
+
+import nemo_flow
+import pytest
+from langchain.agents.middleware import ModelRequest, ModelResponse
```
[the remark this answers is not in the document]
dagardner-nv: Thanks this is a solid catch

.github/ci-path-filters.yml
willkill07, line 1: This change can be reverted

.github/workflows/ci_python.yml
willkill07, line 1: This change can be reverted

.gitignore
```diff
@@ -74,6 +75,7 @@ benchmarks/
 # Local third-party upstream checkouts. These are bootstrapped from
 # third_party/sources.lock and are intentionally excluded from the tracked
 # first-party repository surface.
+third_party/deepagents/
```
willkill07: Do we still add deepagents?

AGENTS.md
```diff
@@ -38,6 +38,8 @@ python/
 go/
   nemo_flow/  # Experimental Go CGo binding and tests
 docs/         # Sphinx documentation site
+integrations/ # Integrations using public APIs without patching
+integrations/**/tests/  # Integration test suites for public API-based integrations
```
willkill07: No longer true

pyproject.toml
```diff
@@ -28,7 +28,6 @@ keywords = [
 classifiers = [
     "Development Status :: 4 - Beta",
     "Intended Audience :: Developers",
-    "License :: OSI Approved :: Apache Software License",
```
willkill07, line 31: Does this classifier now automatically get added?

dagardner-nv: They've been deprecated in [PEP 639](https://peps.python.org/pep-0639/) replaced by the `license = "Apache-2.0"` setting on line 14

justfile
```diff
@@ -742,7 +743,7 @@ test-python:
     #!/usr/bin/env bash
     {{ bash_helpers }}
     output_dir="{{ output_dir }}"
-    pytest_cmd=(pytest python/tests)
+    pytest_cmd=(pytest)
```
dagardner-nv, line 746: Path not needed, test paths are defined in `pyproject.toml`

pyproject.toml
```diff
@@ -78,6 +73,28 @@ docs = [
...
+[project.optional-dependencies]
+langchain = [
+    "langchain>=1.2.11,<2.0.0",
+    "langchain-core",
+]
+
+langchain-nvidia = [
+    "langchain>=1.2.11,<2.0.0",
+    "langchain-core",
+    "langchain-nvidia-ai-endpoints",
```
willkill07: If we also check ChatNVIDIADynamo, then we'd need a minimum version specified.

willkill07 (dismissed):

Salonijain27 (dismissed):

willkill07 (dismissed):

willkill07 (dismissed):

willkill07 (approved):

Salonijain27 (approved):

PR #89, opened 2026-05-12
feat: add core observability plugin

willkill07:
#### Overview

Adds a built-in core `observability` plugin that configures first-party ATOF, ATIF, OpenTelemetry, and OpenInference exporters through the existing plugin lifecycle. The plugin is auto-registered by core plugin APIs and keeps every exporter section disabled unless that section explicitly sets `enabled: true`.

#### Details

- Adds `nemo_flow::observability::plugin_component` with public config structs, singleton component registration, inferred subscriber names, ATOF/ATIF/OTLP wiring, and ATIF per-top-level-agent file splitting.
- Adds FFI, Python, Node.js, Go, and WebAssembly helper surfaces for serializing observability plugin config.
- Adds coverage-focused tests for defaults, disabled sections, invalid values, lifecycle teardown, file outputs, OTLP registration, and multi-agent ATIF isolation across native bindings.
- Updates observability docs, README/API reference entry points, and Node docs generation to include the new plugin and configuration examples.

Validation run locally:

- `cargo test -p nemo-flow observability::plugin_component -- --test-threads=1`
- `just test-rust`
- `just test-python`
- `just test-node`
- `just test-go`
- `just test-wasm`
- `just docs`
- `uv run pytest python/tests/test_observability_plugin.py`
- `node --test --test-force-exit tests/observability_plugin_tests.mjs`
- `go test -v -run 'Observability.*Atif|ObservabilityPluginAtofAndAtif|ObservabilityPluginAtif|ObservabilityAtif' ./...`
- `cargo test -p nemo-flow-ffi test_ffi_observability_plugin -- --test-threads=1`

Start with `crates/core/src/observability/plugin_component.rs`, then review `crates/core/tests/unit/observability/plugin_component_tests.rs` and the binding-specific observability plugin tests.

- Relates to: none

## Summary by CodeRabbit

* **New Features**
  * Built-in "observability" plugin to configure ATOF, ATIF, OpenTelemetry and OpenInference exporters from a single component; auto-registered and discoverable.
  * Cross-language helper APIs and defaults for constructing observability configs and component specs (Rust, Python, Node, Wasm, Go, FFI).

* **Documentation**
  * New observability plugin guide, examples, and updated API/docs across language bindings and packaging.

* **Tests**
  * Extensive unit and integration tests covering validation, file outputs, registration, and teardown.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NeMo-Flow/pull/89)


crates/core/src/observability/plugin_component.rs
```diff
@@ -0,0 +1,1068 @@
...
+            return;
+        }
+        let Some(agent) = self.agents.get_mut(&agent_uuid) else {
+            return;
+        };
+        (agent.exporter.subscriber())(event);
+        agent.observed_events.push(event.clone());
+        if event.uuid() == agent_uuid
+            && event.scope_category() == Some(ScopeCategory::End)
+            && let Err(err) = write_atif_file(agent)
```
mnajafian-nv: This write happens while the AtifDispatcher mutex is held, and write_atif_file does create_dir_all plus fs::write. A slow filesystem, full disk, or network mount can block event delivery for every active agent sharing this dispatcher. Can we collect the write-ready trajectory/path under the lock, drop the guard, and perform filesystem I/O outside the mutex?

mnajafian-nv: After a top-level agent end event is written, the dispatcher still keeps the ManagedAtifExporter and observed_events in self.agents, plus the scope subscriber name in self.scope_subscribers. In a long-running process with many agent scopes, this retains every completed trajectory until plugin.clear(). Can we remove completed agent state after a successful write and add a unit test that closed agents do not remain in dispatcher state?

crates/core/src/observability/plugin_component.rs
```diff
@@ -0,0 +1,1068 @@
...
+    config: AtifSectionConfig,
+    agents: HashMap<Uuid, ManagedAtifExporter>,
+    scope_subscribers: HashMap<Uuid, String>,
+    last_error: Option<String>,
+}
+
+struct ManagedAtifExporter {
+    exporter: AtifExporter,
+    path: PathBuf,
+    observed_events: Vec<Event>,
```
mnajafian-nv, line 497: observed_events clones and stores every event for the full agent lifetime, then writes the raw event stream into ATIF extra. For long-running agents this is unbounded memory growth and duplicates the ATOF JSONL use case. If this is only diagnostic data, can we make it opt-in or remove it from the default ATIF artifact?

crates/core/src/observability/plugin_component.rs
```diff
@@ -0,0 +1,1068 @@
...
+        Self {
+            config,
+            agents: HashMap::new(),
+            scope_subscribers: HashMap::new(),
+            last_error: None,
+        }
+    }
+
+    fn observe_global(&mut self, event: &Event, subscriber_prefix: &str, state: Arc<Mutex<Self>>) {
+        if self.last_error.is_some() || !is_top_level_agent_start(event) {
```
mnajafian-nv, line 523: Once last_error is set, observe_global and observe_scope drop all future ATIF events for all agents, not just the agent that failed. A single transient write or subscriber-registration failure can permanently stop ATIF observability for the process. Can this be tracked per agent, or at least avoid dropping unrelated agents after one failure?

crates/core/src/observability/plugin_component.rs
```diff
@@ -0,0 +1,1068 @@
...
+        // trigger a second write after the dispatcher has flushed.
+        for (scope_uuid, name) in std::mem::take(&mut self.scope_subscribers) {
+            let _ = scope_deregister_subscriber(&scope_uuid, &name);
+        }
+        for agent in self.agents.values_mut() {
+            if !agent.written {
+                write_atif_file(agent)?;
+            }
+        }
+        if let Some(message) = &self.last_error {
```
mnajafian-nv, line 639: After the loop successfully writes all open agents, this still returns Err if last_error was set earlier, possibly for a different agent or event. That makes teardown report failure even when this flush completed. Can we separate historical dispatcher errors from flush failures, or clear/consume the error once it has been reported?

go/nemo_flow/observability_plugin.go
```diff
@@ -0,0 +1,111 @@
...
+	ServiceName          string            `json:"service_name,omitempty"`
+	ServiceNamespace     string            `json:"service_namespace,omitempty"`
+	ServiceVersion       string            `json:"service_version,omitempty"`
+	InstrumentationScope string            `json:"instrumentation_scope,omitempty"`
+	TimeoutMillis        uint64            `json:"timeout_millis,omitempty"`
+}
+
+// ObservabilityComponentSpec wraps one observability config as a top-level plugin component.
+type ObservabilityComponentSpec struct {
+	Enabled bool                `json:"enabled,omitempty"`
```
mnajafian-nv, line 55: This omitempty makes enabled:false impossible to send through the Go helper. The Rust PluginComponentSpec defaults a missing enabled field to true, so a Go caller that sets ObservabilityComponentSpec{Enabled:false} will serialize a component that re-enables itself. The shared Go PluginComponentSpec has the same omitempty behavior, so both layers need to preserve false. Can we make the top-level enabled flag explicit in JSON and add a Go test that a disabled observability component serializes "enabled": false and does not register exporters?

```suggestion
	Enabled bool                `json:"enabled"`
```

mnajafian-nv (commented):
Thanks for the detailed implementation and cross-binding coverage. I focused on the ATIF dispatcher lifecycle, binding parity, and docs consistency.

The main issues I found are around ATIF behavior under long-running workloads: filesystem writes happen under the dispatcher mutex, completed agents appear to remain retained after write, raw observed_events is unbounded, and last_error globally stops future collection. I also found one concrete Go binding parity issue where enabled:false is omitted and therefore becomes enabled again in Rust due to the default.

One non-blocking follow-up: ensure_builtin_plugins_registered is called from list/lookup/validate/initialize flows, and idempotency currently depends on matching "already registered" in an error string. A Once/OnceLock-style registration path would avoid repeated registry write-lock work and remove the dependency on error-message wording.

The docs/API comments are smaller and should be straightforward.

docs/export-observability-data/observability-plugin.md
```diff
@@ -0,0 +1,252 @@
...
+Direct child scopes that are not `agent` scopes do not create ATIF files.
+
+## OpenTelemetry and OpenInference Sections
+
+OpenTelemetry and OpenInference use the same section shape:
+
+| Field | Default | Notes |
+|---|---|---|
+| `enabled` | `false` | Must be `true` to construct and register the subscriber. |
+| `transport` | `http_binary` | `http_binary` or `grpc`. |
```
mnajafian-nv, line 105: This row makes grpc look uniformly supported for both OpenTelemetry and OpenInference, while the OpenInference guide currently says gRPC is unsupported. The implementation appears target-specific: native builds support OpenInference gRPC and wasm rejects it. Can we make the docs consistent here and in advanced-guide.md?

```suggestion
| `transport` | `http_binary` | `http_binary` or `grpc`. OpenInference gRPC is supported on native builds and unsupported on wasm. |
```

go/nemo_flow/observability_plugin.go
willkill07: (1) Go is still experimental and (2) why would you have a configuration for something just to disable it?

mnajafian-nv:
One small follow-up on the Go enabled:false comment: `go/nemo_flow/plugin.go` also needs the same fix for `PluginComponentSpec.Enabled`. Otherwise `enabled:false` can still be omitted before the config reaches Rust.

crates/core/src/observability/plugin_component.rs
willkill07: Coderabbit already gave this feedback -- will address.

willkill07: Will address

crates/core/src/observability/plugin_component.rs
willkill07: This is under a guard. If something fails, it must have occurred in this path

go/nemo_flow/observability_plugin.go
mnajafian-nv: That’s fair. Since Go is experimental, I would not block on this.

I flagged it because the shared plugin contract gives `enabled:false` defined semantics: validate the component, but skip activation. Go’s `omitempty` drops that value, and Rust defaults the missing field back to enabled. So this is a binding-parity gap, but I’m fine tracking it as a follow-up for when Go graduates from experimental.

crates/core/src/observability/plugin_component.rs
willkill07: I think this should be failure and should stop ATIF observability for the process

mnajafian-nv: That makes sense if the intended policy is fail-closed for ATIF after an exporter error. My main concern was making that behavior explicit, since one agent/file failure disables ATIF collection globally. If that is the intended contract, documenting it or covering it in a focused test would address my concern.

crates/core/src/observability/plugin_component.rs
mnajafian-nv: Fair. If the dispatcher is intentionally fail-closed after the first ATIF error, then this behavior is consistent with that policy. I’m fine with this being handled as part of the broader error-handling contract/test coverage.

docs/export-observability-data/observability-plugin.md
willkill07: This PR is correct.

`docs/export-observability-data/advanced-guide.md:29` says (incorrectly):

  > Use http_binary transport for the current OpenInference path. The configuration surfaces expose grpc, but the
  > current core OpenInference subscriber returns an unsupported-transport error for gRPC.

crates/core/src/observability/plugin_component.rs
willkill07: I'm not aware of any other way to implement this feature *without rewriting ATIF exporting outright*. We currently have unbounded growth with ATIF in general. This is out-of-scope for this PR.

crates/core/src/observability/plugin_component.rs
willkill07: We have two options:
1. crash on failure (not desired)
2. report error on failure + disable

This PR chose (2) due to (1) being far worse. Partial degradation is worse and hard to reason about

go/nemo_flow/observability_plugin.go
willkill07: I don't think this is a binding gap. Idiomatic Golang either (1) omit or (2) set to true

crates/core/src/observability/plugin_component.rs
mnajafian-nv: I agree report-and-disable is preferable to crashing, and I’m aligned that partial ATIF degradation would be harder to reason about. My ask is just that the fail-closed behavior is covered by docs or a focused test so operators know one exporter failure disables ATIF for the process.

crates/core/src/observability/plugin_component.rs
mnajafian-nv: Understood, not blocking.

Just to clarify the narrower point I was making: `observed_events` is a second accumulation path separate from `AtifExporter`’s internal trajectory state. The exporter builds the ATIF trajectory through `exporter.subscriber()` independently, while the Vec is later injected into `extra.observed_events`.

If keeping the raw event dump in the ATIF file is intentional, I’m fine with that being the contract. I’d just suggest documenting it or tracking a follow-up to make it opt-in/gated if the extra memory and artifact size become a concern.

docs/export-observability-data/observability-plugin.md
mnajafian-nv: Agreed. `advanced-guide.md` should be updated here so both docs match the implementation.

mnajafian-nv (approved):
LGTM overall. Only remaining suggestion from my review is docs consistency: `advanced-guide.md` still has the old OpenInference gRPC note. Updating that to match the new observability plugin doc should close out my comment.

willkill07:
@mnajafian-nv i will address as a follow up separately due to being out of scope

PR #93, opened 2026-05-13
feat(cli): load plugin configs from `plugin.toml`

bbednarski9:
#### Overview

## Summary

This PR wires the CLI gateway to the generic NeMo Flow plugin configuration path and adds a dedicated `plugin.toml` file option for process-level plugin activation.

The gateway now resolves a single `PluginConfig` from supported configuration sources, activates it when the server starts, and clears it when the server shuts down. The implementation is generic over plugin kind; the tests exercise the observability plugin today while keeping the config path usable for future plugin kinds such as adaptive once those plugins are registered in the CLI process.

## Changes

- Add daemon-mode `--plugin-config` / `NEMO_FLOW_PLUGIN_CONFIG` support.
- Add `plugin.toml` discovery and loading for root-level generic `PluginConfig` TOML.
- Resolve `plugin.toml` alongside explicit `--config path/to/config.toml`.
- Discover implicit `plugin.toml` files from system, nearest project, and user scopes.
- Reject conflicting plugin definitions across `plugin.toml`, `[plugins].config`, and `--plugin-config`.
- Preserve existing `[plugins].config` support in `config.toml`.
- Document the CLI gateway `plugin.toml` shape for observability plugin configuration.

## Validation

```bash
cargo fmt --check
cargo test -p nemo-flow-cli
cargo build -p nemo-flow-cli
```

## Notes

- `plugin.toml` uses the generic plugin config shape at the file root:

```toml
version = 1

[[components]]
kind = "observability"
enabled = true

[components.config]
version = 1
```

- This PR covers observability configuration only in examples/tests.
- Adaptive remains out of scope for this PR, but the resolver is intentionally plugin-kind agnostic.

Start with `crates/cli/src/config.rs`. The key design decision is that the CLI still resolves exactly one generic `PluginConfig`, but that config may now come from `--plugin-config`, `[plugins].config`, or `plugin.toml`. The conflict checks in that file are the main behavior to review.

- Closes #

## Summary by CodeRabbit

* **New Features**
  * Add CLI/env --plugin-config support with startup activation of a single process-level plugin configuration, discovery/precedence of plugin.toml, and daemon-mode triggering via the flag.
  * Plugin initialization and teardown tied to server lifetime.

* **Bug Fixes**
  * Improved validation, conflict detection and user-facing error reporting for plugin configuration sources and merges.

* **Documentation**
  * New CLI gateway plugin configuration docs with examples, precedence, merge semantics, and validation notes.

* **Tests**
  * Expanded integration and coverage tests for discovery, precedence, validation, and server behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NeMo-Flow/pull/93)

### DIFF
--- a/crates/cli/src/config.rs
+++ b/crates/cli/src/config.rs
@@ -703,13 +703,22 @@ fn plugin_config_paths(explicit: Option<&PathBuf>) -> Vec<PathBuf> {
             .map(|parent| vec![parent.join("plugin.toml")])
             .unwrap_or_default();
     }
+    implicit_plugin_config_paths(std::env::current_dir().ok().as_deref(), user_config_dir())
+}
+
+fn implicit_plugin_config_paths(
+    cwd: Option<&std::path::Path>,
+    user_config_dir: Option<PathBuf>,
+) -> Vec<PathBuf> {
+    // Ordered from lowest to highest precedence. User-level plugin config intentionally loads last
+    // so an operator can override project-local plugin defaults without editing the checkout.
     let mut paths = vec![PathBuf::from("/etc/nemo-flow/plugin.toml")];
-    if let Ok(cwd) = std::env::current_dir()
-        && let Some(project) = find_project_plugin_config(&cwd)
+    if let Some(cwd) = cwd
+        && let Some(project) = find_project_plugin_config(cwd)
     {
         paths.push(project);
     }
-    if let Some(user) = user_config_dir() {
+    if let Some(user) = user_config_dir {
         paths.push(user.join("plugin.toml"));
     }
     paths
@@ -884,9 +893,16 @@ struct PluginTomlConfig {
 fn load_plugin_toml_config(
     explicit: Option<&PathBuf>,
 ) -> Result<Option<PluginTomlConfig>, CliError> {
+    load_plugin_toml_config_from_paths(plugin_config_paths(explicit))
+}
+
+fn load_plugin_toml_config_from_paths<I>(paths: I) -> Result<Option<PluginTomlConfig>, CliError>
+where
+    I: IntoIterator<Item = PathBuf>,
+{
     let mut merged = toml::Value::Table(toml::map::Map::new());
     let mut sources = Vec::new();
-    for path in plugin_config_paths(explicit) {
+    for path in paths {
         if path.exists() {
             let raw = std::fs::read_to_string(&path)?;
             let parsed = raw
--- a/crates/cli/src/config.rs
+++ b/crates/cli/src/config.rs
@@ -169,6 +169,9 @@ pub(crate) struct ServerArgs {
     /// OpenInference-compatible OTLP HTTP endpoint for streaming spans (Phoenix, Arize, etc.)
     #[arg(long, env = "NEMO_FLOW_OPENINFERENCE_ENDPOINT")]
     pub(crate) openinference_endpoint: Option<String>,
+    /// Generic plugin configuration JSON for process-level gateway plugin activation.
+    #[arg(long, env = "NEMO_FLOW_PLUGIN_CONFIG")]
+    pub(crate) plugin_config: Option<String>,
 }
 
 impl ServerArgs {
@@ -184,6 +187,7 @@ impl ServerArgs {
             || self.atif_dir.is_some()
             || self.atof_dir.is_some()
             || self.openinference_endpoint.is_some()
+            || self.plugin_config.is_some()
             || self.config.is_some()
     }
 }
@@ -522,7 +526,7 @@ impl Default for GatewayConfig {
 /// server-facing command-line layer so launcher-only settings cannot leak into daemon mode.
 pub(crate) fn resolve_server_config(args: &ServerArgs) -> Result<ResolvedConfig, CliError> {
     let mut resolved = load_shared_config(args.config.as_ref())?;
-    apply_server_overrides(&mut resolved.gateway, args);
+    apply_server_overrides(&mut resolved.gateway, args)?;
     Ok(resolved)
 }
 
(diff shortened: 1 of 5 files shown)


crates/cli/src/config.rs
```diff
@@ -669,6 +693,28 @@ fn config_paths(explicit: Option<&PathBuf>) -> Vec<PathBuf> {
...
+// same directory so `--config path/to/config.toml` can be extended by `path/to/plugin.toml` without
+// reading unrelated implicit project/user/global plugin files.
+fn plugin_config_paths(explicit: Option<&PathBuf>) -> Vec<PathBuf> {
+    if let Some(path) = explicit {
+        return path
+            .parent()
+            .map(|parent| vec![parent.join("plugin.toml")])
+            .unwrap_or_default();
+    }
+    let mut paths = vec![PathBuf::from("/etc/nemo-flow/plugin.toml")];
```
willkill07, line 724: Can we pin down the intended precedence for stacked `plugin.toml` files? This currently loads `/etc/nemo-flow/plugin.toml`, then the nearest project `.nemo-flow/plugin.toml`, then the user config `plugin.toml`, so user-level values override project-local values. If the desired model is global -> user -> project, project should be pushed last and covered by a test. If user-over-project is intentional, the docs should say that explicitly.

_Submitted by Codex_

crates/cli/src/config.rs
```diff
@@ -818,6 +875,70 @@ fn apply_file_plugins_config(gateway: &mut GatewayConfig, plugins: Option<FilePl
...
+            let parsed = raw
+                .parse::<toml::Table>()
+                .map(toml::Value::Table)
+                .map_err(|error| {
+                    CliError::Config(format!(
+                        "invalid plugin TOML in {}: {error}",
+                        path.display()
+                    ))
+                })?;
+            merge_toml(&mut merged, parsed);
```
willkill07: This deep merge is important to document/test for `components`: TOML arrays are replaced, not appended, so multiple discovered `plugin.toml` files do not compose plugin components. A user-level `[[components]]` will replace the project/global `components` list. That may be fine, but the feature should either make that behavior explicit or use plugin-aware merging if stacked components are expected.

_Submitted by Codex_

docs/export-observability-data/observability-plugin.md
```diff
@@ -59,6 +59,46 @@ names from the plugin namespace:
...
+- `plugin.toml` next to the resolved `config.toml`, or in the same discovered
+  system, project, and user scopes as `config.toml`.
+
+Use only one source for plugin config. The gateway reports an error when
+`plugin.toml`, `[plugins].config`, or `--plugin-config` are used together.
+
+`plugin.toml` uses the generic plugin config shape at the file root:
+
+```toml
+version = 1
```
willkill07, line 102: Could we add one sentence saying `version = 1` is recommended but not required? Both the generic `PluginConfig.version` and the observability `components.config.version` default to `1` when omitted, while non-`1` versions fail validation by default. The example should keep the version, but users should not read it as mandatory boilerplate.

_Submitted by Codex_

willkill07 (commented):
Hey, this is an awesome step in the right direction. I cooked with Codex with the checked out PR and noticed a few things. I think one feedback cycle can address these gaps. Great job overall!

The main things I think are worth tightening are the stacking semantics for discovered `plugin.toml` files, whether `version = 1` is required or just recommended, and making the docs comprehensive enough that users can configure the full observability component shape from TOML.

_Submitted by Codex_

docs/export-observability-data/observability-plugin.md
```diff
@@ -59,6 +59,46 @@ names from the plugin namespace:
...
+enabled = true
+output_directory = "logs"
+filename = "events.jsonl"
+mode = "overwrite"
+
+[components.config.atif]
+enabled = true
+output_directory = "logs"
+filename_template = "trajectory-{session_id}.json"
+```
```
willkill07, line 160: The `plugin.toml` example only shows ATOF and ATIF, but the observability component also supports `opentelemetry`, `openinference`, and policy fields. Could we include a full TOML shape or additional snippets for those options, especially headers/resource attributes and the shared OTLP fields, so the new file format is comprehensive for observability users?

_Submitted by Codex_

crates/cli/src/config.rs
bbednarski9: I think global -> project -> user
With final taking priority. This matches config.toml and makes the most intuitive sense at least for now. We can follow-up later and change both if we see a strong rationale in the future (it should be non-breaking)

willkill07: All good!

bbednarski9: addressing in commit 09db23d4500318651270da1ebdbe42ad5c18410b

crates/cli/src/config.rs
bbednarski9: Tightened documentation around the deep merge functionality. The section now states:

  - Multiple discovered plugin.toml files load in order: system, project, user.
  - Later files override earlier files.
  - TOML tables merge recursively.
  - [[components]] is an array, so it is replaced rather than appended.
  - If user config defines components, it must include the full component list it wants active.

docs/export-observability-data/observability-plugin.md
bbednarski9: Added the clarification to NeMo-Flow/docs/export-observability-data/observability-plugin.md:93:

 "version = 1 is recommended for clarity but not required. The root plugin config version and observability component config version default to 1 when omitted; unsupported non-1 versions fail validation by default"

docs/export-observability-data/observability-plugin.md
bbednarski9: [24c1037](https://github.com/NVIDIA/NeMo-Flow/pull/93/commits/24c103711908358a19410e347d6c5b328a789ea5)

crates/cli/src/config.rs
willkill07: > [[components]] is an array, so it is replaced rather than appended.

I think this is undesirable behavior. I think each element should be upserted based on `kind`

bbednarski9: Yes great catch tysm.

Addressed here [831a856](https://github.com/NVIDIA/NeMo-Flow/pull/93/commits/831a8563d685ef831b3f3911e23fdff6b432ab38)

docs/export-observability-data/observability-plugin.md
bbednarski9: [7c314ce](https://github.com/NVIDIA/NeMo-Flow/pull/93/commits/7c314ce64e3898f029fe7b12e62c8e9fcda007b2)

willkill07 (approved):

PR #128, opened 2026-05-19
feat: enable adaptive plugin for CLI and OpenClaw

willkill07:
#### Overview

Enable the adaptive plugin path for `nemo-flow-cli` and the public `nemo-flow-openclaw` integration package.

#### Details

- Register the adaptive plugin component before CLI plugin validation and activation so adaptive plugin config works transparently.
- Forward the effective `LlmRequest` produced by managed execution request intercepts to the upstream HTTP request body and headers.
- Expose the adaptive Node helper module in the OpenClaw integration loader and document the supported hook-backed adaptive configuration path.
- Add CLI and OpenClaw tests covering adaptive config validation, TOML round-trip behavior, request forwarding helpers, and adaptive helper component initialization.

Validation run:

- `cargo fmt --all`
- `npm run typecheck --workspace=nemo-flow-openclaw`
- `cargo test -p nemo-flow-cli`
- `npm test --workspace=nemo-flow-openclaw`
- `git diff --check`
- Commit-time pre-commit hooks, including Cargo clippy/check/deny and docs markdown linkcheck

Start with `crates/cli/src/gateway.rs` for the managed execution request forwarding change, then `crates/cli/src/server.rs` and `crates/cli/src/plugins/config_io.rs` for adaptive plugin registration.

- Closes NMF-150

## Summary by CodeRabbit

* **New Features**
  * Adaptive plugin component integrated alongside observability monitoring.
  * Runtime requests can supply overlayed JSON bodies and headers for managed execution paths; header sanitization is applied.
  * Runtime modules now include adaptive bindings so adaptive features are loadable at startup.

* **Behavior**
  * Adaptive component registration is validated and activated during startup to prevent misconfiguration.

* **Documentation**
  * Plugin configuration guides updated with adaptive component examples and behavior notes.

* **Tests**
  * Expanded unit, integration, and live-smoke tests covering adaptive plugin handling and effective request forwarding.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NeMo-Flow/pull/128?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Salonijain27 (dismissed):

crates/cli/Cargo.toml
```diff
@@ -22,6 +22,7 @@ workspace = true
 
 [dependencies]
 nemo-flow = { workspace = true, features = ["openinference"] }
+nemo-flow-adaptive = { workspace = true, features = ["redis-backend"] }
```
bbednarski9, line 25: Did we consider having this an an optional dep so that we dont have redis-backend pulled in when it might not be on host syste,s?

crates/cli/src/server.rs
```diff
@@ -153,6 +154,9 @@ impl PluginActivation {
         let Some(config) = config else {
             return Ok(Self { active: false });
         };
+        register_adaptive_component().map_err(|error| {
+            CliError::Config(format!("adaptive plugin registration failed: {error}"))
+        })?;
```
bbednarski9, line 159: Just noting unconditional registration here corresponding to Cargo.toml

crates/cli/src/gateway.rs
```diff
@@ -657,9 +671,11 @@ async fn forward_upstream_request(
     url: &str,
     body_bytes: &Bytes,
     headers: &HeaderMap,
+    effective_request: Option<&LlmRequest>,
     route: ProviderRoute,
 ) -> Result<reqwest::Response, reqwest::Error> {
-    let sanitized = gateway_forward_headers(headers, route);
+    let (body_bytes, headers) = effective_upstream_request(body_bytes, headers, effective_request);
+    let sanitized = gateway_forward_headers(&headers, route);
```
bbednarski9: Function name is misleading. I expected gateway_forward_headers to be the generic forward/sanitize step. It isn't — it's a Codex-OAuth-strip step. A name like strip_codex_oauth_if_replaceable would convey intent better. Probably not worth blocking on, but a code-quality nit.

crates/cli/src/gateway.rs
```diff
@@ -670,6 +686,43 @@ async fn forward_upstream_request(
...
+    effective_request: Option<&LlmRequest>,
+) -> (Bytes, HeaderMap) {
+    let Some(request) = effective_request else {
+        return (body_bytes.clone(), headers.clone());
+    };
+
+    let body_bytes = if request.content.is_null() {
+        body_bytes.clone()
+    } else {
+        serde_json::to_vec(&request.content)
```
bbednarski9: Silent body-serialization fallback. If serde_json::to_vec(&request.content) fails, the function returns the original body but still applies the new headers. That's an inconsistent state — headers may advertise the new payload while bytes are old. Quietly degrading is the wrong default here; a serialization failure of intercept output should bubble up as an error (or at minimum emit a warning event).

crates/cli/src/gateway.rs
```diff
@@ -670,6 +686,43 @@ async fn forward_upstream_request(
...
+    };
+    let mut headers = headers.clone();
+    for (name, value) in &request.headers {
+        let Ok(name) = HeaderName::from_bytes(name.as_bytes()) else {
+            continue;
+        };
+        let Some(value) = json_header_value(value) else {
+            continue;
+        };
+        headers.insert(name, value);
```
bbednarski9, line 719: Is there a scenario where we would want to remove a header? Right now there's no way to remove a header. Insert-only semantics. An intercept can override or add, but can't strip (e.g., remove Authorization before forwarding). If the design intends header-removal, this needs an additional channel (e.g., a separate remove_headers: Vec<String> field).

bbednarski9 (changes requested):
Just some comments above asking for clarification on a few things

crates/cli/Cargo.toml
willkill07: Rust features are effectively compile-time guards. We have to include it for it to be configurable. Runtime behavior is separate from compile-time inclusion

crates/cli/src/server.rs
willkill07: Leaving this registration unconditional to match the current CLI dependency behavior. If we make the adaptive dependency optional in a follow-up, this registration should be gated by the same feature so config validation and activation stay aligned.

crates/cli/src/gateway.rs
willkill07: Agreed. Renamed the local gateway helper to `strip_replaceable_agent_auth_headers` and renamed the test helper to `strip_replaceable_agent_auth_headers_with_openai_key_state` so the name matches the Codex OAuth replacement behavior.

crates/cli/src/gateway.rs
willkill07: Adjusted this while keeping the gateway callback path infallible. If rewritten body serialization ever fails, the function now falls back atomically to the original body and original headers, with a local warning, rather than mixing original bytes with rewritten headers. The `null` sentinel still preserves the original body while allowing header overlays.

crates/cli/src/gateway.rs
willkill07: Header removal is intentionally unsupported in the current `LlmRequest` shape. Runtime request headers are add/override overlays only; deletion would need a separate cross-binding API design, such as an explicit `remove_headers` field, rather than being added implicitly in this PR.

bbednarski9 (approved):

Salonijain27 (approved):

PR #129, opened 2026-05-19
feat: add adaptive plugin editor support

willkill07:
#### Overview

Stacked PR on top of #128 (`wkk_enable-adaptive-plugin`). NVIDIA repository rules block creating the stack base branch directly in `NVIDIA/NeMo-Flow`, so this PR targets `main` but is intended to be reviewed and merged after #128.

Focused stacked diff: https://github.com/willkill07/NeMo-Flow/compare/wkk_enable-adaptive-plugin...wkk_adaptive-plugin-editor

Adds `nemo-flow plugins edit` support for Adaptive plugin configuration options without exposing the default adaptive schema `version` as an operator-edited field.

- [x] I searched existing issues and open pull requests, and this does not duplicate existing work. This is intentionally stacked on #128.

#### Details

- Add editor metadata for Adaptive config sections: `agent_id`, `state.backend`, `telemetry`, `adaptive_hints`, `tool_parallelism`, `acg`, `acg.stability_thresholds`, and `policy`.
- Extend the CLI plugin editor to show Observability and Adaptive components in one `plugins.toml` editor flow.
- Add recursive nested-section editing, float field prompts for ACG thresholds, and cursor-position memory when navigating back from selected entries.
- Keep default adaptive `version = 1` out of editor-created and editor-saved adaptive config.
- Update tests and docs for the expanded editor behavior.

Validation run:

- `cargo fmt --all`
- `cargo test -p nemo-flow-cli plugins::tests`
- `cargo test -p nemo-flow-adaptive test_adaptive_editor_schema_covers_canonical_options`
- `cargo clippy -p nemo-flow-cli --all-targets -- -D warnings`
- `git diff --check`
- Commit-time pre-commit hooks, including Cargo clippy/check and docs markdown linkcheck

Start with `crates/cli/src/plugins.rs` for the stacked editor flow. Then review `crates/adaptive/src/config.rs` for the Adaptive editor metadata. This PR is stacked against #128; use the focused compare above to avoid re-reviewing the #128 base changes.

- Closes NMF-151

## Summary by CodeRabbit

* **New Features**
  * Interactive CLI editor now supports both Observability and Adaptive plugin configuration
  * Added Float field type support in plugin configuration editor

* **Bug Fixes**
  * Improved numeric input validation for float fields with clearer error messages
  * Enhanced default initialization for configuration maps

* **Documentation**
  * Clarified `agent_id` as a fallback identifier in Adaptive plugin configuration
  * Updated plugin editing guide to document both configuration types

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NeMo-Flow/pull/129?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Salonijain27 (dismissed):

crates/adaptive/src/config.rs
```diff
@@ -209,6 +215,127 @@ fn default_acg_priority() -> i32 {
     50
 }
 
+nemo_flow::editor_config! {
+    impl AdaptiveConfig {
+        agent_id => { label: "agent_id", kind: String, optional: true },
```
bbednarski9: Reading crates/adaptive/src/adaptive_hints_intercept.rs:122-137 and crates/adaptive/src/context_helpers.rs:147-170:

  fn effective_agent_id(&self) -> String {
      resolve_agent_id().unwrap_or_else(|| self.agent_id.clone())
  }

  resolve_agent_id() walks the active scope stack and returns the first ScopeType::Agent scope's name. The configured agent_id is the fallback used only when no Agent scope is on the stack. So:

  - The "real" agent identity at runtime is dynamic, per-request, scope-resolved.
  - The configured AdaptiveConfig.agent_id is a static default for callers operating outside an Agent scope (e.g., the gateway path, where requests arrive over HTTP without a Rust-side scope having been opened).
  - A single deployed CLI/gateway can absolutely serve many agents (and many sessions per agent) with their own scope-resolved agent_ids. The configured value applies only to traffic where scope resolution
  returns None.

  What this means for the editor:

  1. The field belongs in the config struct — it's a real, persisted setting. The editor exposing it isn't wrong on those grounds.
  2. But the label "agent_id" is misleading. A user editing plugins.toml and seeing agent_id will reasonably assume "this is THE agent." For a gateway deployment serving multiple agents, that mental model is wrong.
  3. Suggested fix: rename the editor label (not the field — that would be a breaking change) to something like "default_agent_id" or "fallback_agent_id", and add a help string explaining "Used as the agent identifier when no Agent scope is active on the runtime stack (e.g., gateway-mode requests)." The editor_config! macro doesn't seem to support help text yet — that might be worth adding.
  4. Alternative: if the team decides this fallback isn't actually a knob users should reach for (and gateway-mode agents should always come from a request header or session manager instead), the editor could nnot expose agent_id and leave it as an advanced/programmatic-only field. Worth asking the author what scenarios they expect users to set this value for.

  So: keep the field, fix the labeling/docs, OR hide it from the editor and let advanced users hand-edit the TOML. Don't leave it as-is — the cognitive trap is real.

crates/cli/src/plugins/editor_model.rs
```diff
@@ -260,6 +418,21 @@ pub(super) fn observability_config_map(
     }
 }
 
+pub(super) fn adaptive_config_map(config: &AdaptiveConfig) -> Result<Map<String, Value>, CliError> {
+    let value = serde_json::to_value(config).map_err(serde_error)?;
+    match value {
+        Value::Object(mut map) => {
+            if map.get("version") == Some(&json!(1)) {
```
bbednarski9: this feels brittle

crates/cli/src/plugins.rs
```diff
@@ -593,6 +1042,20 @@ fn prompt_value(
     }
 }
 
+fn parse_float_value(field: &EditorFieldSpec, value: &str) -> Result<Value, CliError> {
```
bbednarski9, line 1084: Could add optional range validation here for things like stable_threshold/semi_stable_threshold where we can reasonably validate ranges not just finite and non-null

willkill07: No... that's not the point of this. It's intended to be broad/generic. plugins add their own prescriptive validation. This is simple parsing.

crates/cli/src/plugins/editor_model.rs
bbednarski9: specifically "Some(&json!(1))"

bbednarski9:
Stepped through the user experience for this config with a hermes agent... here are my notes:

```
cargo build -p nemo-flow-cli
```
PASS

```
./target/debug/nemo-flow plugins edit --project
```
Notes: 
   - Note: nvext is enabled as default when we turn on adaptive component. This probably makes sense for the testing we've done so far, but we might want to superclass some preset adaptive subcomponents eventually
PASS

```
./target/debug/nemo-flow config hermes
```
PASS

```
nemo-flow doctor hermes
```
PASS

bbednarski9 (approved):

PR #130, opened 2026-05-19
fix: align coding agent trace scopes

willkill07:
#### Overview

This PR improves CLI trace alignment for coding-agent runs so LLM and tool spans correlate to the correct turn, agent, or subagent scope across Codex and Claude Code flows. It keeps long-lived harness sessions as metadata, represents user-visible work as bounded turn agent scopes, and moves provider-specific correlation into alignment modules so session and gateway code can stay focused on lifecycle orchestration.

#### Details

- Adds `crates/cli/src/alignment/` as the owner for Codex- and Claude Code-specific trace alignment, session id recovery, subagent aliasing, Codex ChatGPT OAuth gateway normalization, and reusable metadata helpers.
- Updates session lifecycle handling so Codex and Claude Code `SessionStart` events record metadata but do not emit long-lived top-level session spans.
- Represents user-visible work as provider-named turn agent scopes such as `codex-turn`, `claude-code-turn`, and `hermes-turn`, with `nemo_flow_scope_role`, `turn_index`, and harness metadata for filtering.
- Keeps subagent starts as `agent` scopes under the active turn, including Codex child-thread promotion and Claude Code `Agent` tool completion cleanup.
- Adds aggressive idle cleanup for open turn scopes while preserving active gateway calls, hook-observed LLMs/tools, and subagent cleanup ordering.
- Updates LLM/tool correlation with alias routing, sticky owner recovery, request-affinity matching, final root-owned turn output, and streamed final-response handoff.
- Updates gateway code so provider-specific session/auth/upstream fallbacks are delegated to alignment while the gateway keeps generic request preparation and provider forwarding logic.
- Updates ATIF trajectory dispatch so top-level agent scopes remain the trajectory roots now that turns are modeled as agent scopes.
- Adds out-of-source coverage tests under `crates/cli/tests/coverage` for alignment adapters, gateway identifier/auth behavior, session aliasing, subagent cleanup, turn naming/output, idle cleanup, streaming final responses, and LLM/tool correlation.
- Validation performed:
  - `git diff --check`
  - `cargo test -p nemo-flow-cli`
  - `cargo test -p nemo-flow observability`
  - `cargo clippy -p nemo-flow-cli --all-targets -- -D warnings`
  - commit pre-commit hooks, including `cargo fmt`, `cargo clippy`, and `cargo check`

- Start with `crates/cli/src/alignment/mod.rs` to understand the boundary: alignment owns provider-specific interpretation and cross-session alias state, while session/gateway code stays generic.
- Then read `crates/cli/src/session.rs` by lifecycle path rather than linearly: `start_agent`, `open_turn`, `start_subagent`, `close_turn`, `resolve_llm_owner`, and `resolve_tool_owner` are the main decisions.
- Review the turn design strategically first: Codex and Claude Code sessions no longer create top-level session spans; turn scopes are provider-named `agent` scopes; subagent scopes still emit under the active turn.
- Review `crates/cli/src/alignment/codex.rs` and `crates/cli/src/alignment/claude_code.rs` as adapters. Codex covers child-thread parentage and ChatGPT OAuth gateway normalization; Claude Code covers native session headers and `Agent` tool completion.
- Review `crates/cli/src/gateway.rs` for the managed execution handoff, especially streaming final-response capture and active-call accounting for idle cleanup.
- Use tests as executable examples before reading every implementation detail: `alignment_tests.rs` for generic alias routing, `alignment_codex_tests.rs` / `alignment_claude_code_tests.rs` for adapter-specific behavior, `session_tests.rs` for turn/subagent ownership, and `gateway_tests.rs` for gateway handoff cases.
- To avoid getting overwhelmed, verify the PR in layers: provider adapter extraction first, turn scope lifecycle second, subagent alias promotion third, LLM/tool owner metadata fourth, and gateway auth/session fallback last.
- The highest-risk behavior to scrutinize is whether child sessions are promoted only while empty, whether aliases clear on child or parent terminal events, and whether ambiguous tool/LLM calls fall back to the turn instead of guessing a subagent.

- Relates to: none

## Summary by CodeRabbit

* **New Features**
  * Provider-aware session alignment and subagent correlation for Claude Code and Codex; more reliable session/subagent id handling and cross-session metadata.

* **Refactor**
  * Centralized gateway URL/header/auth normalization and session-id resolution into a shared alignment layer.
  * Session lifecycle and subagent scope handling redesigned for concurrent and out-of-order events.

* **Tests**
  * Extensive new coverage for gateway routing, alignment, subagent promotion, metadata extraction, and session lifecycle.

* **Other**
  * Background idle-session sweeper added; observability docs updated; prompt-submission event handling adjusted.


crates/cli/src/session.rs
```diff
@@ -340,6 +485,183 @@ impl SessionManager {
...
+        Ok(alias)
+    }
+}
+
+// Mutates a gateway LLM start in place after alias resolution. The parent session id is what the
+// runtime session manager should open, while the subagent id and alias metadata preserve the child
+// thread as the LLM owner.
+fn apply_start_alias(start: &mut LlmGatewayStart, alias: &SessionAlias) {
+    start.session_id = Some(alias.parent_session_id.clone());
+    if start.subagent_id.is_none() {
```
mnajafian-nv: I think the alias should probably be authoritative here. `resolve_start_alias()` has already proven that this child session should run under `alias.parent_session_id` as `alias.subagent_id`, and we rewrite `session_id` accordingly. If we preserve a pre-existing `start.subagent_id`, a stale/conflicting subagent header can still win in `explicit_llm_owner()` and attach the child-thread LLM to a sibling. Would it make sense to overwrite `start.subagent_id` from the alias and add a regression test for that conflicting-header case?

crates/cli/src/session.rs
```diff
@@ -1331,67 +2109,63 @@ fn push_tool_hint(
...
-        Some(current.clone())
-    })
+fn owner_status_teaches_request_affinity(status: &str) -> bool {
+    matches!(
+        status,
+        "explicit"
+            | "single_hint"
+            | "matched_hint"
+            | "active_subagent"
+            | "subagent_start"
```
mnajafian-nv, line 2164: I think `subagent_start` may be too weak to teach reusable request affinity. The source is useful as a short-lived "new worker likely owns the next unhinted LLM" signal, but `llm_request_affinity` persists ownership by request payload and is checked before sticky ownership. Would it be safer to keep `subagent_start` as sticky/one-shot only and exclude it from affinity learning?

mnajafian-nv (dismissed):
Thanks, this trace-shape direction makes sense to me. The overall model of bounded turn scopes, child-session aliases, and request-affinity fallback looks like the right direction for coding-agent observability.

I left two small inline questions around alias authority and whether `subagent_start` should teach reusable request affinity. Neither blocks my approval, but I think they are worth considering before merge or as a follow-up.

crates/cli/src/session.rs
willkill07: Already addressed on this branch -- we overwrite both fields:

```
start.session_id = Some(alias.parent_session_id.clone())
start.subagent_id = Some(alias.subagent_id.clone())
```

So a stale/conflicting incoming `start.subagent_id` cannot win in `explicit_llm_owner()` after alias resolution. There is also an existing regression test: `apply_start_alias_overrides_conflicting_subagent_id`

crates/cli/src/session.rs
willkill07: It's a hint and not an absolute.

mnajafian-nv (approved):
LGTM!

PR #131, opened 2026-05-19
feat: add nemo_guardrails plugin contract

afourniernv:
#### Overview

Adds the first core-owned `nemo_guardrails` plugin contract for NeMo Flow.

This PR intentionally stays at the contract layer:

- adds a public typed Rust config surface for the planned first-party `nemo_guardrails` plugin
- models `remote` and `local` backend modes
- adds validation, schema support, and editor metadata
- does not auto-register the plugin yet
- does not implement backend runtime behavior yet

#### Details

- add `crates/core/src/plugins/nemo_guardrails/` as the public core module for the planned Guardrails plugin surface
- add `ComponentSpec`, `NeMoGuardrailsConfig`, backend config sections, request defaults, and manual registration helpers
- model the backend split explicitly:
  - `mode = remote | local`
  - local config-source pass-through:
    - `config_path`
    - `config_yaml`
    - `colang_content`
  - remote config identity:
    - `remote.endpoint`
    - `remote.config_id`
    - `remote.config_ids`
- keep NeMo Flow-owned interception toggles in the contract:
  - `input`
  - `output`
  - `tool_input`
  - `tool_output`
- widen the contract to include backend request defaults without claiming full backend parity:
  - `context`
  - request-time rail selection
  - `llm_params`
  - `output_vars`
  - `llm_output`
  - `log`
- add focused unit coverage for:
  - schema/editor metadata
  - singleton behavior
  - explicit registration vs auto-registration
  - invalid config shapes and values
  - nested unknown-field policy
  - fail-fast initialization until backend implementation lands
- move the module under `crates/core/src/plugins/nemo_guardrails/` and use the `nemo_guardrails` naming consistently

Validation:

- `cargo fmt --all`
- `cargo test -p nemo-flow plugin --lib --tests`
- `cargo test -p nemo-flow nemo_guardrails --lib --tests`

Start in `crates/core/src/plugins/nemo_guardrails/plugin_component.rs`.

The most important design decision in this PR is that it stops at the public contract layer: it makes the Guardrails surface broad enough to avoid repeating the old narrow example-plugin shape, but it does not auto-register or implement the backend yet.

- Relates to #NMF-72
- Relates to #NMF-131

## Summary by CodeRabbit

* **New Features**
  * Exposed a first‑party Guardrails plugin in the core distribution with configurable remote and local backends, registration helpers, and optional config schema generation.
  * Enhanced configuration model with surface toggles, request defaults, and policy-driven behavior.

* **Validation**
  * Added comprehensive config validation with detailed diagnostics for shape, semantic rules, and nested unknown-field handling.

* **Tests**
  * Added extensive unit tests covering registration, validation, schema generation, and initialization behaviors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NeMo-Flow/pull/131?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -61,6 +61,7 @@ pub mod config_editor;
 mod context;
 pub mod error;
 pub mod json;
+pub mod nemoguardrails;
 pub mod observability;
 pub mod plugin;
 pub mod registry;
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -61,9 +61,9 @@ pub mod config_editor;
 mod context;
 pub mod error;
 pub mod json;
-pub mod nemoguardrails;
 pub mod observability;
 pub mod plugin;
+pub mod plugins;
 pub mod registry;
 #[doc(hidden)]
 pub mod shared_runtime;
--- a/crates/core/src/nemoguardrails/mod.rs
+++ b/crates/core/src/nemoguardrails/mod.rs
@@ -1,0 +1,14 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Planned NeMo Guardrails integrations for NeMo Flow Core.
+
+#[cfg(test)]
+use std::sync::Mutex;
+
+#[cfg(test)]
+pub(crate) fn test_mutex() -> &'static Mutex<()> {
+    crate::shared_runtime::runtime_owner_test_mutex()
+}
+
+pub mod plugin_component;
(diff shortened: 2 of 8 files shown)


willkill07:
From a design choice -- do we want this to live in the core crate or be part of a new crate?

If it's part of core, we should be forward thinking and have a `src/plugins` module

afourniernv:
@willkill07  I think a separate crate makes sense if this starts adding deps to core.

Right now PR 1 doesn’t. Remote mode (PR2) doesn’t either. Local mode (PR3) as far as I researched is really the Python SDK/runtime path, not just another transport aka “call a localhost server.” If we want faithful local behavior, I’d prefer a Python helper/subprocess over in-process Python embedding from Rust.

So I don’t think there’s a strong dependency reason to split into a separate crate yet. 

If you just want it totally modularized out of core I can design it that way.

willkill07:
@afourniernv the only shape difference for crate vs part of core would be where things live.

We need to switch to either:
1. `crates/core/src/plugins/nemo_guardrails`
2. `crates/nemo_guardrails/src/`

If you don't think there will be added deps, then we should prefer option (1)

`nemoguardrails` is too hard to read so we should prefer `nemo_guardrails`

Let me know when this is updated.

willkill07 (approved):

PR #144, opened 2026-05-21
feat: implement nemo_guardrails remote backend

afourniernv:
#### Overview

Adds the first real built-in `nemo_guardrails` remote backend for NeMo Relay core.

This PR moves beyond the contract-only surface and implements the first shippable remote slice:

- built-in plugin auto-registration in core
- remote backend activation for `mode = remote`
- non-streaming and streaming LLM execution through the Guardrails server
- managed `tool_output` execution checks through the same remote backend
- request-default pass-through for remote Guardrails request semantics
- explicit validation that stock remote mode does not currently support managed `tool_input`
- focused runtime validation and coverage for transport, malformed responses, tool rewrites, and error handling

#### Details

- auto-register the built-in `nemo_guardrails` plugin through `ensure_builtin_plugins_registered()`
- add a dedicated default core feature `guardrails-remote` for the remote backend dependency path
- implement the remote Guardrails runtime across:
  - `crates/core/src/plugins/nemo_guardrails/component.rs`
  - `crates/core/src/plugins/nemo_guardrails/remote.rs`
- support the native remote server contract for `codec = "openai_chat"`
- support both non-streaming and streaming LLM execution
- support remote request defaults pass-through:
  - `context`
  - `thread_id`
  - `state`
  - `rails`
  - `llm_params`
  - `llm_output`
  - `output_vars`
  - `log`
- support managed `tool_output` execution checks via the remote backend
- reject managed `tool_input` in stock remote mode with a focused validation error instead of silently leaving it non-enforcing
- emit coarse backend-level `nemo_guardrails.remote.*` marks for both LLM and managed tool remote checks
- precompute stable remote Guardrails payloads during runtime initialization instead of rebuilding them per request
- split remote runtime coverage into `component_tests.rs` and `remote_tests.rs` to keep contract/config tests separate from transport/runtime behavior
- add focused unit coverage for:
  - built-in registration
  - remote config validation
  - transport and malformed-response lanes
  - non-streaming and streaming remote execution
  - managed `tool_output` blocking and rewrite behavior
  - tool-only config isolation from LLM execution
  - context/state/thread propagation on managed tool checks

Rail and observability boundary in this PR:

- Native NeMo Relay-managed surfaces:
  - `input`
  - `output`
  - `tool_output`
- These are real runtime interception surfaces in NeMo Relay. The built-in plugin installs managed LLM and tool execution behavior around them, and the remote backend can block or rewrite those executions directly.
- Managed `tool_input` remains a known NeMo Relay surface, but stock remote mode now rejects it explicitly because the Guardrails remote contract does not currently activate pre-execution tool-call rails from externally submitted chat-completions history.
- Remote request-time rail pass-through:
  - `request_defaults.rails.input`
  - `request_defaults.rails.output`
  - `request_defaults.rails.retrieval`
  - `request_defaults.rails.dialog`
  - `request_defaults.rails.tool_input`
  - `request_defaults.rails.tool_output`
- These are forwarded to the Guardrails server as remote request semantics. In this PR, `retrieval` and `dialog` are supported this way only: NeMo Relay does not currently expose separate managed retrieval or dialog execution surfaces to intercept locally.
- The remote backend emits coarse backend-level marks:
  - `nemo_guardrails.remote.start`
  - `nemo_guardrails.remote.end`
  - `nemo_guardrails.remote.error`
- These marks cover managed LLM remote execution and managed `tool_output` remote checks. They provide backend-level visibility into remote Guardrails activity without introducing separate NeMo Relay-native retrieval or dialog scopes in this slice.

Intentional PR2 boundary:

- native remote support is `openai_chat` only
- local backend is still not implemented
- stock remote mode supports managed `tool_output` but not managed `tool_input`
- this stays core-only; binding surface enablement is deferred

Validation:

- `cargo fmt --all`
- `cargo test -p nemo-relay nemo_guardrails --lib --tests`
- `uv run pre-commit run --files crates/core/src/plugins/nemo_guardrails/component.rs crates/core/tests/unit/plugins/nemo_guardrails/component_tests.rs crates/core/tests/unit/plugins/nemo_guardrails/remote_tests.rs`

Targeted live validation covered:

- allowed and blocked non-streaming LLM requests
- streaming requests
- trusted HTTPS transport
- managed `tool_output` blocked path
- direct stock Guardrails repro confirming externally supplied pre-exec tool-call history does not activate remote managed `tool_input`

Notes:

- `uv run pre-commit run --all-files` still reports an unrelated repo-environment failure in `ty` for unresolved `deepagents` integration imports
- unrelated `package-lock.json` churn was restored and is not part of this PR

Start in `crates/core/src/plugins/nemo_guardrails/component.rs`, then `crates/core/src/plugins/nemo_guardrails/remote.rs`.

The most important design decision in this PR is that the built-in plugin now implements a real remote execution backend while keeping the remote contract honest:

- native remote support is OpenAI chat-completions shaped
- LLM and managed `tool_output` execution are supported through the core runtime
- stock remote managed `tool_input` is explicitly rejected instead of being left as a silent best-effort path
- broader non-chat codec parity and the local Python runtime remain out of scope for this slice

- Relates to #NMF-131
- Relates to #NMF-148


afourniernv:
@willkill07 @dagardner-nv live tested + no cargo blast radius. Did have to add some minimal deps. ready for review.

crates/core/src/plugins/nemo_guardrails/remote.rs
```diff
@@ -0,0 +1,944 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#[cfg(all(not(target_arch = "wasm32"), feature = "guardrails-remote"))]
```
willkill07: If you have this at the top of the file, I don't believe you need it on every single include:

```rust
#![cfg(all(not(target_arch = "wasm32"), feature = "guardrails-remote"))]
```

crates/core/src/plugins/nemo_guardrails/component.rs
```diff
@@ -17,9 +17,25 @@ use crate::plugin::{
     lookup_plugin, register_plugin,
 };
 
+#[cfg(all(feature = "guardrails-remote", not(target_arch = "wasm32")))]
+#[path = "remote.rs"]
+mod remote;
```
willkill07, line 22: Actually, since you have this, I don't think you need all of the `cfg` entries in `remote.rs`

crates/core/src/plugins/nemo_guardrails/remote.rs
```diff
@@ -0,0 +1,944 @@
...
+#[derive(Clone, Copy)]
+enum RemoteCheckKind {
+    Input,
+    Output,
+}
+
+#[cfg(all(not(target_arch = "wasm32"), feature = "guardrails-remote"))]
+impl RemoteBackendRuntime {
+    fn new(config: &NeMoGuardrailsConfig, remote: &RemoteBackendConfig) -> PluginResult<Self> {
```
willkill07: Doesn't `NeMoGuardrailsConfig` already hold the optional `RemoteBackendConfig` ?

Shouldn't we just access that?

crates/core/src/plugins/nemo_guardrails/remote.rs
```diff
@@ -0,0 +1,944 @@
...
+                "nemo_guardrails remote backend does not support OpenAI tool definitions or tool_choice yet"
+                    .to_string(),
+            ));
+        }
+
+        let mut body = request.content.as_object().cloned().ok_or_else(|| {
+            FlowError::Internal("LLM request content is not a JSON object".to_string())
+        })?;
+        body.insert("stream".to_string(), Json::Bool(stream));
+        if let Some(guardrails) = self.build_guardrails_config() {
```
willkill07: We should not have to build the guardrails config with each outgoing request. Recommendation: hoist all logic as part of Runtime initialization

crates/core/tests/unit/plugins/nemo_guardrails/component_tests.rs
willkill07, line 1: This file appears to be doing more than just the component contract. Maybe split remote tests into their own file.

willkill07 (changes requested):
First brief pass. I'll have a more complete pass later.

willkill07:
/ok to test [60999ac](https://github.com/NVIDIA/NeMo-Relay/pull/144/commits/60999aca0f886f72d2bad9aafb6ec6fc7d113fd6)

crates/core/src/plugins/nemo_guardrails/remote.rs
afourniernv: done - that stunk.

crates/core/src/plugins/nemo_guardrails/component.rs
afourniernv: done.

crates/core/src/plugins/nemo_guardrails/remote.rs
afourniernv: true - done

crates/core/src/plugins/nemo_guardrails/remote.rs
afourniernv: yea a singleton makes way more sense. resolved.

crates/core/tests/unit/plugins/nemo_guardrails/component_tests.rs
afourniernv: split. done.

willkill07:
/ok to test [38aa2e3](https://github.com/NVIDIA/NeMo-Relay/pull/144/commits/38aa2e39809da96f68b4e47ae1be86084b0c1fa8)

crates/core/src/plugin.rs
```diff
@@ -762,9 +762,11 @@ pub fn register_plugin(plugin: Arc<dyn Plugin>) -> Result<()> {
 /// Built-in plugins are available to validation and initialization without a
 /// binding or application-specific registration call.
 pub fn ensure_builtin_plugins_registered() -> Result<()> {
-    match BUILTIN_PLUGIN_REGISTRATION
-        .get_or_init(crate::observability::plugin_component::register_observability_component)
-    {
+    let register_builtins = || {
+        crate::observability::plugin_component::register_observability_component()?;
+        crate::plugins::nemo_guardrails::component::register_nemo_guardrails_component()
```
willkill07, line 767: Shouldn't this be guarded on `!wasm` ?

crates/core/src/plugins/nemo_guardrails/remote.rs
```diff
@@ -0,0 +1,989 @@
...
+            tool_output_guardrails: build_tool_check_guardrails_config(
+                RemoteCheckKind::Output,
+                &remote.config_id,
+                &remote.config_ids,
+                request_defaults,
+            ),
+        })
+    }
+
+    async fn execute(&self, request: LlmRequest, stream: bool) -> crate::error::Result<Json> {
```
willkill07, line 109: cognitive complexity of this function is really high -- can we factor out common logic? (same with `execute_stream`)

crates/core/src/plugin.rs
afourniernv: I think “known but unavailable” is the better behavior.

but If you  specifically want builtin registration to mean “fully usable on this target,” then yes, we should gate it.

crates/core/src/plugins/nemo_guardrails/remote.rs
afourniernv: Done.

I factored the common request/response plumbing out of `execute` / `execute_stream`, and also pulled the spawned SSE decode loop into a helper so the top-level methods read more as orchestration than transport detail.

crates/core/src/plugin.rs
willkill07: Ack. Sensible.

willkill07:
/ok to test [50a9a44](https://github.com/NVIDIA/NeMo-Relay/pull/144/commits/50a9a4423da60580aa7ec5f21d1b347ff5b35f16)

willkill07 (approved):

PR #171, opened 2026-05-28
fix: correlate ATIF tool observations

willkill07:
#### Overview

Fix ATIF export so tool outputs can be correlated to LLM tool calls even when the tool execution events are emitted before the LLM events that declare `tool_calls`.

#### Details

- Adds an internal ATIF correlation pre-pass that maps tool scope UUIDs to LLM tool call IDs using explicit IDs first, then exact tool name plus normalized arguments.
- Defers correlated tool observations until the matching agent step exists, so Hermes-style traces attach observations to the correct `tool_calls` instead of emitting standalone system observations.
- Keeps ambiguous unmatched tool results standalone rather than guessing, and preserves structured JSON observation content for object/array tool outputs.
- Adds unit coverage for Hermes-style ordering, repeated identical calls, ambiguous no-argument calls, and structured observation content.

Start with `crates/core/src/observability/atif.rs`, especially the correlation lookup and deferred observation handling. The regression coverage is in `crates/core/tests/unit/atif_tests.rs`.

- Closes RELAY-169

Validation:

- `cargo test -p nemo-relay atif`
- commit hooks: `cargo fmt`, `cargo clippy`, `cargo check`, docs linkcheck, and file hygiene hooks

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved per-agent correlation of tool lifecycle events to LLM tool calls, reducing misattributed observations and ensuring tool metadata/lineage attach correctly or are deferred and flushed as standalone observations.
  * Removed duplicate projected tool-call steps and ignored runtime “llm.chunk” marks; observation content is now enforced non-null in exports.

* **Tests**
  * Added tests covering out-of-order and repeated tool-call correlation, duplicate removal, chunk-mark skipping, and observation content validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NeMo-Relay/pull/171?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/core/src/observability/atif.rs
+++ b/crates/core/src/observability/atif.rs
@@ -464,8 +464,8 @@ fn atif_content_value(value: &Json) -> Json {
     }
 }
 
-fn observation_content_value(value: &Json) -> Json {
-    value.clone()
+fn observation_content_value(value: &Json) -> Option<Json> {
+    (!value.is_null()).then(|| value.clone())
 }
 
 fn is_atif_content_parts(value: &Json) -> bool {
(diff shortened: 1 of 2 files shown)


bbednarski9:
I ran an end to end validation on this - while the PR seems to be fixing the situation where a tool output arrives before the LLM tool-call declaration, im still seeing some duplicate tool calls in the hermes atif shape. Here's what codex thinks is happening:

Hermes hook-forward emits enough projected events that the ATIF exporter creates one agent step with the tool call and no observation, then another agent step with the same tool call plus the observation. Example from the tool-call run:

  - step 40: tool_calls=[terminal call_v31o81gd], no observation
  - step 41: same terminal call_v31o81gd, with observation result linked to call_v31o81gd

So PR #171 fixed correlation, but not deduplication/projection cleanup.

bbednarski9:
Also seeing a potential issue with streaming chunk aggregation from ATOF -> ATIF. Now might be the time to fix it. Will follow-up after running a more isolated end to end test

bbednarski9 (approved):
I tested this end to end and the issues are resolved. Created issue #176 to resolve the duplicate step issue that was observed during this reproduction. But it could introduce scope creep to this PR so not going to handle it here

PR #196, opened 2026-05-31
fix: pass a real ScopeHandle to Node withScope callbacks

zhongxuanwang-nv:
#### Overview

`withScope` in the Node.js binding handed its callback a plain JSON object (`{ uuid, name, scopeType, ... }`) instead of a real `ScopeHandle`. Passing that object back into `event`, `toolCallExecute`, or `llmCallExecute` (which accept `null` or a `ScopeHandle`) threw `internal error: Failed to recover ScopeHandle type from napi value` on the first call, and because the rejection skipped `deregisterSubscriber`, the process hung. This fixes the binding so the callback receives a real, reusable `ScopeHandle`, matching the Rust, Python, and WebAssembly bindings where the scope helper hands back a usable handle. The existing Node quick start and `nemo-relay-node` README examples already pass the handle, so they now run as written with no doc changes.

#### Details

Root cause: Node's `with_scope` invokes the callback across a `ThreadsafeFunction` boundary, and a `napi_wrap`'d `ScopeHandle` cannot cross that boundary, so the previous code serialized the handle to a plain JSON object. That object carries no wrapped native handle, so `napi_unwrap` fails when it is later passed to an API typed `Option<&ScopeHandle>`.

Changes:

- `crates/node/src/promise_call.rs`: add an `Arg0Builder` (`Box<dyn FnOnce(&Env) -> napi::Result<JsUnknown> + Send>`) and a `call_with_arg0` method so the first callback argument can be materialized on the JS thread inside the threadsafe call. The existing `call` / `call_with_json_next` / `call_with_stream_next` JSON paths are unchanged.
- `crates/node/src/api/mod.rs`: `with_scope` now builds a real `ScopeHandle` instance through that builder (`ToNapiValue::to_napi_value`) instead of `serde_json::json!({...})`, so the callback receives a handle it can pass straight into `event`, `toolCallExecute`, and `llmCallExecute`. The auto-pop and error paths are unchanged.
- `crates/node/tests/scope_tests.mjs`: add `callback receives a reusable ScopeHandle`, which passes the callback handle into `event`, `pushScope` (as an explicit parent), `toolCallExecute`, and `llmCallExecute` and asserts the results.

This aligns Node with the existing cross-binding behavior and does not change the shared runtime contract or the other bindings. No documentation changes are required: the quick start and README already pass the handle, and `LlmRequest` is delivered to `llmCallExecute` correctly, so those examples run unchanged once the binding is fixed.

Validation:

- `just build-node` and `just test-node`: 226/226 pass, including the new test.
- Ran the existing Node quick start end to end against the fixed build: it prints the scope/mark/tool/LLM event lines, `{ echo: 'hello' }`, and the LLM result, then exits `0` with no hang (previously it threw on the first `event(...)` call and hung).
- `cargo fmt --all`, `cargo clippy -p nemo-relay-node --all-targets -- -D warnings`, and `uv run pre-commit` pass.

Start with `with_scope` in `crates/node/src/api/mod.rs`: the callback argument is now a real `ScopeHandle` built on the JS thread through the new `Arg0Builder` path in `crates/node/src/promise_call.rs`. Cross-check `crates/node/tests/scope_tests.mjs` (`callback receives a reusable ScopeHandle`), which reuses the handle across `event`, `pushScope`, `toolCallExecute`, and `llmCallExecute` — the calls that previously threw.

- Relates to: none

### DIFF
--- a/docs/getting-started/quick-start/nodejs.mdx
+++ b/docs/getting-started/quick-start/nodejs.mdx
@@ -44,6 +44,7 @@ const {
   registerSubscriber,
   deregisterSubscriber,
   flushSubscribers,
+  LlmRequest,
   withScope,
   event,
   toolCallExecute,
@@ -70,7 +71,7 @@ async function main() {
 
     const llmResult = await llmCallExecute(
       "demo-provider",
-      { headers: {}, content: { messages: [{ role: "user", content: "hi" }] } },
+      new LlmRequest({}, { messages: [{ role: "user", content: "hi" }] }),
       (request) => ({ ok: true, messages: request.content.messages }),
       handle,
       null,
--- a/docs/getting-started/quick-start/nodejs.mdx
+++ b/docs/getting-started/quick-start/nodejs.mdx
@@ -44,7 +44,6 @@ const {
   registerSubscriber,
   deregisterSubscriber,
   flushSubscribers,
-  LlmRequest,
   withScope,
   event,
   toolCallExecute,
@@ -57,13 +56,15 @@ async function main() {
   });
 
   await withScope("demo-agent", ScopeType.Agent, async (handle) => {
-    event("initialized", handle, { binding: "node" }, null);
+    // `handle` describes the active scope (handle.uuid, handle.name, handle.scopeType).
+    // For lifecycle calls inside the scope, pass null to target the current scope.
+    event("initialized", null, { binding: "node" }, null);
 
     const toolResult = await toolCallExecute(
       "search",
       { query: "hello" },
       (args) => ({ echo: args.query }),
-      handle,
+      null,
       null,
       null,
       null,
@@ -71,9 +72,9 @@ async function main() {
 
     const llmResult = await llmCallExecute(
       "demo-provider",
-      new LlmRequest({}, { messages: [{ role: "user", content: "hi" }] }),
+      { headers: {}, content: { messages: [{ role: "user", content: "hi" }] } },
       (request) => ({ ok: true, messages: request.content.messages }),
-      handle,
+      null,
       null,
       null,
       null,
--- a/docs/getting-started/quick-start/nodejs.mdx
+++ b/docs/getting-started/quick-start/nodejs.mdx
@@ -56,15 +56,13 @@ async function main() {
   });
 
   await withScope("demo-agent", ScopeType.Agent, async (handle) => {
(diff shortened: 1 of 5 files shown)


willkill07 (changes requested):
Manually specifying a handle should work though. What was the raised panic?

zhongxuanwang-nv:
> Manually specifying a handle should work though. What was the raised panic?

@willkill07  Hey Will, yes I think manually specifying a handle does work; here it's mainly a rejected Promise / thrown napi error; I should've included the error output:

```
(nemo-relay) daniewang@DKF2M99Y76 1 % node 1.js
event=scope name=demo-agent
event=scope name=demo-agent
[Error: internal error: Failed to recover `ScopeHandle` type from napi value] {
  code: 'GenericFailure'
}
```
And then it's stuck there forever.

It's because `withScope` gives its callback a plain JSON object (`{ uuid, name, scopeType, ... }`) across the ThreadsafeFunction thread boundary, not a `napi_wrap`'s `ScopeHandle`. So when that object hits `event`/`toolCallExecute`/`llmCallExecute` (`Option<&ScopeHandle>`), `napi_unwrap` returns `invalid_arg` → the error above. It throws on the first such call (`event("initialized", handle, ...)`), which skips the following `deregisterSubscriber` and leaves the subscriber hang.

Therefore, I made `withScope` pass a real `ScopeHandle`. I'm wondering what do you think?

willkill07:
@zhongxuanwang-nv i think this PR should also fix that bug -- in other languages, we do get a scope handle back.

zhongxuanwang-nv:
> @zhongxuanwang-nv i think this PR should also fix that bug -- in other languages, we do get a scope handle back.

Thanks Will! I think the new proposed changes fixed that to make it consistent with other languages :)

docs/getting-started/quick-start/nodejs.mdx
willkill07, line 1: Does LlmRequest not work?

willkill07 (dismissed):
Approving with one open question

docs/getting-started/quick-start/nodejs.mdx
zhongxuanwang-nv: Actually, LlmRequest does work! With the new fix, we can bring it back. Thank you Will!

willkill07 (approved):

PR #199, opened 2026-06-01
feat: layer code-driven plugin config

zhongxuanwang-nv:
#### Overview

Layers code-driven plugin configuration through each binding's `plugin.initialize(...)` path over materialized file-backed `plugins.toml` configuration, while preserving deterministic file discovery and clear source reporting.

#### Details

- Added shared core resolution for discovered plugin config files plus an optional code-driven `plugin.initialize(...)` overlay.
- Preserved file-backed discovery from system, project, and user `plugins.toml` locations, with code config layered at higher precedence when supplied to binding initialization.
- Removed the public `.layer(...)` helper surfaces and the ad-hoc `--plugin-config` / `x-nemo-relay-plugin-config` paths so callers use config files plus binding initialization instead.
- Kept direct materialized-config initialization APIs available for callers that already have a complete `PluginConfig`.
- Added clearer effective-source/error reporting for plugin config activation and doctor/dry-run output.
- Clarified that layered configs reject duplicate component `kind` values in either layer because matching repeated plugin instances by `kind` alone is ambiguous; fully materialized configs can still contain repeated kinds when the plugin supports multiple components.
- Added tests across core, CLI, FFI, Python, Node.js, Go, and WebAssembly surfaces, including Node coverage for `plugin.initialize()` with no arguments loading project `plugins.toml`.
- Updated plugin configuration docs and agent skill guidance to describe discovery, precedence, file-source conflicts, quick testing, and code-driven overlays.

Quick functional check:

```toml
# .nemo-relay/plugins.toml
version = 1

[[components]]
kind = "observability"
enabled = false

[components.config.atof]
enabled = true
filename = "events.jsonl"
```

```python
from nemo_relay import plugin

await plugin.initialize({
    "components": [{
        "kind": "observability",
        "config": {
            "atof": {
                "mode": "overwrite",
            },
        },
    }],
})
```

Effective behavior: NeMo Relay discovers `.nemo-relay/plugins.toml`, keeps `enabled = false`, `atof.enabled = true`, and `filename = "events.jsonl"` from the file, then applies `mode = "overwrite"` from the code-driven overlay. If either layer contains duplicate `kind` entries, initialization fails instead of guessing which instance to update.

Validation:

- `cargo fmt --all`
- `cargo test -p nemo-relay`
- `cargo test -p nemo-relay-adaptive`
- `cargo test -p nemo-relay-ffi`
- CLI integration and targeted CLI plugin/session tests
- `just test-python`
- `just test-node`
- `just test-go`
- `cargo test -p nemo-relay-wasm`
- `just docs`
- `SKIP=attributions-rust uv run pre-commit run --all-files`

Local environment notes: `just test-wasm` is blocked on this machine because the Homebrew Rust sysroot is missing the `wasm32-unknown-unknown` target, and the Rust attribution pre-commit hook is skipped locally because `cargo-about` is not installed.

Start with `crates/core/src/plugin.rs` for discovered config resolution, overlay semantics, and duplicate-kind rejection. Then review `crates/node/tests/plugin_tests.mjs` for the no-argument `plugin.initialize()` coverage and `docs/build-plugins/plugin-configuration-files.mdx` for the public behavior description.

- Relates to: RELAY-183 (https://linear.app/nvidia/issue/RELAY-183/layer-code-driven-plugin-config-over-materialized-file-configs)

## Summary by CodeRabbit

* **New Features**
  * Automatic discovery and layered merging of file-backed plugin config; initialize() across bindings accepts an optional overlay.
  * Dry-run and live status now show the resolved plugin configuration and which file/source provided it.
  * Diagnostics/doctor output includes the effective plugin-config source.

* **Breaking Changes**
  * Removed the CLI --plugin-config flag and the corresponding relay header.

* **Documentation**
  * Expanded plugin config docs with discovery, precedence, layering, and examples.


willkill07 (changes requested):
code-driven plugin configuration is accomplished through each language binding's plugin initialize function.

1. I do not see the "base" layer being inferred automatically from user/project/global config (which was the entire point of layering).
2. Providing per-language binding layering is nice, but I believe we can drop `--plugin-config` outright. Folks should be using the configuration files rather than ad-hoc.
3. The added/improved error output is good!

willkill07 (changes requested):
There shouldn't be any API changes required.

We just want to augment the behavior of `initialize_plugins` to have a base layer applied.

- There should be no language binding changes
- Tests should be updated based on behavioral changes
- There should be no public API changes
- There should be no changes to internal structs/datatypes

It might be easier to start from a clean slate than with the current changes.

PR #200, opened 2026-06-02
docs: register plugin kind before activating it in the plugin guide

zhongxuanwang-nv:
#### Overview

Reorder the "Register Plugin Behavior" guide so a plugin kind is registered before it is activated, and document the `validate()`-warns / `initialize()`-raises behavior. Run on its own, the **Activation APIs** example previously failed with `RuntimeError: ... 'header-plugin' is not registered`, because the kind it activates is only registered later on the page in the **Header Plugin Example**.

#### Details

- `docs/build-plugins/register-behavior.mdx`: move the **Header Plugin Example** (which calls `plugin.register("header-plugin", ...)`) above the **Activation APIs** lifecycle so the page reads and runs top to bottom. Drop the now-redundant "Register the plugin kind" item from the activation list, and note that an unregistered kind is reported by `validate()` only as a warning under the default `unknown_component="warn"` policy, while `initialize()` raises. The reorder moves whole sections only; no code-block content changed (verified by an order-independent line diff).
- `docs/build-plugins/validate-configuration.mdx`: add the same clarification to **Validate Before Initialization** — the error-only `has_errors` check does not catch an unknown/unregistered kind; register kinds first, or set `unknown_component="error"`.

Heads-up: this edits the same `register-behavior.mdx` **Activation APIs** section as #199 (which adds `plugin.layer(...)` to that example but does not change the registration ordering). The two overlap, so whichever lands second will need a small conflict resolution.

`docs/build-plugins/register-behavior.mdx` — confirm the section reorder preserved every code block (an order-independent line diff shows only the numbered list and intro changed) and that the Activation APIs example now follows the Header Plugin Example that registers the kind.

Verified locally: running the Header Plugin Example block and then the Activation APIs block in one process registers `header-plugin`, after which `validate()` / `initialize()` / `clear()` succeed with empty diagnostics and no `RuntimeError`. Targeted `pre-commit` (trailing whitespace, end-of-files, docs markdown linkcheck) passes on both files.

- Relates to: #199 (edits the same `register-behavior.mdx` section)

## Summary by CodeRabbit

* **Documentation**
  * Clarified that plugin kinds must be registered before calling initialize()
  * Explained that unknown/unregistered component kinds are reported as warnings by default during validation and may still cause errors at initialization unless policy is changed
  * Reorganized and updated the activation guide and language across Python/Node/Rust examples for clearer ordering and guidance

### DIFF
--- a/docs/build-plugins/validate-configuration.mdx
+++ b/docs/build-plugins/validate-configuration.mdx
@@ -126,6 +126,14 @@ Prefer stable diagnostic codes over prose-only messages. The message can improve
 
 Use the validation API before initialization and fail deployment if the report contains errors.
 
+<Warning>
+This error check does not catch an unknown or unregistered component kind. Under
+the default `unknown_component="warn"` policy that case is reported as a warning,
+not an error, so the check below passes while `initialize()` still raises for a
+kind that is not registered. Register every kind before initialization, or set
+`unknown_component="error"` to make validation fail on unknown kinds.
+</Warning>
+
 <Tabs>
 <Tab title="Python" language="python">
 ```python
--- a/docs/build-plugins/register-behavior.mdx
+++ b/docs/build-plugins/register-behavior.mdx
@@ -25,82 +25,6 @@ That gives the plugin system three important guarantees:
 
 Use the context only after validation succeeds. Validation should inspect config and return diagnostics; registration should create runtime objects and attach them to the context.
 
-## Activation APIs
-
-Use the plugin APIs in this order:
-
-1. Register the plugin kind.
-2. Build a `PluginConfig`.
-3. Validate the config.
-4. Initialize the config.
-5. Inspect the activation report.
-6. Clear active config during teardown when needed.
-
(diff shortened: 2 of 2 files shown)


docs/build-plugins/validate-configuration.mdx
```diff
@@ -126,6 +126,8 @@ Prefer stable diagnostic codes over prose-only messages. The message can improve
 
 Use the validation API before initialization and fail deployment if the report contains errors.
 
+This error check does not catch an unknown or unregistered component kind. Under the default `unknown_component="warn"` policy that case is reported as a warning, not an error, so the check below passes while `initialize()` still raises for a kind that is not registered. Register every kind before initialization, or set `unknown_component="error"` to make validation fail on unknown kinds.
```
willkill07: Maybe surround this with `<Warn>`/`</Warn>`

willkill07 (dismissed):

willkill07 (approved):

PR #201, opened 2026-06-02
docs: flush subscribers before deregister in package README examples

zhongxuanwang-nv:
#### Overview

The package READMEs' "Getting Started" examples register a subscriber, emit events, then deregister without calling flush. Native subscriber delivery is non-blocking, so an example that exits right after deregistering can lose queued events. This updates all four package READMEs to flush before deregister, matching the quickstarts (which already flush).

#### Details

**Functional fixes** (verified by running — the example dropped events without flush):
- `python/nemo_relay/README.md`: printed nothing, or intermittently a single partial line; the `to_dict()` / `to_json()` callback never ran. Adds `nemo_relay.subscribers.flush()` before each `deregister(...)`.
- `go/nemo_relay/README.md`: flaky — dropped the trailing tool-end and scope-end events on exit (2 of 3 runs partial). Adds `defer nemo.FlushSubscribers()`, ordered to run after the deferred `scope.Pop` emits the scope-end event and before the deferred deregister.

**Parity** (not currently broken; updated so every README models the same pattern):
- `crates/node/README.md`: Node keeps the event loop alive via the ref'd ThreadsafeFunction, so it already delivers; adds `flushSubscribers()` for consistency.
- `crates/wasm/README.md`: WASM's `flushSubscribers()` is a single-threaded no-op (tested as such); adds it for consistency.

The four README diffs — each adds a flush before `deregister(...)` (Node/WASM also import `flushSubscribers`). Verified by running each example with the flush added: events deliver deterministically (Go 5/5, Python and Node 3/3 across repeated runs). WASM's `flushSubscribers` is an exported no-op (`crates/wasm/src/api/mod.rs`; covered by `crates/wasm/tests-js/scope_tests.mjs`).

Note: the `crates/node/README.md` example also depends on the Node binding change in #196 (pass a real `ScopeHandle` to `withScope` callbacks). That is orthogonal to this flush change — #196 is code-only and does not touch the README.

- Relates to: #196 (Node `withScope` ScopeHandle; orthogonal, touches the same Node README example)

## Summary by CodeRabbit

* **Documentation**
  * Updated Getting Started examples across Python, Node.js, WebAssembly, and Go language bindings to reflect the proper sequence for flushing subscribers before deregistration.

### DIFF
--- a/crates/node/README.md
+++ b/crates/node/README.md
@@ -66,6 +66,7 @@ const {
   ScopeType,
   deregisterSubscriber,
   event,
+  flushSubscribers,
   registerSubscriber,
   withScope,
 } = require("nemo-relay-node");
@@ -80,6 +81,7 @@ async function main() {
     event("initialized", handle, { binding: "node" }, null);
   });
 
+  flushSubscribers();
   deregisterSubscriber("printer");
 }
 
--- a/crates/wasm/README.md
+++ b/crates/wasm/README.md
@@ -83,6 +83,7 @@ const {
   ScopeType,
   deregisterSubscriber,
   event,
+  flushSubscribers,
   registerSubscriber,
   withScope,
 } = require("nemo-relay-wasm");
@@ -97,6 +98,7 @@ async function main() {
     event("initialized", handle, { binding: "wasm" }, null);
   });
 
+  flushSubscribers();
   deregisterSubscriber("printer");
 }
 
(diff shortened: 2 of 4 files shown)


zhongxuanwang-nv:
I'm sorry about putting this PR ready and then immediately removing this PR's ready status...

willkill07 (approved):

PR #205, opened 2026-06-03
feat: Improve Hermes Relay hook injection and event alignment

yczhang-nv:
#### Overview

Adds Relay-side support needed for NVIDIA NeMo Agent Toolkit to consume Hermes Relay telemetry.

This change:
- Updates transparent Hermes Relay runs to temporarily patch Hermes hook config, route events to the local Relay gateway, and restore the original config afterward.
- Improves Hermes session alignment by correlating task-scoped tool events back to the owning Hermes session.
- Handles Hermes pre/post tool hook ID mismatches by matching active tools by name and arguments when provider tool IDs differ.

## Why

Hermes invokes hooks from its own config and can execute through a different process boundary than the caller. For NeMo Agent Toolkit integration, Relay needs to capture Hermes lifecycle events reliably and make them available as replayable ATOF JSONL so the toolkit can import them into its telemetry stream.

- Closes #

## Summary by CodeRabbit

* **New Features**
  * Hermes hook lifecycle: backup, patch, restore (dry-run supported); restores on failed runs/health checks.
  * Improved tool tracking retains tool name, arguments, and owner context for better correlation.
  * Better LLM/response parsing and span de-duplication for observability exports.

* **Bug Fixes**
  * Hermes task→session routing and lifecycle cleanup corrected to avoid duplicate sessions and stale mappings.
  * More robust tool-end resolution with owner-context matching.

* **Tests**
  * New unit/integration tests covering Hermes hooks, routing, tool correlation, and observability dedupe.

* **Documentation**
  * Readme, CLI, Hermes, Cursor, integrations, and release notes updated; Node/WebAssembly/Python examples call out subscriber flush and Node.js 24+ requirement.


python/nemo_relay/nat_exporter.py
```diff
@@ -0,0 +1,79 @@
...
+
+import json
+import threading
+from pathlib import Path
+from typing import Literal
+
+from nemo_relay import subscribers
+
+
+class NatTelemetryExporter:
```
willkill07: We already have an AtofExporter with a configurable filename. Why does that not work?

willkill07 (dismissed):
The improvements are more representative than the added exporter. Could the PR title be updated accordingly?

Otherwise, my only feedback is related to the added exporter and what value it adds over the existing AtofExporter.

Conditionally approving.

python/nemo_relay/nat_exporter.py
yczhang-nv: Yeah, `nat_exporter` seems redundant code as `ATOFExporter` is sufficient here. Removed `nat_exporter`.

willkill07 (changes requested):
The commit history and changes with your rebase are causing issues that should not be present. Please rewrite the history to *only* contain your commits. There are changes which are out-of-scope and/or outdated.

yczhang-nv:
> The commit history and changes with your rebase are causing issues that should not be present. Please rewrite the history to _only_ contain your commits. There are changes which are out-of-scope and/or outdated.

Thanks Will! Just cleaned up the PR and I'll appreciate another review

crates/cli/src/alignment/mod.rs
```diff
@@ -110,13 +110,15 @@ pub(crate) struct SessionAlignmentState {
     aliases: HashMap<String, SessionAlias>,
     completed_aliases: HashMap<String, SessionAlias>,
     pending_subagents: HashMap<String, PendingSubagentStart>,
+    hermes_task_sessions: HashMap<String, HashMap<String, String>>,
```
willkill07: could we generalize to `task_sessions` or `pending_task_sessions` ?

crates/cli/src/alignment/mod.rs
willkill07, line 1: Everything `hermes_` specific in this file should probably be generalized

willkill07 (dismissed):
logic looks good! let's just relax the naming a bit to be more generic.

crates/cli/src/alignment/mod.rs
yczhang-nv: done!

yczhang-nv:
The naming has been generalized. @willkill07 could you re-approve when you get a chance?

willkill07 (approved):

PR #210, opened 2026-06-04
feat: add nested subagent session lineage for OpenClaw

mnajafian-nv:
#### Overview

This PR adds nested subagent session lineage for OpenClaw so child subagent activity can materialize under the requester session when stable lineage is available, while keeping replay state and emitted metadata honest when hooks arrive out of order.

#### Details

- Adds OpenClaw subagent lineage tracking in the hook replay backend so child sessions can materialize as nested subagent scopes under the requester session when `subagent_spawned` provides stable parent/child lineage.
- Introduces stable internal session ownership for replay state and alias resolution, so deferred child sessions no longer depend on mutable exported `sessionId` values.
- Keeps spawn-first child sessions deferred until a requester root is available, then promotes them under the requester scope instead of materializing them as top-level sessions.
- Preserves original timestamps and queued lifecycle emissions when child session activity arrives before lineage can be resolved.
- Preserves nested lineage when a spawn-first child session closes before the requester `session_start` arrives.
- Materializes child-only `subagent_ended` paths under the child session scope instead of dropping or misparenting the terminal mark.
- Keeps the change scoped to the OpenClaw adapter layer without changing ATOF/ATIF exporter behavior.
- Adds focused coverage for the main subagent ordering cases, including spawn-first, child-first, child close before requester start, child-only terminal paths, and deferred model timing followed by explicit child `session_start`.
- Updates the OpenClaw integration docs to describe nested child-session behavior when stable lineage is available.

Start in `integrations/openclaw/src/hooks-backend.ts`, especially the subagent lineage tracking, requester-anchor handling, and deferred child-root materialization flow. Then review `integrations/openclaw/test/hooks-backend.test.ts` for the contract-level ordering cases covered here. `integrations/openclaw/src/hook-replay/session.ts` is supporting context for the stable internal session ownership model used by the replay state.

- Relates to OpenClaw observability consistency work.

### DIFF
--- a/docs/supported-integrations/openclaw-plugin.mdx
+++ b/docs/supported-integrations/openclaw-plugin.mdx
@@ -268,7 +268,7 @@ features such as adaptive hints require a managed execution path.
 | `after_tool_call` | Replays successful tool calls as tool spans; blocked tools emit marks. |
 | `agent_end` | Emits an agent lifecycle mark, flushes recorded assistant-turn LLM spans, and preserves the final assistant answer as the session output. |
 | `before_agent_finalize` | Preserves the last assistant message as fallback session output and emits a lifecycle mark without mutating the finalization payload. |
-| `subagent_spawned` / `subagent_ended` | Emits subagent lifecycle marks under the best available parent or child session. |
+| `subagent_spawned` / `subagent_ended` | Emits subagent lifecycle marks and nests child subagent session scopes under the requester session when stable lineage is available. |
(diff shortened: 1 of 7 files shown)


willkill07 (approved):
This logic seems similar to what is done in the CLI gateway. Makes sense to me.

We should add to the request with OpenClaw folks about added information with subagent hooks so we don't need to external guess/resolve.

PR #211, opened 2026-06-04
feat: layer code-driven plugin config over materialized file configs

zhongxuanwang-nv:
#### Overview

Layer code-driven plugin config on top of the materialized file-based `plugins.toml` configs (RELAY-183). A direct integration that calls `plugin.initialize(...)` now resolves the same `plugins.toml` layering the gateway uses as its base instead of starting empty; the code-driven config is the authoritative top layer.

#### Details

- `initialize_plugins(config)` resolves the discovered `plugins.toml` layering (system → project → user) as a base and layers `config` on top, so every binding inherits the file layering with **no binding code changes**.
- `initialize_plugins_exact(config)` activates an already-materialized config without discovery; the gateway uses it so `--config` / `--plugin-config` source rules are unchanged.
- File discovery + TOML parsing live in core: `toml` is a direct dependency and `load_plugin_config_files` (`#[doc(hidden)]`) is the shared read/parse/merge primitive the CLI delegates to.
- Trade-off: `toml` + the filesystem discovery compile into every core build, including wasm (where discovery finds no files at runtime, so the base is empty).

**Precedence — code-vs-file differs from file-vs-file.** File-vs-file uses `omit = inherit`. Code-vs-file treats the code document as authoritative: you inherit any component kind or `config` key you omit, but the typed fields (`version`, `policy`, and a declared component's `enabled`) always take the code value (serde defaults included) — because bindings deserialize to the typed `PluginConfig` before core sees it, so "set" vs "defaulted" is already lost.

#### Constraints

Kept:
- No struct/datatype changes — `PluginConfig` / `PluginComponentSpec` / `ConfigPolicy` untouched.
- No public binding API changes — `plugin.initialize(...)` signatures unchanged across bindings.

Relaxed (with reason):
- *No public API changes* — two `pub fn`s added (`initialize_plugins_exact`, `load_plugin_config_files`), both `#[doc(hidden)]` for cross-crate gateway use; the gateway needs a no-discovery path distinct from the now-discovering `initialize_plugins`.

#### Tests

Core unit tests for `load_plugin_config_files` (merge precedence, per-file duplicate-kind rejection, absent files); existing CLI file-discovery tests exercise the same shared primitive.

`crates/core/src/plugin.rs` — `initialize_plugins` (base-layer + overlay), `initialize_plugins_exact`, and the discovery / `load_plugin_config_files`. The precedence decision is the key thing to confirm.

- Relates to RELAY-183

## Release Notes

* **New Features**
  * Plugin configuration now supports layering across multiple sources with clear precedence rules, allowing configuration files, inline settings, and code-driven overrides to be combined effectively.
  * Improved handling of multi-instance component configurations.

* **Documentation**
  * Updated plugin configuration documentation to explain the layering mechanism and configuration discovery process.


zhongxuanwang-nv:
@coderabbitai review

docs/build-plugins/plugin-configuration-files.mdx
```diff
@@ -70,18 +70,27 @@ The gateway reads only files named `plugins.toml`.
...
 | `[plugins].config` in `config.toml` | Inline gateway config for small or generated setups. |
 | `--plugin-config '<json>'` | CI, tests, wrappers, or one-off automation. |
 
-Use only one source class for a given gateway run. The gateway fails clearly if
-file-based plugin config and `--plugin-config` are both present, or if
+Use only one file/CLI source class for a given gateway run. The gateway fails
+clearly if file-based plugin config and `--plugin-config` are both present, or if
 `plugins.toml` and `[plugins].config` are both present.
 
+A code-driven layer is a separate, programmatic source set through the runtime
```
willkill07: I don't think this short paragraph adds value.

docs/build-plugins/plugin-configuration-files.mdx
```diff
@@ -155,7 +164,9 @@ menu to reset, clear, preview, or save.
 
 When more than one `plugins.toml` file is discovered, later files have higher
 precedence. User config overrides project config, and project config overrides
-system config.
+system config. A code-driven layer, when set, sits above all of these as the
```
willkill07: Same here

docs/build-plugins/plugin-configuration-files.mdx
```diff
@@ -195,6 +206,70 @@ components that reach plugin validation also fail validation.
 Arrays inside component config are replaced by the higher-precedence value.
 Tables inside component config merge recursively.
 
+## Code-Driven Configuration Layer
```
willkill07: This should NOT be exposed as a public API.

The layering should be described but it's not something that should be used invoked. Plugin `initialize` should be the only way it's done.

docs/build-plugins/plugin-configuration-files.mdx
```diff
@@ -195,6 +206,70 @@ components that reach plugin validation also fail validation.
...
+The gateway fails fast, naming the sources, in these conflict cases:
+
+| Situation | Result |
+|---|---|
+| `--plugin-config` and any file config (`plugins.toml` or `[plugins].config`) | Error: choose one source |
+| `plugins.toml` and `[plugins].config` both present | Error: choose one source |
+| `[plugins].config` defined in more than one `config.toml` file | Error: consolidate into one block |
+| Code-driven layer set together with any file/CLI source | No conflict — the layer overrides the file/CLI config field-by-field |
+
+The code-driven layer is process-local: it only affects `initialize_plugins`
```
willkill07: Remove this paragraph

docs/build-plugins/plugin-configuration-files.mdx
```diff
@@ -195,6 +206,70 @@ components that reach plugin validation also fail validation.
 Arrays inside component config are replaced by the higher-precedence value.
 Tables inside component config merge recursively.
 
+## Code-Driven Configuration Layer
+
+A host application can supply a *code-driven* plugin configuration layer
```
willkill07: Just phrase this (and the following section) as configuration layering. Code-driven configuration is already established.

crates/cli/src/doctor.rs
willkill07, line 1: I'm a little confused by the file changes present here.

The feature of the PR is to simply layer programmatic configuration on top of file-resolved.

`doctor` has no way of knowing what could be added programmatically. So why was this changed at all?

crates/core/src/plugin.rs
```diff
@@ -899,6 +901,96 @@ pub fn validate_plugin_config(config: &PluginConfig) -> ConfigReport {
     report
 }
 
+/// Sets (or clears with `None`) the process-global code-driven plugin layer.
+///
+/// The layer is the highest-precedence source: [`initialize_plugins`] merges it on
+/// top of the config it is given. This Rust-only API is for embedding hosts and is
+/// not exposed to the language bindings.
+pub fn set_code_driven_plugin_config(config: Option<PluginConfig>) {
```
willkill07: No public API changes

crates/core/src/plugin.rs
```diff
@@ -899,6 +901,96 @@ pub fn validate_plugin_config(config: &PluginConfig) -> ConfigReport {
...
+    let mut guard = PLUGIN_CONFIG_CODE_DRIVEN_LAYER
+        .lock()
+        .unwrap_or_else(|err| err.into_inner());
+    *guard = config;
+}
+
+/// Returns the code-driven layer merged over `config`, or [`None`] when no layer
+/// is set (so `config` is already effective). `Some`/`None` also tells callers
+/// whether a layer contributed.
+pub fn apply_code_driven_plugin_config(config: &PluginConfig) -> Option<PluginConfig> {
```
willkill07: No public API changes

crates/core/src/plugin.rs
willkill07, line 1: The CLI already has config merge logic.

I believe we shouldn't duplicate code so let's:
- move the logic for merging over to here
- make the minimal API set public (since it has to be accessible outside of the crate). I claim this should just be a single `layer_config` function. I forget the exact name used.
- update the CLI to call the moved logic where appropriate.

willkill07 (changes requested):

zhongxuanwang-nv:
@coderabbitai review

crates/cli/src/config.rs
```diff
@@ -759,16 +760,24 @@ where
                     ))
                 })?;
             validate_plugin_toml_component_kinds(&path, &parsed)?;
-            merge_plugin_toml(&mut merged, parsed);
+            // Merge file layers with the shared core primitive (each file converted to JSON first).
```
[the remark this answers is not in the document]
zhongxuanwang-nv: will resolve

willkill07:
This is the core required feature:

Right now, if a user has direct integration and calls `plugin.initialize(...)` for the appropriate language binding, they _only_ get configured what they explicitly configured.

We already have existing configuration layering for the CLI through various `plugin.toml` files.

*Need:* instead of the base configuration being empty, it should be resolved from pre-existing NeMo Relay plugin configuration layering.

zhongxuanwang-nv:
@coderabbitai review

zhongxuanwang-nv:
@willkill07 , sorry it took so long — I spent most of the time trying to decide where the toml discovery should belong and decide on the general architecture — but I;m wondering what do you think of the new changes? I also updated the PR body for this :)

crates/core/src/plugin.rs
willkill07, line 1: A lot of the logic in this file is already present in `crates/cli/src/config.rs`:
- layering
- resolving locations of plugin configuration files
- helpers for resolving global/project/user locations

These helpers should:
- live in core
- be accessed under `cli`

crates/cli/src/server.rs
willkill07, line 1: Why are we switching from `initialize_plugins` to `initialize_plugins_exact`? They have the same API shape so I'd expect it to work.

`initialize_plugins_exact` should not be `pub`

willkill07 (changes requested):
This is much better! Minor feedback related to structure and logic duplication

crates/core/src/plugin.rs
zhongxuanwang-nv: Thanks Will!! That's true.. I mistakenly prioritized on not adding too many pub fn, but indeed duplicating code is a much bigger problem.

crates/cli/src/server.rs
zhongxuanwang-nv: Hey Will, `initialize_plugins` will read the filesystem (reading the default `plugins.toml` file and then merging, activating), but `initialize_plugins_exact` doesn't read the filesystem and only activates exactly the `config`. 

We are separating because `nemo-relay-cli` already resolved its config under rules the core can't see. If it called the `initialize_plugins`, core would then re-discover the default `plugins.toml` file so `--plugin-config` would silently activate the default `plugins.toml` instead, and since we don't want to change the language bindings, I just added the `initialize_plugins_exact`.

And since `nemo-relay-cli` needs `initialize_plugins_exact`, it's now pub but it's also not discoverable by doc.

I'm wondering what do you think?

willkill07:
@zhongxuanwang-nv i noticed your new commit! This looks good.

The only issue now is updating this branch with main -- there's logic with cost layering that needs to be correctly handled

willkill07 (dismissed):

docs/build-plugins/plugin-configuration-files.mdx
```diff
@@ -70,16 +70,19 @@ The gateway reads only files named `plugins.toml`.
 
 ## Discovery
 
-The gateway can receive plugin configuration from three source classes:
+The gateway resolves plugin configuration from two kinds of input: file/CLI
```
lvojtku:
```suggestion
The gateway resolves plugin configuration using two kinds of input: file/CLI
```

docs/build-plugins/plugin-configuration-files.mdx
```diff
@@ -201,6 +204,16 @@ components that reach plugin validation also fail validation.
 Arrays inside component config are replaced by the higher-precedence value.
 Tables inside component config merge recursively.
 
+## Configuration Layering
+
+A direct integration that calls `plugin.initialize(...)` (any binding) resolves the
+discovered `plugins.toml` layering as a base, then layers the code-driven config on
```
lvojtku:
```suggestion
discovered `plugins.toml` layering as a base. Then, it layers the code-driven config on
```

docs/build-plugins/plugin-configuration-files.mdx
```diff
@@ -201,6 +204,16 @@ components that reach plugin validation also fail validation.
...
+top (the code-driven config wins). Precedence differs slightly from the file-vs-file
```
lvojtku: I'm not really sure what "precedence" is referring to. Are we talking about the order of the layers? The rest of this paragraph is confusing to me. Recommendation:

_The order of layers are difference depending on the rules of each file. A code-driven config is a typed document, where `version`, `policy`, and a declared component's `enabled` take the code value (their defaults included). The omitted components and `config` keys are inherited from the files._

lvojtku: Can you explain to me the last sentence? I'm not sure what the targets are referring to.

lvojtku (changes requested):
There is a paragraph that requires rewriting. Please send me a message if you need help with rewrite. Thanks!

docs/build-plugins/plugin-configuration-files.mdx
zhongxuanwang-nv: 1. Thank you Linette for reviewing! Precedence here is which value would win when both sides (file / code) set the same thing, it's not about the order of the layers, because the order is always fixed (files first, then code on top). The only thing that differs from the previous file-vs-file section is that leaving a setting out means "inherit it" for a file but "use the default" for code
2. "Targets" here refer to the kind of system the software is built to run on. Some, like WebAssembly (wasm), don't run on a file system, so it doesn't have a plugins.toml file to find. The starting point is therefore empty and only the code settings apply.

I will also rewrite these!

zhongxuanwang-nv:
/ok to test [fce7ba0](https://github.com/NVIDIA/NeMo-Relay/pull/211/commits/fce7ba063104d47b287ae13a0eeb72c584ef5193)

docs/build-plugins/plugin-configuration-files.mdx
```diff
@@ -189,7 +192,7 @@ lower-precedence component. A component with a different `kind` is added to the
 effective configuration.
 
 The `pricing` component has one additional merge rule: when both lower- and
```
lvojtku:
```suggestion
The `pricing` component has one additional merge rule: when both lower and
```

docs/build-plugins/plugin-configuration-files.mdx
```diff
@@ -201,6 +204,25 @@ components that reach plugin validation also fail validation.
 Arrays inside component config are replaced by the higher-precedence value.
 Tables inside component config merge recursively.
 
+## Configuration Layering
+
+Plugin settings can come from two places, and they combine: the `plugins.toml`
```
lvojtku: - What are the 2 places? Files and code?
- How does this relate to configuration layering? Do the files and code represent specific layers? 
- Can this be a set of steps? For example, 1. The plugin.toml files are read and merged into a starting point (would user understand what starting point means?). 2. The code is the passed on as a settings applied on top of the files to resolve conflicts.

docs/build-plugins/plugin-configuration-files.mdx
```diff
@@ -201,6 +204,25 @@ components that reach plugin validation also fail validation.
...
+files are read and merged into a starting point, then settings passed in code to
+`plugin.initialize(...)` are applied on top and win on conflicts.
+
+Files and code differ in what happens to a setting you leave out. In a **file**,
```
lvojtku: Leave out where?

docs/build-plugins/plugin-configuration-files.mdx
```diff
@@ -70,16 +70,19 @@ The gateway reads only files named `plugins.toml`.
 
 ## Discovery
 
-The gateway can receive plugin configuration from three source classes:
+The gateway resolves plugin configuration using two kinds of input: file/CLI
```
bbednarski9: nit, but better docs practice to not say "file/CLI" as it looks like a path, lets say "file and CLI"

docs/build-plugins/plugin-configuration-files.mdx
```diff
@@ -70,16 +70,19 @@ The gateway reads only files named `plugins.toml`.
...
+sources and an optional code-driven layer.
+
+File/CLI configuration comes from one of three mutually exclusive source classes:
```
bbednarski9: same here

docs/build-plugins/plugin-configuration-files.mdx
```diff
@@ -70,16 +70,19 @@ The gateway reads only files named `plugins.toml`.
...
 
 | Source | Use case |
 |---|---|
 | `plugins.toml` | Normal operator- and project-managed gateway plugin configuration. |
 | `[plugins].config` in `config.toml` | Inline gateway config for small or generated setups. |
 | `--plugin-config '<json>'` | CI, tests, wrappers, or one-off automation. |
 
-Use only one source class for a given gateway run. The gateway fails clearly if
-file-based plugin config and `--plugin-config` are both present, or if
+Use only one file/CLI source class for a given gateway run. The gateway fails
```
bbednarski9: same here "docs or CLI"

docs/build-plugins/plugin-configuration-files.mdx
bbednarski9: this sentence is unclear.

bbednarski9 (changes requested):

docs/build-plugins/plugin-configuration-files.mdx
lvojtku: Plugin settings can come from files and the code. The files are the configuration base layer and the code sits on top of it. If there is a conflict the code will always take precedence over the files. The layering process is as follows:

1. The `plugin.toml` files are read and merged into a starting point (would user understand what starting point means?). 
2. The code is then passed on through `plugin.initialize(...)` as a setting applied on top of the files to resolve conflicts, if needed.

docs/build-plugins/plugin-configuration-files.mdx
```diff
@@ -201,6 +204,25 @@ components that reach plugin validation also fail validation.
...
+## Configuration Layering
+
+Plugin settings can come from two places, and they combine: the `plugins.toml`
+files are read and merged into a starting point, then settings passed in code to
+`plugin.initialize(...)` are applied on top and win on conflicts.
+
+Files and code differ in what happens to a setting you leave out. In a **file**,
+an omitted setting is simply absent, so a value from a lower-priority file shows
+through. In **code**, the config is a structured object whose fields always hold
+a value — their defaults if you didn't set one — so `version`, `policy`, and a
```
lvojtku:
```suggestion
a value — their defaults if you did not set one — so `version`, `policy`, and a
```

docs/build-plugins/plugin-configuration-files.mdx
```diff
@@ -201,6 +204,25 @@ components that reach plugin validation also fail validation.
...
+declared component's `enabled` always take the code value, even when it's just
+the default. The components you list and the keys inside each component's
+`config` still merge with the files, so anything you don't set in code is
+inherited.
+
+If the program runs without a filesystem — for example a WebAssembly (wasm)
```
lvojtku:
```suggestion
If the program runs without a filesystem — for example, a WebAssembly (wasm)
```

lvojtku (changes requested):

docs/build-plugins/plugin-configuration-files.mdx
```diff
@@ -201,6 +204,34 @@ components that reach plugin validation also fail validation.
...
+
+
+Files and code differ only in how they treat a setting you **omit**:
+| You omit | In a file | In code |
+|---|---|---|
+| `version`, `policy`, or the `enabled` flag of a component you declare | Inherited from a lower-precedence file | **Always taken from code** — its default if you did not set it |
+| A whole component kind, or a key inside a component's `config` | Inherited from a lower-precedence file | Inherited from the file layer |
+
+In other words, a file's gaps are filled by the layers beneath it, but code's
+typed fields are never gaps — they always carry a value, so they always win.
```
lvojtku:
```suggestion
typed fields are never gaps — they always carry a value, so they always take precedence.
```

docs/build-plugins/plugin-configuration-files.mdx
```diff
@@ -201,6 +204,34 @@ components that reach plugin validation also fail validation.
 Arrays inside component config are replaced by the higher-precedence value.
 Tables inside component config merge recursively.
 
+## Configuration Layering
+
+Plugin settings can come from files and the code. The files are the 
+configuration base layer and the code sits on top of it. If there is a 
+conflict the code will always take precedence over the files. The layering
```
lvojtku:
```suggestion
conflict, the code will always take precedence over the files. The layering 
```

lvojtku (dismissed):

lvojtku (approved):

willkill07 (approved):

zhongxuanwang-nv:
oops `/merge`

bbednarski9 (approved):

PR #212, opened 2026-06-04
refactor: drop process-level --plugin-config in favor of config files

zhongxuanwang-nv:
#### Overview

Drops the process-level `--plugin-config` ad-hoc CLI flag in favor of configuration files. Gateway plugin configuration should come from `plugins.toml` or an inline `[plugins].config` block, not one-off command-line JSON.

#### Details

- Removed `--plugin-config` (and the `NEMO_RELAY_PLUGIN_CONFIG` env var) from `nemo-relay serve` (`ServerArgs`) and `nemo-relay run` (`RunCommand`).
- Removed `apply_cli_plugin_config` and the run-vs-inherited precedence shim that only existed because both structs carried the flag.
- Docs: the gateway now resolves plugin configuration from two mutually exclusive file source classes (`plugins.toml` or `[plugins].config`); removed the `--plugin-config` row, conflict clause, and the "use `--plugin-config` for automation" guidance.
- Removed the now-obsolete `--plugin-config` tests and trimmed the flag from the remaining `run`/`serve` config tests.
- The per-session `hook-forward` / install `--plugin-config` (which sets the `x-nemo-relay-plugin-config` session header) is a separate mechanism and is unchanged.

**Breaking changes:** `--plugin-config` and `NEMO_RELAY_PLUGIN_CONFIG` are no longer accepted by `nemo-relay serve` or `nemo-relay run`; use `plugins.toml` or `[plugins].config` instead.

`crates/cli/src/config.rs` — the removal of the `ServerArgs`/`RunCommand` `plugin_config` fields, `apply_cli_plugin_config`, and the simplified `resolve_run_config`.

- Relates to relay-183 (internal tracking ticket; no public GitHub issue). Split out from the code-driven plugin config layer PR to keep that change scoped.

## Summary by CodeRabbit

* **Configuration Changes**
  * Removed `--plugin-config` CLI flag from daemon and run modes. Plugin configuration is now available only through `plugins.toml` or `[plugins].config` in the configuration file.

* **Documentation**
  * Updated plugin configuration documentation to reflect configuration source changes.


willkill07:
This PR conflicts with #211 -- without additional context I don't know whether this is still applicable.

zhongxuanwang-nv:
> This PR conflicts with #211 -- without additional context I don't know whether this is still applicable.

@willkill07 Hey Will, I might have misunderstood but you mentioned this in [PR-199](https://github.com/NVIDIA/NeMo-Relay/pull/199), since I thought this might be a bigger breaking change, I just initiated another PR here — I'm wondering if you still think we should drop ` --plugin-config `?

> Providing per-language binding layering is nice, but I believe we can drop --plugin-config outright. Folks should be using the configuration files rather than ad-hoc.

willkill07:
@zhongxuanwang-nv Yes, I did say that, but it's a breaking change, so we should be wary. I was more so stating that in passing given the complexity of your original PR.

Let's not drop this (yet).

zhongxuanwang-nv:
> @zhongxuanwang-nv Yes, I did say that, but it's a breaking change, so we should be wary. I was more so stating that in passing given the complexity of your original PR.
> 
> Let's not drop this (yet).

@willkill07 Ohh got it!! Yes, maybe we will revisit later. It should be a quick change!

PR #221, opened 2026-06-04
chore: Set status on OTel spans

dagardner-nv:
#### Overview

Set the `statusCode` field on OTel/OpenInference Spans, this is tracked in the metadata of the Scope.

#### Details

* Add the ability to update the scope's metadat when popping the current scope.
* This represents a breaking change to FFI/Go bindings, but is non-breaking for other bindings.

* `crates/core/src/api/scope.rs`
* `crates/core/src/api/runtime/state.rs`
* `crates/core/src/observability/mod.rs`

- Closes #

## Summary by CodeRabbit

* **New Features**
  * Scope pop/end operations accept optional end-metadata that is merged with scope metadata.
  * End events now include OpenTelemetry status fields (otel.status_code / otel.status_message).

* **Improvements**
  * SDKs and language bindings propagate end-metadata and automatically record OK/ERROR status for scope/tool/LLM completions.

* **Tests**
  * Added cross-language tests validating OTEL status metadata and metadata-merge behavior on end events.


dagardner-nv:
@coderabbitai review

crates/node/tests/tools_tests.mjs
```diff
@@ -38,6 +39,13 @@ function rejectWithPrimitive(value) {
   return Promise.reject(value);
 }
 
+async function flushSubscriberCallbacks() {
+  flushSubscribers();
+  for (let i = 0; i < 10; i += 1) {
+    await new Promise((resolve) => setImmediate(resolve));
+  }
```
willkill07: Why the loop of 10?

willkill07 (dismissed):
One question. Otherwise, I like the additional changes here. Thanks for digging.

crates/core/src/api/shared.rs
```diff
@@ -37,6 +38,25 @@ pub(crate) fn ensure_runtime_owner() -> Result<()> {
...
+        "otel.status_code".to_string(),
+        Json::String(status_code.to_string()),
+    );
+    if let Some(status_message) = status_message {
+        status.insert(
+            "otel.status_message".to_string(),
+            Json::String(status_message),
+        );
+    }
+    merge_json(metadata, Some(Json::Object(status)))
```
mnajafian-nv: Could we explicitly clear `otel.status_message` when we emit `OK` or `UNSET` here? Since this helper shallow-merges onto existing metadata, omitting the key looks like it could preserve a prior `otel.status_message` and leave a successful end event/span carrying a stale error message. Would it make sense to add a regression test for that case too?

crates/core/src/api/llm.rs
```diff
@@ -587,19 +587,22 @@ pub async fn llm_call_execute(params: LlmCallExecuteParams) -> Result<Json> {
                 .as_ref()
                 .and_then(|codec| codec.decode_response(&response).ok())
                 .map(Arc::new);
+            let end_metadata = metadata_with_otel_status(metadata, "OK", None);
```
mnajafian-nv: Could we double-check that we want to explicitly set `otel.status_code = "OK"` on successful spans rather than leaving success as `Unset`? I can see the value in making success explicit, but since this becomes a cross-binding contract, it seems worth confirming that this is the intended observability semantics.

crates/core/src/api/shared.rs
```diff
@@ -37,6 +38,25 @@ pub(crate) fn ensure_runtime_owner() -> Result<()> {
...
+    status_message: Option<String>,
+) -> Option<Json> {
+    let mut status = serde_json::Map::new();
+    status.insert(
+        "otel.status_code".to_string(),
+        Json::String(status_code.to_string()),
+    );
+    if let Some(status_message) = status_message {
+        status.insert(
+            "otel.status_message".to_string(),
```
mnajafian-nv: Could we sanity-check the choice of `otel.status_message` as the metadata key here? OTel’s reserved attribute naming uses `otel.status_description`, so I wanted to confirm whether we intentionally want a NeMo Relay-specific event metadata key or whether aligning with the standard name would make the surface easier for downstream consumers.

mnajafian-nv (dismissed):
LGTM, approved upon alignment on inline questions.

crates/node/tests/tools_tests.mjs
dagardner-nv: Replaced with an explicit check on the events the test is expecting within a 15 second timeout

crates/core/src/api/shared.rs
dagardner-nv: We can change it to `status_description` that isn't a problem and likely is a better choice since that is what OTel calls it.

willkill07 (dismissed):

crates/core/src/api/llm.rs
dagardner-nv: I this case I believe so. The reason is that for `llm_call_execute` and `tool_call_execute` they both push and pop their own scope without an opportunity for the caller to update the scope later. 

So we must assume/depend that the underlying call is returning an error on failure.

mnajafian-nv (approved):
Thanks for addressing the feedback, LGTM.

PR #173, opened 2026-05-28
feat: add streaming ATOF exporter

omkarmehta06:
#### Overview

Add a streaming ATOF exporter for NeMo Relay so downstream tools can consume canonical ATOF JSONL from a separate local process in real time.

- [x] I confirm this contribution is my own work, or I have the right to submit it under this project's license.
- [x] I searched existing issues and open pull requests, and this does not duplicate existing work.

#### Details

- Adds a TCP-backed `AtofStreamingExporter` that connects to a local receiver and writes newline-delimited canonical ATOF events.
- Keeps socket writes out of the subscriber callback path with a bounded queue and dedicated writer thread.
- Preserves Relay as the recorder/exporter owner; downstream cockpit/viewer tools can consume the stream without introducing a second trace contract.
- Adds focused tests for canonical delivery, connection/config errors, stored stream errors, shutdown/flush behavior, and future-events-only runtime registration.
- Updates the ATOF observability docs with streaming exporter and receiver examples.

#### Where should the reviewer start?

Start with `crates/core/src/observability/atof.rs`, especially `AtofStreamingExporter`, then review the socket receiver tests in `crates/core/tests/unit/observability/atof_tests.rs`.

#### Validation

Already run locally:

- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test -p nemo-relay streaming_exporter -- --test-threads=1`
- `just test-rust`
- `just docs`

#### Breaking changes

None intended. This adds a new exporter surface and docs; existing ATOF/ATIF file export behavior remains unchanged.

#### Related Issues

- Supersedes #172

## Release Notes

* **New Features**
  * Streaming observability exporter: Send agent trajectory events over TCP in real-time with bounded queue buffering and event dropping on overflow

* **Documentation**
  * Added architecture and design documentation for streaming observability
  * Extended configuration guides with examples for Rust, Node.js, Python, and Go SDKs

* **Tests**
  * Added test coverage for streaming observability configuration and runtime behavior

[Review Change Stack](https://app.coderabbit.ai/change-stack/NVIDIA/NeMo-Relay/pull/173?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #232, opened 2026-06-05
feat: add streaming ATOF endpoints

willkill07:
#### Overview

Adds optional streaming endpoints to the ATOF exporter while preserving the existing JSONL file output path. Each raw ATOF event can now fan out to configured HTTP POST, WebSocket, or long-lived NDJSON endpoints, and `nemo-relay doctor` probes configured endpoints with a synthetic ATOF mark event.

#### Details

- Extended Rust ATOF exporter config with endpoint definitions, transport parsing, startup validation, per-endpoint unbounded FIFO queues, terminal flush behavior, and best-effort streaming failure handling.
- Added observability plugin schema/editor support, endpoint validation, wasm unsupported diagnostics, and doctor endpoint probes for `http_post`, `websocket`, and `ndjson`.
- Updated Python, Node.js, FFI, Go, and Wasm type/config surfaces, preserving the existing FFI constructor and adding a JSON-config constructor for endpoint-aware creation.
- Added focused core, CLI, Python, Node.js, Go, and plugin tests, plus ATOF/configuration/troubleshooting documentation updates.

Start with `crates/core/src/observability/atof.rs` for exporter lifecycle and transport behavior, then `crates/core/src/observability/plugin_component.rs` and `crates/cli/src/doctor.rs` for config validation and doctor probing.

Validation run:

- `cargo test -p nemo-relay observability::atof --lib`
- `cargo test -p nemo-relay observability::plugin_component --lib`
- `cargo test -p nemo-relay-cli collect_observability_probes_atof_streaming_endpoint`
- `cargo test -p nemo-relay-cli --test cli_tests`
- `just build-python`
- `just build-node`
- `just build-go`
- `uv run pytest python/tests/test_observability_plugin.py python/tests/test_types.py -q`
- `node --test tests/atof_tests.mjs tests/observability_plugin_tests.mjs`
- `go test -v -run 'Test(NewAtofExporterConfigDefaults|AtofExporterLifecycleWritesRawJSONL|AtofExporterAppendAndOverwriteModes|ObservabilityConfigHelpers|ObservabilityPluginAtofAndAtifFiles|ObservabilityPluginValidationRejectsBadValues)' ./...`
- `cargo check -p nemo-relay --no-default-features`
- `just docs`
- Commit hook: pre-commit checks, including `cargo clippy`, `cargo check`, `cargo deny`, attribution sync, Go vet/fmt, and Node prettier checks.

- Closes RELAY-185
- Closes #173

## Summary by CodeRabbit

* **New Features**
  * ATOF streaming: send observability events to multiple endpoints (http_post, websocket, ndjson) with per-endpoint headers/timeouts; exporter lifecycle, flush/shutdown, and language/FFI bindings updated across Python, Node, Go, Rust, WASM, and C.

* **Tests**
  * New unit/integration tests covering streaming transports, validation, FIFO/reconnect behavior, doctor probing, and FFI error cases.

* **Documentation**
  * Updated docs and runbooks with examples, validation rules, and diagnostics for streaming endpoints.

* **Chores**
  * License attributions standardized to SPDX identifiers and inlined license texts.


crates/core/src/observability/atof.rs
```diff
@@ -252,15 +385,377 @@ fn open_file(path: &Path, mode: AtofExporterMode) -> Result<File> {
...
+    while let Some(message) = rx.recv().await {
+        match message {
+            EndpointMessage::Event(raw_json) => {
+                if let Err(error) = body_tx.send(Ok(format!("{raw_json}\n").into_bytes())) {
+                    eprintln!("nemo_relay: ATOF endpoint[{index}] NDJSON send failed: {error}");
+                }
+            }
+            EndpointMessage::Close(done) => {
+                drop(body_tx);
+                match request.await {
```
mnajafian-nv: Could this NDJSON close path block shutdown indefinitely if the server accepts the connection but never completes the HTTP response? `close()` waits for the worker ack, and on `Close` we await the upload task with only a connect timeout. Would it make sense to add an overall request/close timeout or make shutdown best-effort so `force_flush()`/`shutdown()` cannot hang?

crates/core/src/observability/atof.rs
```diff
@@ -204,20 +327,22 @@ impl AtofExporter {
             .state
             .lock()
             .map_err(|_| AtofExporterError::LockPoisoned)?;
-        state
+        if state.closed {
+            return stored_failure_result(&self.path, &state);
+        }
+        state.closed = true;
```
mnajafian-nv, line 355: Do we want `force_flush()` to become terminal here? This now sets `state.closed = true` and drops post-flush events, which differs from prior ATOF behavior and from the other exporters where `force_flush()` is non-terminal and `shutdown()` is the terminal call. If that divergence is intentional, could we either move the terminal behavior to `shutdown()` or update the public contract/docs across bindings?

crates/cli/src/doctor.rs
```diff
@@ -725,6 +740,303 @@ fn section_endpoint(config: &Value, section: &str) -> Option<String> {
...
+            name: "ATOF endpoint",
+            status: Status::Pass,
+            details: format!(
+                "endpoints[{index}] http_post {url} (HTTP {})",
+                response.status()
+            ),
+        },
+        Ok(response) => Check {
+            name: "ATOF endpoint",
+            status: Status::Warn,
```
mnajafian-nv: How do we want `doctor` to classify non-2xx responses for `http_post` here and the similar `ndjson` path below? Right now 401/403/500 are reported as `Warn`, which still exits 0 even though event delivery is effectively broken. Should these be `Fail`, or is the intended contract that `doctor` only checks reachability rather than deliverability?

mnajafian-nv (dismissed):
This looks directionally good, and I don’t want to block the overall approach. I left a few inline questions that I think should be resolved before merge, especially around the NDJSON shutdown path and the force_flush() contract change. Happy to re-review quickly once those are addressed.

crates/core/src/observability/atof.rs
willkill07: switched to shutdown

mnajafian-nv (approved):
LGTM!

Salonijain27 (approved):

PR #236, opened 2026-06-05
feat: add LLM cost layering and pricing lookup

AjayThorve:
#### Overview

Adds a durable LLM cost layer for annotated provider responses. Relay now preserves upstream-reported cost, can estimate cost from configured pricing sources when token usage and model pricing are available, and leaves cost unknown instead of fabricating `$0` when pricing or required token data is missing.

This PR does not make Relay the owner of a canonical bundled pricing catalog. Relay owns the cost schema, pricing-source/resolver behavior, precedence rules, CLI/plugin validation, and observability propagation; pricing data is supplied by configured sources today, with DB/service-backed sources possible later by adapting them into validated catalog snapshots.

#### Details

- Extends normalized response usage with `CostEstimate`, `CostSource`, and `Usage.cost` while keeping provider response annotations limited to response fields such as provider response `id`, `model`, message, tool usage, and token usage.
- Uses the managed LLM call name as provider/route identity for provider-aware pricing, so routes such as `azure/openai` can resolve differently from generic `openai` without adding a `model_provider` field to `AnnotatedLlmResponse`.
- Adds a source-backed pricing resolver with file, inline, and custom `PricingSource` support. The resolver handles provider-qualified/routed model names and uses first-match source precedence.
- Keeps provider/framework-reported cost authoritative. Model pricing estimates are only computed when `Usage.cost` is absent and a configured pricing source has enough token-pricing data.
- Represents normalized cost amounts as currency-neutral fields (`total`, `input`, `output`, `cache_read`, `cache_write`) plus `currency`; generic pricing APIs stay currency-neutral, top-level provider/framework `usage.cost_usd` input remains supported, and exporters emit either their schema-required cost names or Relay cost `total` plus `currency`.
- Treats unknown models, stale/auditable pricing, non-token pricing units, and missing token fields explicitly. Unknown cost remains absent instead of being rounded or defaulted to zero.
- Propagates normalized or estimated cost into ATIF step/final metrics, OpenInference attributes, and OpenTelemetry attributes without changing existing token behavior.
- Adds `nemo-relay pricing validate`, `init`, `add-source`, and `resolve` so users can validate catalog files, wire project/user config, and inspect which pricing source matched a model.
- Extends `nemo-relay doctor` to validate configured pricing sources, so missing/unreadable files or invalid pricing catalogs surface before gateway launch.
- Updates Rust, Python, Node, WASM, Go, and FFI coverage plus end-user docs for provider response codecs, CLI pricing setup, plugin configuration layering, direct application instrumentation, and embedded plugin initialization across harnesses/custom agents/framework integrations.

Start with `crates/core/src/codec/response.rs` for the normalized response cost shape and `crates/core/src/codec/pricing.rs` for provider-aware lookup. Then review `crates/core/src/api/llm.rs` / `crates/core/src/stream.rs` for how LLM call names feed pricing, `crates/cli/src/pricing.rs` and `crates/cli/src/doctor.rs` for CLI setup/validation behavior, and `docs/integrate-into-frameworks/provider-response-codecs.mdx` / `docs/instrument-applications/instrument-llm-call.mdx` for the public behavior contract across CLI and non-CLI consumers.

- Resolves RELAY-184
- Resolves RELAY-227
- Resolves RELAY-228
- Resolves RELAY-229
- Resolves RELAY-230

#### Validation

- `cargo test -p nemo-relay provider_reported_cost -- --nocapture`
- `cargo test -p nemo-relay test_attach_estimated_cost_uses_event_provider -- --nocapture`
- `cargo test -p nemo-relay pricing -- --nocapture`
- `cargo test -p nemo-relay atif -- --nocapture`
- `cargo test -p nemo-relay openinference -- --nocapture`
- `cargo test -p nemo-relay otel -- --nocapture`
- `cargo test -p nemo-relay --test pipeline_integration -- --nocapture`
- `cargo test -p nemo-relay-cli cli_pricing -- --nocapture`
- `cargo test -p nemo-relay-cli collect_observability_ -- --nocapture`
- `cargo test -p nemo-relay-cli doctor::tests -- --nocapture`
- `cargo run -p nemo-relay-cli -- doctor codex --json | jq '.observability'`
- `cargo run -p nemo-relay-cli -- pricing --help`
- `cargo fmt --check && git diff --check`
- `RUST_TEST_THREADS=1 just test-rust`
- `just test-go`
- `PATH=/Users/athorve/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH just test-node`
- `PATH=/Users/athorve/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH just test-wasm`
- `env -u CONDA_PREFIX just test-python`
- `PATH=/Users/athorve/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH just docs`

Notes:
- `just docs` passed with the existing Fern redirect warning because this local environment is not authenticated with Fern.
- `just test-python` required using the repo `.venv` without the outer `CONDA_PREFIX`.

## Summary by CodeRabbit

* **New Features**
  * Built-in pricing plugin and resolver to annotate LLM responses with cost estimates; CLI pricing commands (validate, init, add-source, resolve) with scope flags and preserved source-order merging.
  * LLM end events and exporters (OTEL/ATIF/OpenInference) now emit normalized cost metadata when available.

* **Tests**
  * Extensive new and updated tests for CLI flows, codecs, resolver behavior, observability exports, and merge ordering.

* **Documentation**
  * New and expanded guides for pricing setup, CLI usage, catalog schema, and cost-estimation behavior.

### DIFF
--- a/crates/core/src/codec/response.rs
+++ b/crates/core/src/codec/response.rs
@@ -221,7 +221,9 @@ pub(crate) fn provider_reported_cost(
 
     Some(CostEstimate {
         total,
-        currency: if has_currency_native_amount {
+        currency: if provider_total_cost.is_some() {
+            default_cost_currency()
+        } else if has_currency_native_amount {
             cost.currency.unwrap_or_else(default_cost_currency)
         } else {
             default_cost_currency()
--- a/crates/core/src/codec/response.rs
+++ b/crates/core/src/codec/response.rs
@@ -213,10 +213,16 @@ pub(crate) fn provider_reported_cost(
     cost: Option<RawUsageCost>,
 ) -> Option<CostEstimate> {
     let cost = cost.unwrap_or_default();
-    let input = cost.input;
-    let output = cost.output;
-    let cache_read = cost.cache_read;
-    let cache_write = cost.cache_write;
+    let provider_total_uses_default_currency = provider_total_cost.is_some();
+    let nested_currency_is_default = cost
+        .currency
+        .as_deref()
+        .is_none_or(|currency| currency.eq_ignore_ascii_case("USD"));
+    let keep_component_costs = !provider_total_uses_default_currency || nested_currency_is_default;
+    let input = keep_component_costs.then_some(cost.input).flatten();
+    let output = keep_component_costs.then_some(cost.output).flatten();
+    let cache_read = keep_component_costs.then_some(cost.cache_read).flatten();
+    let cache_write = keep_component_costs.then_some(cost.cache_write).flatten();
     let has_currency_native_amount = cost.total.is_some()
         || cost.input.is_some()
         || cost.output.is_some()
@@ -243,7 +249,7 @@ pub(crate) fn provider_reported_cost(
 
     Some(CostEstimate {
         total,
-        currency: if provider_total_cost.is_some() {
+        currency: if provider_total_uses_default_currency {
             default_cost_currency()
         } else if has_currency_native_amount {
             cost.currency.unwrap_or_else(default_cost_currency)
--- a/crates/core/src/codec/response.rs
+++ b/crates/core/src/codec/response.rs
@@ -151,12 +151,34 @@ pub struct CostEstimate {
 }
 
 impl CostEstimate {
+    /// Returns the explicit total, or the sum of component costs when no total was supplied.
+    #[must_use]
+    pub fn total_or_component_sum(&self) -> Option<f64> {
+        self.total.or_else(|| {
+            let (has_component, total) =
+                [self.input, self.output, self.cache_read, self.cache_write]
+                    .into_iter()
+                    .flatten()
+                    .fold((false, 0.0), |(_, total), value| (true, total + value));
+            has_component.then_some(total)
+        })
+    }
+
     /// Returns the total only when it is denominated in the requested currency.
     #[must_use]
     pub fn total_for_currency(&self, currency: &str) -> Option<f64> {
         self.currency
             .eq_ignore_ascii_case(currency)
             .then_some(self.total)
+            .flatten()
+    }
+
+    /// Returns the explicit or component-derived total in the requested currency.
+    #[must_use]
+    pub fn total_or_component_sum_for_currency(&self, currency: &str) -> Option<f64> {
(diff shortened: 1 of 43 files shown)


crates/core/src/codec/response.rs
```diff
@@ -85,6 +94,176 @@ pub struct Usage {
...
+        && output.is_none()
+        && cache_read.is_none()
+        && cache_write.is_none()
+    {
+        return None;
+    }
+
+    Some(CostEstimate {
+        total,
+        currency: if provider_total_cost.is_some() {
```
mnajafian-nv: Question on normalization here: if `provider_total_cost` is present we force `currency = "USD"`, but we still preserve `input` / `output` / `cache_*` from the nested `cost` object unchanged. Is it possible to end up with a USD `total` plus non-USD component fields in the same `CostEstimate`? If so, should we drop those components when `cost_usd` wins, or only keep them when the nested currency is also USD/unspecified?

mnajafian-nv (dismissed):
Overall this looks solid. I’d be happy to give this a conditional approval once the mixed-currency normalization case above is cleaned up.

crates/core/src/codec/response.rs
AjayThorve: Good catch, this is a real normalization edge case.

The intended behavior should be:

- If `cost_usd` is present, Relay treats that as an authoritative provider-reported USD total.
- If the nested `cost.currency` is `USD` or unspecified, we can preserve `input` / `output` / `cache_*` because those components are USD-compatible.
- If the nested `cost.currency` is non-USD, we should not carry those component fields into the normalized `CostEstimate`, because that would create a mixed object: USD `total` with non-USD component values under a single `currency`.

So the success case is `cost_usd + cost.currency = USD` preserving total and components. The failure case is `cost_usd + cost.currency = EUR`, where Relay should keep only the USD total and drop the component breakdown rather than pretending we converted it.

I’ll update the normalization and add regression coverage for both cases.

willkill07 (approved):

PR #245, opened 2026-06-08
feat: add PII redaction plugin crate with local backend

afourniernv:
#### Overview

Add a dedicated `nemo-relay-pii-redaction` crate for Relay’s first-party privacy plugin. This PR ships the deterministic hot-path `builtin` backend now, keeps a future `local_model` seam for model-backed detection/redaction later, and separates the plugin from raw middleware by giving users a supported config-driven privacy surface instead of requiring custom sanitize callbacks.

#### Details

- add a new `nemo-relay-pii-redaction` first-party crate
- keep `pii_redaction` distinct from raw guardrails middleware:
  - middleware remains the generic runtime hook mechanism
  - `pii_redaction` becomes the packaged, first-party, config-driven privacy policy surface on top of those hooks
  - this gives users a supported privacy feature with typed config, editor support, codec-aware LLM handling, and shared behavior across runtimes instead of requiring bespoke callback code
- ship the deterministic `builtin` backend in this PR, including:
  - `remove`, `redact`, `regex_replace`, `hash`, and `mask`
  - built-in detector presets for common PII, structured secrets, and cloud credentials
  - codec-aware sanitization for supported LLM surfaces
  - tool and LLM sanitize guardrail registration for observability hot-path use
- keep `local_model` in the plugin contract as the next backend seam:
  - no runtime implementation is shipped yet
  - the seam is there so a future model-backed local detector/redactor can be added without redesigning the plugin surface again
- explicitly register the plugin from CLI, Python, Node, FFI, and WASM startup/validation paths instead of treating it as a core auto-registered plugin
- update CLI typed editor support and coverage for the externalized plugin crate
- move the plugin into its own crate now because we expect the second backend lane to grow a heavier dependency/runtime surface, and keeping both lanes together under one plugin-owned crate boundary is cleaner than splitting only one backend out of core later

Start with `crates/pii-redaction/src/component.rs` and `crates/pii-redaction/src/lib.rs`, then review `crates/pii-redaction/src/builtin.rs` for the shipped deterministic backend and the explicit plugin registration changes in `crates/cli/src/server.rs` and `crates/python/src/lib.rs`.

- Relates to RELAY-181

## Summary by CodeRabbit

* **New Features**
  * Added a first‑party PII Redaction plugin with configurable actions (remove, redact, mask, hash, regex replace), built‑in detectors for common PII/secrets, codec‑aware LLM request/response sanitization, and interactive CLI editor support.
  * Plugin integrated across runtimes (CLI, FFI, Node, Python, WASM).

* **Dependencies**
  * Updated sha2 from 0.10.9 to 0.11.0.

* **Documentation**
  * Added comprehensive plugin docs and configuration guides.


crates/pii-redaction/src/builtin.rs
```diff
@@ -0,0 +1,430 @@
...
+        self.sanitize_json_preorder_dfs_at_path(value, &mut Vec::new())
+            .unwrap_or(Json::Null)
+    }
+
+    fn sanitize_json_preorder_dfs_at_path(
+        &self,
+        value: Json,
+        path_segments: &mut Vec<String>,
+    ) -> Option<Json> {
+        if self.matches_current_preorder_path(path_segments)
```
mnajafian-nv: Could we tighten the empty target_paths behavior for remove here? The docs and config comments say an empty target list means every matching string leaf, but because empty targets currently match the root, action = "remove" returns None before traversal and can turn the entire emitted payload into null. Since remove is also the default builtin action, would it be safer to either make empty paths leaf-only for remove or require explicit target_paths for remove, with a regression test for the empty-path case?

crates/pii-redaction/src/detectors.rs
```diff
@@ -0,0 +1,347 @@
...
+        detector: BuiltinDetector::IpAddress,
+        name: "ip_address",
+        category: BuiltinDetectorCategory::CommonPii,
+        regex_pattern: r"\b(?:\d{1,3}\.){3}\d{1,3}\b",
+    },
+    BuiltinDetectorSpec {
+        detector: BuiltinDetector::Ipv6,
+        name: "ipv6",
+        category: BuiltinDetectorCategory::CommonPii,
+        regex_pattern: r"\b(?:[A-Fa-f0-9]{1,4}:){2,7}[A-Fa-f0-9]{1,4}\b",
```
mnajafian-nv: Could we support compressed IPv6 forms in this preset and add a small regression test? The current pattern requires explicit segments, so common valid addresses like 2001:db8::1 or ::1 will not be detected or masked even though the plugin exposes ipv6 as a built-in detector. That feels worth closing before this becomes the documented privacy contract.

crates/pii-redaction/src/builtin.rs
```diff
@@ -0,0 +1,430 @@
...
+
+pub(crate) fn mask_text(
+    text: &str,
+    mask_char: &str,
+    unmasked_prefix: usize,
+    unmasked_suffix: usize,
+) -> String {
+    let chars: Vec<char> = text.chars().collect();
+    let len = chars.len();
+    if len <= unmasked_prefix + unmasked_suffix {
```
mnajafian-nv: Could we make this boundary check overflow-safe? These counts come from config, so an extreme unmasked_prefix plus unmasked_suffix can overflow before the comparison and later make the subtraction compute an unexpectedly huge mask length. A saturating_add check, or validation that caps the configured counts, would keep the masking path predictable for adversarial or malformed config values.

crates/pii-redaction/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+
+[lints]
+workspace = true
+
+[features]
+default = []
+schema = ["dep:schemars"]
+
+[dependencies]
+nemo-relay = { workspace = true, features = ["schema"] }
```
mnajafian-nv: Could we route the core schema feature through this crate own schema feature instead of enabling it unconditionally? As written, any runtime that depends on nemo-relay-pii-redaction also enables nemo-relay/schema even when it is not generating JSON Schema. Since this crate already has an optional schema feature, would feature wiring like schema = ["dep:schemars", "nemo-relay/schema"] keep the package boundary cleaner?

crates/pii-redaction/src/builtin.rs
afourniernv: Huge bug, thanks for catching.

When target_paths is empty, we should traverse the whole payload and prune matching string leaves during preorder DFS. Before this fix, remove matched the root immediately and could prune the entire payload before traversal reached the leaves.

addressed.

crates/pii-redaction/src/detectors.rs
afourniernv: addressed. good catch.

crates/pii-redaction/src/builtin.rs
afourniernv: addressed. no overflow risk here anymore.

crates/pii-redaction/Cargo.toml
afourniernv: addressed.

crates/pii-redaction/README.md
willkill07, line 1: If we want to publish this crate, then we need to have a more comprehensive README (similar to adaptive)

Though we do not need to publish, necessarily.

crates/pii-redaction/src/overlay.rs
```diff
@@ -0,0 +1,420 @@
...
+                    && let Some(block) = block.as_object_mut()
+                {
+                    block.remove("text");
+                }
+            }
+        }
+        return;
+    };
+
+    let parts: Vec<&str> = text.split('\n').collect();
```
mnajafian-nv, line 231: Worth preserving the full text here when a single provider text block contains newlines? The normalized `/message` path joins text parts with `\n`, but this overlay splits it back across provider blocks and appears to write only the first part when there is one raw `output_text` block, which could silently drop part of the sanitized emitted response.

crates/pii-redaction/src/component.rs
```diff
@@ -0,0 +1,830 @@
...
+            policy.unsupported_value,
+            "pii_redaction.unsupported_value",
+            Some(PII_REDACTION_PLUGIN_KIND.to_string()),
+            Some("builtin.action".to_string()),
+            "builtin.action must be 'remove', 'redact', 'regex_replace', 'hash', or 'mask'"
+                .to_string(),
+        );
+    }
+
+    if matches!(builtin.action.as_str(), "regex_replace" | "redact")
```
mnajafian-nv, line 595: Should we compile-check `builtin.pattern` during validation as well as activation? That would let CLI/doctor catch malformed regex config before startup tries to register the plugin.

crates/pii-redaction/src/component.rs
```diff
@@ -0,0 +1,830 @@
...
+            codec: None,
+            builtin: None,
+            local: None,
+            policy: ConfigPolicy::default(),
+        }
+    }
+}
+
+/// Built-in redaction backend settings.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
```
mnajafian-nv: Would it be safer for `BuiltinBackendConfig::default()` to match the serde/docs default for `action`? The editor section default uses this Rust default, so the derived default currently materializes `action = ""` rather than the documented `remove`.

crates/pii-redaction/src/component.rs
```diff
@@ -0,0 +1,830 @@
...
+    diagnostics: &mut Vec<ConfigDiagnostic>,
+    policy: &ConfigPolicy,
+    plugin_config: &Map<String, Json>,
+) {
+    validate_section_fields(
+        diagnostics,
+        policy,
+        plugin_config,
+        "policy",
+        &["unknown_field", "unsupported_value"],
```
mnajafian-nv: Is `unknown_component` intentionally excluded from the component-local policy allowlist? Since this uses the shared `ConfigPolicy` and the docs example includes `policy.unknown_component`, the current validation looks like it can warn on the policy shape we tell users to write.

willkill07 (commented):
Leaving a deferred feedback note. Commenting only to gate merge based on @mnajafian-nv final approval.

crates/pii-redaction/src/component.rs
afourniernv: Addressed in c8ed5b13.
Validation now compile-checks builtin.pattern, so malformed regexes fail in CLI/doctor validation instead of only during plugin activation. Added a regression test for invalid patterns.

crates/pii-redaction/src/overlay.rs
afourniernv: Addressed in c41d145a.

We now preserve the full sanitized text when the provider payload has a single raw text block, and keep the existing split behavior only for multi-block payloads. Added regression coverage for both OpenAI Responses and Anthropic single-block multiline cases.

crates/pii-redaction/src/component.rs
afourniernv: Addressed in 08dcbc63.
 
BuiltinBackendConfig::default() now aligns with the documented/schema default and materializes action = "remove" instead of the empty string. Added a regression test to lock that default in.

crates/pii-redaction/src/component.rs
afourniernv: Addressed in ecdf49a6. 
The component-local policy allowlist now includes unknown_component, which matches the documented config shape and avoids warning on the example we tell users to use. Added  coverage for that validation path.

crates/pii-redaction/README.md
afourniernv: added a lite readme.

crates/node/pii_redaction.js
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Create deterministic built-in redaction backend settings with defaults applied.
+ *
+ * @param {object} [config={}] - Partial built-in settings to override.
+ * @returns {object} A normalized built-in backend config object.
+ */
+function builtinConfig(config = {}) {
+  return {
+    action: 'redact',
```
mnajafian-nv: Would it make sense to keep the JavaScript helper default aligned with the Rust/Python builtin default here? `builtinConfig()` currently returns `action: "redact"`, but validation requires a `pattern` or `detector` for `redact`, so a bare JS helper config can fail while the Rust/Python default `remove` remains valid. This likely also wants the matching WASM wrapper/tests updated for parity.

python/nemo_relay/pii_redaction.py
```diff
@@ -0,0 +1,199 @@
...
+
+def _normalize_object(value: object) -> JsonObject:
+    return cast(JsonObject, _normalize(value))
+
+
+@dataclass(slots=True)
+class ConfigPolicy:
+    """Policy for unsupported PII redaction configuration."""
+
+    unknown_field: UnsupportedBehavior = "warn"
```
mnajafian-nv, line 63: Should the typed PII helper policy expose `unknown_component` as well for parity with the Rust config contract? The Rust `PiiRedactionConfig` uses the shared `ConfigPolicy`, and validation accepts `policy.unknown_component`, so adding it here plus the `.pyi`/TS declarations would keep the helper surface aligned with the canonical config shape.

crates/pii-redaction/src/component.rs
```diff
@@ -0,0 +1,859 @@
...
+        "local",
+        &[
+            "backend",
+            "model_id",
+            "detector_profile",
+            "allow_network",
+            "max_latency_ms",
+        ],
+    );
+    validate_mode(&mut diagnostics, &config.policy, &config);
```
mnajafian-nv, line 466: Is it worth adding an explicit `version` check in this validation path before merge? Other first-party config surfaces reject unsupported schema versions, so having `pii_redaction` silently accept `version = 2` could make future migrations harder to reason about.

crates/node/pii_redaction.js
afourniernv: Addressed in 32de1b6d.
  builtinConfig() now defaults to action: "remove" to match the Rust/Python helper contract, and I updated the matching Node/WASM helper tests for parity.

python/nemo_relay/pii_redaction.py
afourniernv: Addressed in 32de1b6d.
  The typed PII helper policy now exposes unknown_component in the Python helper and .pyi, and in the Node/WASM TypeScript declarations, so the helper surface matches the canonical Rust config shape.

crates/pii-redaction/src/component.rs
afourniernv: Addressed in 32de1b6d.
pii_redaction validation now rejects unsupported config versions explicitly, matching the pattern used by the other first-party config surfaces. I also added regression coverage for the unsupported-version path.

crates/pii-redaction/src/overlay.rs
```diff
@@ -0,0 +1,462 @@
...
```
mnajafian-nv, line 231: Would it be safer to avoid using split('\n') as the provider-block boundary here? The response codec normalizes multiple output_text blocks by joining them with newline characters, but an individual raw block can also contain newlines, so when output_text_count > 1 this can write only the first line into block 0 and drop the remainder. Maybe the overlay should preserve/sanitize each original block independently, or use a conservative fallback that never drops sanitized text when the block mapping is ambiguous?

crates/pii-redaction/src/overlay.rs
```diff
@@ -0,0 +1,462 @@
...
+        }
+    }
+}
+
+fn overlay_anthropic_text_blocks(blocks: &mut [Json], message_text: Option<String>) {
+    let text_block_count = blocks
+        .iter()
+        .filter(|block| block.get("type").and_then(Json::as_str) == Some("text"))
+        .count();
+    let parts = message_text
```
mnajafian-nv, line 265: Can we apply the same boundary check here for Anthropic text blocks? The normalized response text has already lost the original block boundaries, so splitting on newline can truncate a block that legitimately contained newlines when text_block_count > 1. Would preserving per-block text through the overlay path, or falling back to a no-drop mapping when boundaries are ambiguous, make this safer?

crates/pii-redaction/src/overlay.rs
afourniernv: I don’t see a clean way to solve this without preserving per-block text structure through the normalized/annotated response shape, and I’m worried about the blast radius of
changing that contract in this PR. The single-block case was straightforward because there’s no boundary ambiguity there; the multi-block case isn’t.

I added local repro tests for both openai_responses and anthropic and they do fail today, so your point is valid, but it looks like the real fix is wider than I’m comfortable landing in this PR.

crates/pii-redaction/src/overlay.rs
afourniernv: Same concern ... text_block_count > 1 case feels like it would require a change to annotatedrequest/response.

mnajafian-nv (approved):
LGTM

willkill07 (approved):

afourniernv:
merge

PR #279, opened 2026-06-16
feat: add dynamic plugin control-plane foundation

afourniernv:
#### Overview

Add the first Rust-side dynamic plugin control-plane slice for RELAY-337.

#### Details

- add `crate::plugin::dynamic` as the foundation for dynamic plugin control-plane types
- introduce durable dynamic plugin record modeling with `metadata`, `source`, `spec`, and `status`
- add authored `relay-plugin.toml` manifest types, parsing, path loading, strict validation, and conversion into registry records
- add an in-memory dynamic plugin registry with add/get/list/enable/disable/remove semantics, tombstone revival, and status updates
- enforce kind-specific manifest and raw-record shape validation for `rust_dynamic` and `worker` plugin lanes
- add focused unit coverage for manifest validation, canonical path handling, registry lifecycle semantics, tombstones, generation behavior, and raw-record integrity checks
- add `PluginError::Conflict` and propagate it through adaptive and FFI error mappings

Validation run:
- `cargo fmt --all --check`
- `cargo test -p nemo-relay plugin::dynamic::tests --lib`

Known validation gap:
- `just test-rust` is currently red on this branch because of the pre-existing failure in `crates/core/tests/integration/pipeline_tests.rs` for `test_response_codec_failure_non_fatal`, which appears unrelated to this change.

Start in `crates/core/src/plugin/dynamic.rs` and then review:
- `crates/core/src/plugin/dynamic/manifest.rs`
- `crates/core/src/plugin/dynamic/registry.rs`
- `crates/core/tests/unit/plugin_dynamic_tests.rs`

- Relates to RELAY-337

## Release Notes

* **New Features**
  * Added manifest-based dynamic plugin support with a durable control-plane schema.
  * Introduced a dynamic plugin registry for enable/disable/remove plus validation and runtime status tracking.
  * Added `relay-plugin.toml` parsing and conversion into durable plugin records.

* **Bug Fixes**
  * Improved conflict error handling by treating plugin conflicts as “already exists” across components.

* **Tests**
  * Expanded unit and coverage tests for dynamic plugin models, manifest validation/loading, and conflict error mappings.

* **Chores**
  * Exposed the dynamic plugin module publicly and added the `Conflict` plugin error variant.

### DIFF
--- a/crates/core/src/plugin/dynamic.rs
+++ b/crates/core/src/plugin/dynamic.rs
@@ -52,12 +52,6 @@ pub enum DynamicPluginCapability {
     PluginNative,
     /// Isolated worker-based extension capability.
     PluginWorker,
-    /// Guardrail-style middleware registration capability.
-    MiddlewareGuardrail,
-    /// Interceptor-style middleware registration capability.
-    MiddlewareInterceptor,
-    /// Telemetry exporter registration capability.
-    TelemetryExporter,
     /// Typed configuration schema contribution capability.
     ConfigSchema,
 }
--- a/crates/core/src/plugin/dynamic.rs
+++ b/crates/core/src/plugin/dynamic.rs
@@ -1,0 +1,412 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Dynamic plugin control-plane and registry model.
+//!
+//! This module owns the durable control-plane record shape for dynamic plugins.
+//! Authored manifest parsing/validation and in-memory registry mutation logic
+//! live in dedicated submodules so those responsibilities do not accumulate in
+//! one file as the feature grows.
+
+use chrono::Utc;
+use serde::{Deserialize, Serialize};
+
+/// Canonical identifier for one dynamic plugin record.
+pub type DynamicPluginId = String;
+
+/// Canonical filename for authored Relay plugin manifests.
+pub const DYNAMIC_PLUGIN_MANIFEST_FILENAME: &str = "relay-plugin.toml";
+
+mod manifest;
+mod registry;
+
+pub use manifest::*;
+pub use registry::*;
+
+/// Plugin execution lane.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+#[serde(rename_all = "snake_case")]
+pub enum DynamicPluginKind {
+    /// Trusted in-process native plugin.
+    RustDynamic,
+    /// Isolated worker-based plugin runtime.
+    Worker,
+}
+
+/// Managed runtime identity for worker-based plugins.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+#[serde(rename_all = "snake_case")]
+pub enum WorkerRuntime {
+    /// Python worker runtime.
+    Python,
+}
+
+/// Relay-enforced capability declared by a dynamic plugin.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+#[serde(rename_all = "snake_case")]
+pub enum DynamicPluginCapability {
+    /// Trusted in-process native extension capability.
+    PluginNative,
+    /// Isolated worker-based extension capability.
+    PluginWorker,
+    /// Guardrail-style middleware registration capability.
+    MiddlewareGuardrail,
+    /// Interceptor-style middleware registration capability.
+    MiddlewareInterceptor,
+    /// Telemetry exporter registration capability.
+    TelemetryExporter,
+    /// Typed configuration schema contribution capability.
+    ConfigSchema,
+}
+
+/// Host policy startup classification for a plugin.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+#[serde(rename_all = "snake_case")]
+pub enum DynamicPluginStartupClass {
+    /// Failure is tolerated and the host may start in degraded mode.
+    Optional,
+    /// Failure is startup-fatal under current host policy.
+    Required,
+}
+
+/// Host attestation policy mode.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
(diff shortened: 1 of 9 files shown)


afourniernv:
A little more implementation detail on what this foundation slice is actually establishing.

This PR is centered on one durable control-plane record and one authored manifest contract.

## Record contract

The core type is `DynamicPluginRecord`, which is split into:
- `metadata`
- `source`
- `spec`
- `status`

The intended contract is:
- `metadata` = stable identity and record-level bookkeeping
- `source` = resolved manifest/artifact/environment references known to the control plane
- `spec` = desired state owned by user-facing lifecycle operations
- `status` = observed validation/runtime state owned by reconciliation/runtime updates

That separation is the main mechanism that keeps what the user wants separate from what Relay has observed.

`spec` currently carries:
- `present`
- `enabled`
- `config_ref`

`status` currently carries:
- decomposed validation state
- runtime state
- startup classification
- attestation mode
- last failure summary

Generation lives on `metadata`, and only desired-state mutations bump generation. Status updates do not. That is the reconciliation mechanism the later runtime and CLI work will build on.

## Manifest contract

The authored manifest is `relay-plugin.toml`.

This PR establishes:
- kind-specific manifest parsing
- strict validation
- manifest-to-record conversion

The current manifest model distinguishes:
- `plugin.kind = "rust_dynamic"`
- `plugin.kind = "worker"`

with kind-specific requirements for:
- compatibility declarations
- load contract fields
- capability shape

The durable load contract is now lane-specific rather than one shared optional-field bag:
- worker records carry `runtime + entrypoint`
- native records carry `library + symbol`

That makes the worker and native registration targets deterministic in the control-plane model instead of relying on loosely related optional fields.

Examples of what is enforced now:
- `compat.relay` is required for all plugins
- `compat.native_api` is required for `rust_dynamic`
- `compat.worker_protocol` is required for `worker`
- worker manifests require runtime + entrypoint
- native manifests require library + symbol
- plugins are rejected if they claim invalid kind/load/compatibility combinations
- plugins are rejected if `defaults.enabled = true`

So this PR is not just adding structs; it is making the authored contract executable.

## Registry mechanisms

`DynamicPluginRegistry` is intentionally in-memory for now, but the lifecycle semantics are already explicit.

It currently provides:
- `add`
- `add_manifest`
- `get`
- `list`
- `enable`
- `disable`
- `remove`
- validation/runtime/last-error status updates

Important behaviors already in place:
- `add_manifest` validates and converts authored manifests before registration
- duplicate live plugin IDs fail with `PluginError::Conflict`
- `remove` is tombstone-based, not destructive
- tombstoned IDs can be revived while preserving lineage
- `list(false)` hides tombstones
- `list(true)` includes them
- `enable` / `disable` / `remove` are desired-state mutations and bump generation
- status updates stamp observed timestamps but do not bump generation

That gives later CLI/runtime work a stable lifecycle model instead of re-deciding semantics.

## Boundary decisions in this PR

This slice intentionally stops at the control-plane boundary.

It does **not** yet implement:
- CLI lifecycle commands
- worker runtime protocol
- Python worker supervision
- native dynamic loading
- host policy evaluation
- persistence backend / cross-process coordination

Those are follow-on issues, but this PR is meant to establish the contracts they plug into.

## Why this shape

The main reason for this split is to avoid backing into the plugin model through CLI or runtime implementation details.

The order here is:
1. make the record model concrete
2. make the authored manifest contract concrete
3. make lifecycle mutation semantics concrete
4. build CLI, policy, worker, and native runtime work on top of those contracts

That is the actual purpose of this foundation PR.

crates/core/src/plugin/dynamic/manifest.rs
```diff
@@ -0,0 +1,436 @@
...
+    pub runtime: Option<WorkerRuntime>,
+    /// Worker entrypoint.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub entrypoint: Option<String>,
+    /// Native library path.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub library: Option<String>,
+    /// Native registration symbol.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub symbol: Option<String>,
```
willkill07, line 128: Split into concrete discriminated union (enum) types.

afourniernv:
/ok to test [dc158dc](https://github.com/NVIDIA/NeMo-Relay/pull/279/commits/dc158dcb5b445afa5af8d9c595ec25b365d0389d)

willkill07:
/ok to test [8451479](https://github.com/NVIDIA/NeMo-Relay/pull/279/commits/8451479b2f54296f53a4afc3f7815b8175cbec9d)

crates/core/src/plugin/dynamic.rs
```diff
@@ -0,0 +1,424 @@
...
+pub struct DynamicPluginCompatibility {
+    /// Compatible NeMo Relay version or version range.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub relay: Option<String>,
+    /// Native host API/ABI contract version for `rust_dynamic`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub native_api: Option<String>,
+    /// Worker protocol version for `worker`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub worker_protocol: Option<String>,
```
willkill07: These also seem like these should be split on native vs worker

crates/core/src/plugin/dynamic.rs
```diff
@@ -0,0 +1,424 @@
...
+    /// Trusted in-process native extension capability.
+    PluginNative,
+    /// Isolated worker-based extension capability.
+    PluginWorker,
+    /// Guardrail-style middleware registration capability.
+    MiddlewareGuardrail,
+    /// Interceptor-style middleware registration capability.
+    MiddlewareInterceptor,
+    /// Telemetry exporter registration capability.
+    TelemetryExporter,
```
willkill07: I thought we were going to omit this to avoid metadata-based permissions?

crates/core/src/plugin/dynamic.rs
```diff
@@ -0,0 +1,424 @@
...
+
+/// Durable control-plane record for a dynamic plugin.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+pub struct DynamicPluginRecord {
+    /// Stable plugin metadata.
+    pub metadata: DynamicPluginMetadata,
+    /// Declared capability set.
+    #[serde(default)]
+    pub capabilities: Vec<DynamicPluginCapability>,
```
willkill07: I think this won't be necessary

willkill07 (changes requested):
Sorry for the additional round of back-and-forth

crates/core/src/plugin/dynamic.rs
afourniernv: good catch. resolved.

crates/core/src/plugin/dynamic.rs
afourniernv: yep. removed.

crates/core/src/plugin/dynamic.rs
afourniernv: forgot about that comment - removed.

willkill07 (approved):

PR #291, opened 2026-06-23
refactor: hoist provider schema detection and normalization

zhongxuanwang-nv:
#### Overview

Consolidate provider schema handling for OpenAI Chat Completions, OpenAI Responses, and Anthropic Messages behind a single codec-layer path, so exporters and plugins consume normalized provider data instead of re-implementing provider parsing from raw JSON. This adds new public Rust core API (`codec::resolve`, `Event::normalized_llm_request` / `normalized_llm_response`); the change is additive and Rust-only, with no breaking changes.

Local validation: `cargo test -p nemo-relay` (638 passing), `cargo clippy -p nemo-relay --all-features --all-targets` clean, single-feature (`otel` / `openinference`) builds warning-free, and adaptive request-surface tests pass. The full cross-language matrix is left to CI.

#### Details

- **`codec::resolve`** (new) — the canonical detect-then-decode entry point: `ProviderSurface` + `detect_request_surface` (priority order hoisted from adaptive) / `detect_response_surface` (strict; ambiguous or empty payloads classify to `None`) + fail-open `normalize_request` / `normalize_response`.
- **`Event::normalized_llm_request` / `normalized_llm_response`** — annotation-first `Cow` helpers over the detector; `annotated_*()` left unchanged. Rustdoc documents the `{headers, content}` `LlmRequest` precondition for the request fallback.
- **`OpenAIResponsesCodec`** — response-only extension to decode top-level `output_text` and item-level `output[].type == "output_text"` (structured `output` keeps precedence).
- **`observability::manual`** (new) — the OpenTelemetry/OpenInference non-provider usage/cost fallback de-duplicated into one module, with the USD-only vs emit-any-currency policy parameterized.
- **OpenInference** — un-annotated provider payloads are decoded through the normalized helpers to emit structured request/response messages; usage merges codec-normalized fields (e.g. Anthropic's computed total) with the manual scraper per field so neither source's coverage is lost.
- **ATIF** — reads normalized message text from annotations when present (multimodal falls back to the raw extractor); text-less responses (tool-call-only, or reasoning/thinking-only) render an empty message instead of a raw-JSON dump, with tool calls and metrics still recovered from the raw path.
- **adaptive** — `resolve_request_surface_from_request` delegates to `codec::resolve::detect_request_surface`.
- Docs and docstrings: documented the annotation → normalized-helper → raw fallback rule and per-exporter behavior in `provider-response-codecs.mdx`; fixed the misleading "global codec registry" / "request codecs' priority order" docstrings and noted the detection limitation for `system`-less Anthropic requests.

`crates/core/src/codec/resolve.rs` (the single detect-then-decode path), then `crates/core/src/observability/atif.rs` and `openinference.rs` — the annotation-first / raw-fallback consumers are the trickiest equivalence surface, where the raw parser is intentionally retained because the normalized form would be lossy (tool-call extras, multimodal content, metrics).

- Relates to RELAY-325

## Release Notes

* **New Features**
  * Automatic detection and normalization of LLM provider requests and responses when provider-specific annotations are unavailable, with intelligent fallback handling.
  * Enhanced token usage and cost extraction from various LLM output formats, supporting multiple field aliases and currency handling.
  * Improved text extraction from OpenAI Responses payloads with better support for output_text fields.

* **Improvements**
  * Better preference ordering for request/response data: prioritizes annotated values, then normalized codec results, then raw payloads.
  * More robust message construction with multiple fallback strategies for different payload structures.


zhongxuanwang-nv:
@coderabbitai review

zhongxuanwang-nv:
@coderabbitai review

zhongxuanwang-nv:
/ok to test [633820a](https://github.com/NVIDIA/NeMo-Relay/pull/291/commits/633820ae7c9234c350a2412dc9f0c6de84b4a6e9)

crates/core/src/codec/resolve.rs
```diff
@@ -0,0 +1,94 @@
...
+use super::anthropic::AnthropicMessagesCodec;
+use super::openai_chat::OpenAIChatCodec;
+use super::openai_responses::OpenAIResponsesCodec;
+use super::request::AnnotatedLlmRequest;
+use super::response::AnnotatedLlmResponse;
+use super::traits::{LlmCodec, LlmResponseCodec};
+
+/// A built-in provider request/response surface.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ProviderSurface {
```
bbednarski9, line 20: Are we not planning to support Gemini generateContent as well? @willkill07

willkill07: We don't have anything related to Gemini-specific models currently in Relay. It would have to be a follow-up PR for the 0.6 release due to 0.5 already being scoped

bbednarski9:
The resolver itself is cheap, but the provider-aware ambiguity handling appears unused in the main adaptive translation path. translate_request knows provider, but build_semantic_request_view resolves from shape only. Should build_semantic_request_view take provider and call resolve_request_surface(provider, request) so Anthropic messages-only requests are decoded with AnthropicMessagesCodec?

crates/core/src/codec/resolve.rs
```diff
@@ -0,0 +1,94 @@
...
+///
+/// Priority: OpenAI Responses (`input`/`instructions`) > Anthropic Messages
+/// (`system`) > OpenAI Chat (`messages`). `None` when no key matches or `body`
+/// is not an object. This is a best-effort heuristic: an Anthropic request that
+/// omits the optional top-level `system` is indistinguishable from OpenAI Chat
+/// and classifies as `OpenAiChat`.
+#[must_use]
+pub fn detect_request_surface(body: &Json) -> Option<ProviderSurface> {
+    let obj = body.as_object()?;
+    if obj.contains_key("input") || obj.contains_key("instructions") {
```
bbednarski9, line 39: simple field lists probably are not expressive enough by themselves. I’d push for codec-owned detector metadata or detector functions, not just hard-coded logic centralized here

bbednarski9: In other words... Could this PR move provider-surface detection into codec-owned descriptors and have codec::resolve iterate a small built-in registry? That would keep resolve.rs provider-agnostic, make future codecs like Gemini additive, and let provider hints resolve ambiguous shapes near the codec that owns the ambiguity.

crates/core/src/codec/resolve.rs
bbednarski9: Alright, out of scope for now.

bbednarski9:
AI code review comment, but please validate: 
In crates/core/src/observability/openinference.rs:733, end_attributes calls event.normalized_llm_response() to get usage. Later, push_llm_response_attributes calls it again in crates/core/src/observability/openinference.rs:836.

For unannotated provider-shaped responses, that means two full detect/decode passes on the same event.

For unannotated provider outputs, end_attributes decodes once for usage and push_llm_response_attributes decodes again for response fields. Can we compute the normalized response once in end_attributes and pass it through, or cache it in a small local view, so OpenInference does not do duplicate provider decode work per LLM end event?

bbednarski9 (changes requested):

crates/core/src/codec/resolve.rs
zhongxuanwang-nv: Thanks Bryan! I do agree with this; I only did this because that's essentially what the old code did, and considering that this PR's ticket is only asking refactoring, what about scoping this task to a follow-up PR instead?

bbednarski9: Follow-up PR sounds good. Can you make an issue to track that follow-up item?

zhongxuanwang-nv: Will do!

zhongxuanwang-nv:
> The resolver itself is cheap, but the provider-aware ambiguity handling appears unused in the main adaptive translation path. translate_request knows provider, but build_semantic_request_view resolves from shape only. Should build_semantic_request_view take provider and call resolve_request_surface(provider, request) so Anthropic messages-only requests are decoded with AnthropicMessagesCodec?

I think this is intentional. In `acg_component` we intentionally pick the request surface from the shape, and there's a test `acg_component_source_resolves_request_surfaces_instead_of_decoding_from_provider` guarding this. So an ambiguous "message_only" request that could be from OpenAI or Anthropic deliberately passes through untranslated instead of guessing from the provider. Making it provider-aware may reverse that design?

> AI code review comment, but please validate: In crates/core/src/observability/openinference.rs:733, end_attributes calls event.normalized_llm_response() to get usage. Later, push_llm_response_attributes calls it again in crates/core/src/observability/openinference.rs:836.
> 
> For unannotated provider-shaped responses, that means two full detect/decode passes on the same event.
> 
> For unannotated provider outputs, end_attributes decodes once for usage and push_llm_response_attributes decodes again for response fields. Can we compute the normalized response once in end_attributes and pass it through, or cache it in a small local view, so OpenInference does not do duplicate provider decode work per LLM end event?

Indeed it is something my later changes created.. Just fixed! Thanks Bryan!

willkill07:
> > The resolver itself is cheap, but the provider-aware ambiguity handling appears unused in the main adaptive translation path. translate_request knows provider, but build_semantic_request_view resolves from shape only. Should build_semantic_request_view take provider and call resolve_request_surface(provider, request) so Anthropic messages-only requests are decoded with AnthropicMessagesCodec?
> 
> I think this is intentional. In acg_component we intentionally pick the request surface from the shape, and there's a test acg_component_source_resolves_request_surfaces_instead_of_decoding_from_provider guarding this. So an ambiguous "message_only" request that could be from OpenAI or Anthropic deliberately passes through untranslated instead of guessing from the provider. Making it provider-aware may reverse that design?

The design was that ACG adopted the provider codec first. If we are extending Relay to use provider codecs in more places, we can change the surface to pass this in. The consequences of this are that (1) ACG is really only support with Anthropic and (2) we need to ensure compatibility without breaking the existing public configuration surface.

zhongxuanwang-nv:
> > > The resolver itself is cheap, but the provider-aware ambiguity handling appears unused in the main adaptive translation path. translate_request knows provider, but build_semantic_request_view resolves from shape only. Should build_semantic_request_view take provider and call resolve_request_surface(provider, request) so Anthropic messages-only requests are decoded with AnthropicMessagesCodec?
> > 
> > 
> > I think this is intentional. In acg_component we intentionally pick the request surface from the shape, and there's a test acg_component_source_resolves_request_surfaces_instead_of_decoding_from_provider guarding this. So an ambiguous "message_only" request that could be from OpenAI or Anthropic deliberately passes through untranslated instead of guessing from the provider. Making it provider-aware may reverse that design?
> 
> The design was that ACG adopted the provider codec first. If we are extending Relay to use provider codecs in more places, we can change the surface to pass this in. The consequences of this are that (1) ACG is really only support with Anthropic and (2) we need to ensure compatibility without breaking the existing public configuration surface.

That makes sense! I just dropped the unused provider-aware tie-break code, and since that making `build_semantic_request_view` would be a behavior change (starts translating system-less Ant requests and changes the tests), could I do this also in a follow-up PR, so we can get this PR merged ASAP which is blocking Maryam's PR? Thanks Will!

docs/integrate-into-frameworks/provider-response-codecs.mdx
```diff
@@ -479,6 +479,14 @@ registerSubscriber('response-debugger', (event) => {
...
+
+For events without one — manual lifecycle spans, replayed payloads, or unrecognized shapes — reach for the normalized helpers before hand-parsing provider JSON. In Rust, `Event::normalized_llm_request` / `normalized_llm_response` return the annotation when present, and otherwise attempt a best-effort detect-then-decode of the raw payload via `codec::resolve`. That decode only succeeds for provider-shaped payloads; manual or unrecognized shapes yield nothing, so you read the raw payload directly.
+
+Built-in exporters apply this to their own needs rather than following one fixed precedence. OpenInference decodes un-annotated provider payloads this way to emit structured messages, and recovers a provider-neutral view (model, tokens, cost) from the raw payload for manual outputs. ATIF reads normalized message text from the annotation when present, but deliberately keeps un-annotated events on the raw extractor so provider-specific tool-call extras and metrics are preserved. Reserve hand-written raw-payload parsing for the fields the normalized path does not provide.
+
```
willkill07: I'm not sure if we need an implementation detail surfaced to the documentation.

crates/core/src/observability/atif.rs
```diff
@@ -2212,7 +2243,10 @@ impl StepConversionState {
         self.steps.push(AtifStep {
             step_id: 0,
             source: "user".to_string(),
-            message: extract_user_messages(&content),
+            message: event
+                .annotated_request()
+                .and_then(|request| atif_message_from_annotated_request(request))
+                .unwrap_or_else(|| extract_user_messages(&content)),
```
willkill07, line 2249: Do we know how often it fails to reach the `unwrap_or_else` ? Ideally we should be able to get rid of `extract_user_messages`

crates/core/src/observability/atif.rs
```diff
@@ -2253,7 +2287,10 @@ impl StepConversionState {
         self.steps.push(AtifStep {
             step_id: 0,
             source: "agent".to_string(),
-            message: extract_llm_response_message(output),
+            message: event
+                .annotated_response()
+                .and_then(|response| atif_message_from_annotated_response(response))
+                .unwrap_or_else(|| extract_llm_response_message(output)),
```
willkill07, line 2293: Do we know how often it fails to reach the `unwrap_or_else` ? Ideally we should be able to get rid of `extract_llm_response_message`

crates/core/src/observability/manual.rs
willkill07, line 1: long term / follow-up PR: this should be hoisted to the `AnnotatedResponse` object

willkill07 (approved):

docs/integrate-into-frameworks/provider-response-codecs.mdx
zhongxuanwang-nv: Thanks Will!! That's true! In the future I would only keep the public + user needed functions in the user docs, and keep other things only as code comments when necessary. This is also an oversight of me — indeed this bloats the docs!

crates/core/src/observability/atif.rs
zhongxuanwang-nv: (responded in the last comment (line 2246-2249))

crates/core/src/observability/atif.rs
zhongxuanwang-nv: Thanks Will! I think I kept the fallback because there might be cases (notably multimodal, and also possibly non-provider / custom shapes / manual spans) that may need the fallback? Also, `extract_user_messages/extract_llm_response_message` builds on the start/end correction signatures, which run on every event, and they're the only path for un-annotated/non-provider/multimodal payloads.

I think for compatibility purposes, we might need to keep them?

zhongxuanwang-nv:
/ok to test [a369cd4](https://github.com/NVIDIA/NeMo-Relay/pull/291/commits/a369cd4a012821b77c0445792500716615e64bc8)

bbednarski9 (approved):

crates/core/src/codec/resolve.rs
```diff
@@ -0,0 +1,94 @@
...
+use super::response::AnnotatedLlmResponse;
+use super::traits::{LlmCodec, LlmResponseCodec};
+
+/// A built-in provider request/response surface.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ProviderSurface {
+    /// OpenAI Chat Completions.
+    OpenAiChat,
+    /// OpenAI Responses.
+    OpenAiResponses,
```
willkill07: Small consistency nit:

```suggestion
    OpenAIChat,
    /// OpenAI Responses.
    OpenAIResponses,
```

willkill07 (changes requested):

zhongxuanwang-nv:
/ok to test [e307e8d](https://github.com/NVIDIA/NeMo-Relay/pull/291/commits/e307e8dc760a730cb0708e7d549a278225b0b776)

willkill07 (approved):

PR #296, opened 2026-06-24
fix: keep late mark spans in parent traces

bbednarski9:
#### Overview

This PR keeps late mark spans attached to the trace of their completed parent scope in the OpenTelemetry and OpenInference exporters. When a mark event arrives after the parent span has already ended, the exporters now reuse the completed parent's span context instead of starting a new root trace.

#### Details

- Add a bounded completed-span-context cache to the OpenTelemetry event processor.
- Add the same completed-span-context cache to the OpenInference event processor.
- Preserve the existing active-parent lookup path before falling back to completed parent contexts.
- Preserve the existing true-orphan mark behavior when no active or completed parent context exists.
- Add regression tests that cover mark events emitted after a tool span has completed.

Validation:

- `cargo fmt --check`
- `cargo test -p nemo-relay observability::otel::tests::`
- `cargo test -p nemo-relay observability::openinference::tests::`
- `cargo test -p nemo-relay late_parented_marks_reuse_completed_parent_trace_context`

Start with `crates/core/src/observability/otel.rs` and `crates/core/src/observability/openinference.rs`, specifically the parent context lookup and completed span context cache. The matching regression tests are in `crates/core/tests/unit/observability/otel_tests.rs` and `crates/core/tests/unit/observability/openinference_tests.rs`.

- Relates to: RELAY-360

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved observability span lineage for late or out-of-order events by reusing trace context from recently completed spans, ensuring correct parent-child relationships and accurate remote-parent handling.
  * OTEL NeMo Relay now preserves trace parentage for orphan/late “End” and “Mark” scenarios using a bounded in-memory cache.

* **Tests**
  * Added regression coverage to verify marks correctly reuse the completed parent’s trace context.
  * Added unit coverage for cache bookkeeping when spans are restarted.

### DIFF
--- a/crates/core/src/observability/otel.rs
+++ b/crates/core/src/observability/otel.rs
@@ -16,7 +16,7 @@
 //! - [`OpenTelemetrySubscriber`] exposes a NeMo Relay [`EventSubscriberFn`] and
 //!   convenience `register` / `deregister` / `force_flush` / `shutdown` methods
 
-use std::collections::HashMap;
+use std::collections::{HashMap, VecDeque};
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
@@ -38,6 +38,8 @@ use opentelemetry_sdk::Resource;
 use opentelemetry_sdk::trace::{SdkTracer, SdkTracerProvider, Span};
 use serde::Serialize;
 use uuid::Uuid;
+
+const COMPLETED_SPAN_CONTEXT_LIMIT: usize = 4096;
 
 #[cfg(target_arch = "wasm32")]
 use async_trait::async_trait;
@@ -494,6 +496,8 @@ struct ActiveSpan {
 
 struct OtelEventProcessor {
     active_spans: HashMap<Uuid, ActiveSpan>,
+    completed_span_contexts: HashMap<Uuid, SpanContext>,
+    completed_span_order: VecDeque<Uuid>,
     provider: SdkTracerProvider,
     tracer: SdkTracer,
 }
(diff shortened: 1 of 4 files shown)


crates/core/src/observability/otel.rs
```diff
@@ -578,10 +586,13 @@ impl OtelEventProcessor {
...
-        self.find_parent_span(event)
-            .map(|active_span| {
-                Context::new().with_remote_span_context(active_span.span_context.clone())
-            })
+        if let Some(active_span) = self.find_parent_span(event) {
+            return Context::new().with_remote_span_context(active_span.span_context.clone());
+        }
+        event
+            .parent_uuid()
+            .and_then(|uuid| self.completed_span_contexts.get(&uuid))
```
mnajafian-nv, line 594: This fallback looks right for keeping late marks inside the completed parent trace. One nearby semantic question: after we reuse the completed parent context, the mark is no longer trace-orphaned, but it is still emitted as a standalone mark span because the active parent span is gone. Is `nemo_relay.mark.orphan=true` intended to mean "not attached to a live span as an event" rather than "no known parent"? If yes, could we capture that in a short comment or follow-up so downstream readers do not interpret it as parentless?

mnajafian-nv:
This seems like the right small fix to land ahead of the larger exporter/normalization refactors. One thing I would like us to preserve when #291/#293 rebase over this area is the completed-parent context fallback, so late mark spans continue to stay enclosed in the original parent trace.

mnajafian-nv (approved):
LGTM, pending addressing inline suggestions and green CI :)

willkill07 (approved):
Coderabbit feedback seems sensible to resolve. Otherwise LGTM.

bbednarski9:
Ack @mnajafian-nv will check for conflicts on those PRs when this is in and comment if anything looks off

mnajafian-nv (approved):
Thanks, LGTM.

willkill07 (approved):

bbednarski9:
Thanks @willkill07

PR #300, opened 2026-06-24
refactor: align exporter cost fallback with normalized responses

mnajafian-nv:
#### Overview

Aligns OTel and OpenInference cost fallback extraction with normalized LLM responses, so provider-shaped responses can use the codec-normalized path before falling back to raw/manual parsing.

#### Details

- Uses `normalized_llm_response()` for OTel and OpenInference cost fallback where available.
- Keeps exporter-specific projection and schema logic local.
- Preserves manual/raw fallback behavior when normalized response data is unavailable.
- Falls back to the requested model when the normalized/provider-echoed model is not present in pricing, preventing silent cost loss for aliased model names.
- Adds regression coverage for unannotated OpenAI-shaped responses to ensure provider-aware normalized cost extraction is used instead of event-name-only fallback behavior.
- Adds negative coverage for normalized provider-shaped responses without usage, so exporters do not fabricate cost attributes.
- Does not change ATIF projection, usage/token extraction, model-name extraction, tool-call naming, or exporter output schemas.

Validation:
- `cargo test -p nemo-relay unannotated_openai_response -- --nocapture`
- `cargo test -p nemo-relay unpriced_response_model -- --nocapture`
- `cargo test -p nemo-relay observability::otel::tests -- --nocapture`
- `cargo test -p nemo-relay observability::openinference::tests -- --nocapture`
- `cargo check -p nemo-relay --no-default-features`
- `cargo check -p nemo-relay --no-default-features --features otel`
- `cargo check -p nemo-relay --no-default-features --features openinference`
- `cargo llvm-cov test --no-report -p nemo-relay observability::otel::tests`
- `cargo llvm-cov test --no-report -p nemo-relay observability::openinference::tests`
- Targeted `cargo llvm-cov` report confirmed all added production lines in `otel.rs` and `openinference.rs` are hit.

Start with the shared fallback helper in `crates/core/src/observability/mod.rs`, then review its use in `crates/core/src/observability/otel.rs` and `crates/core/src/observability/openinference.rs`. The focused regression coverage is in the matching observability test files.

No public issue.

### DIFF
--- a/crates/core/src/observability/otel.rs
+++ b/crates/core/src/observability/otel.rs
@@ -20,7 +20,7 @@ use std::collections::{HashMap, VecDeque};
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
-use super::manual;
+use super::{estimate_cost_for_response_or_requested_model, manual};
 use crate::api::event::Event;
 use crate::api::event::ScopeCategory;
 use crate::api::runtime::EventSubscriberFn;
@@ -722,9 +722,12 @@ fn cost_from_llm_event(event: &Event) -> Option<(f64, String)> {
             if let Some(cost) = usage.cost.as_ref() {
                 return cost_total_and_currency(cost);
             }
-            if let Some(model_name) = response.model.as_deref().or_else(|| event.model_name()) {
-                return estimate_cost_for_provider(Some(event.name()), model_name, usage)
-                    .and_then(|cost| cost_total_and_currency(&cost));
+            if let Some(cost) = estimate_cost_for_response_or_requested_model(
+                event,
+                response.model.as_deref(),
+                usage,
+            ) {
+                return cost_total_and_currency(&cost);
             }
         }
     }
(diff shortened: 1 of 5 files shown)


crates/core/src/observability/otel.rs
```diff
@@ -716,15 +716,16 @@ fn cost_from_llm_event(event: &Event) -> Option<(f64, String)> {
...
-                .and_then(|cost| cost_total_and_currency(&cost));
+    if let Some(response) = event.normalized_llm_response() {
+        let response = response.as_ref();
+        if let Some(usage) = response.usage.as_ref() {
+            if let Some(cost) = usage.cost.as_ref() {
+                return cost_total_and_currency(cost);
+            }
+            if let Some(model_name) = response.model.as_deref().or_else(|| event.model_name()) {
+                return estimate_cost_for_provider(Some(event.name()), model_name, usage)
+                    .and_then(|cost| cost_total_and_currency(&cost));
```
zhongxuanwang-nv: Nitty thing, here the code prices the model the API echoed back; if that exact name isn't in the price list already, seems like the API would give up and report there would be no cost, instead of falling back to the model requested?

So a fix might be to try the requested name before giving up, so the cost data would not be silently lost.

zhongxuanwang-nv (approved):
Thanks Maryam! I think this LGTM!

willkill07 (approved):

PR #309, opened 2026-06-25
feat(worker): activate grpc-v1 worker plugins in relay

willkill07:
#### Overview

Add the host activation and proxy layer for `grpc-v1` worker dynamic plugins.

Stack dependency: this PR cannot be merged until parent stacked PR #308 is merged: https://github.com/NVIDIA/NeMo-Relay/pull/308

- [x] I confirm this contribution is my own work, or I have the right to submit it under this project's license.
- [x] I searched existing issues and open pull requests, and this does not duplicate existing work.

#### Details

- Add the core worker plugin activation/proxy implementation behind `worker-grpc`.
- Start local workers, perform handshake/validation/registration, and proxy every supported registration surface.
- Add worker fixture and integration tests for callbacks, continuations, streaming, cancellation, marks, scopes, and teardown.

Validation run across the completed stack:

- `cargo test -p nemo-relay-types`
- `cargo test -p nemo-relay-plugin`
- `cargo test -p nemo-relay-worker-proto`
- `cargo test -p nemo-relay-worker`
- `cargo check -p nemo-relay-cli`

#### Where should the reviewer start?

Start with `crates/core/src/plugin/dynamic/worker.rs`, then review `crates/core/tests/integration/worker_plugin_tests.rs`.

#### Related Issues: (use one of the action keywords Closes / Fixes / Resolves / Relates to)

- Relates to: #302
- Relates to: #230

## Summary by CodeRabbit

* **New Features**
  * Added dynamic worker plugin support over gRPC (**worker-grpc**) and enabled it in both core and the CLI.
  * Worker plugins can now be dynamically loaded and participate in tool, LLM, and streaming request handling.
  * Added a worker runtime helper to run code within a specific scope stack.
* **Bug Fixes**
  * Improved worker plugin activation/teardown, including clearer validation and compatibility handling.
  * Enhanced tool execution marking to reflect isolated and restored scope-stack phases.
* **Tests**
  * Added comprehensive unit and integration coverage, including a dedicated worker plugin fixture and streaming/error scenarios.

PR #230, opened 2026-06-05
feat: add explicit plugin config file override

bbednarski9:
#### Overview

Add `--plugin-config-file <path>` as an explicit TOML plugin configuration source for gateway startup and `nemo-relay run`.

This gives operators, demos, and harnesses a file-based override path for large plugin configs without shell-escaping JSON through `--plugin-config` or relying on sibling/project/user `plugins.toml` discovery.

- [x] I confirm this contribution is my own work, or I have the right to submit it under this project's license.
- [ ] I searched existing issues and open pull requests, and this does not duplicate existing work.

#### Details

- Added `--plugin-config-file <path>` and `NEMO_RELAY_PLUGIN_CONFIG_FILE`.
- Added it to both daemon startup args and `nemo-relay run`.
- It loads TOML from the exact file path.
- It skips normal discovered `plugins.toml` loading when present.
- It conflicts with `--plugin-config`.
  These two explicit plugin config sources are mutually exclusive:
  ```bash
  nemo-relay \
    --plugin-config '{"version":1,"components":[]}' \
    --plugin-config-file /path/to/plugins.toml
  ```
  Relay fails fast with:
  ```text
  configuration error: choose only one of --plugin-config or --plugin-config-file
  ```
- `[plugins].config` in `config.toml` still conflicts with plugin TOML sources, including the explicit file.
- Missing or invalid plugin TOML produces a startup config error.
- Updated plugin configuration docs to describe the new source and conflict behavior.
- Added focused config-resolution coverage for override behavior, missing/invalid files, and conflict cases.

Validation run:

```text
cargo test -p nemo-relay-cli config::tests::
cargo test -p nemo-relay-cli cli_help_exits_successfully
```

Manual E2E startup checks:

- Started Relay with `--plugin-config-file` while a sibling `plugins.toml` was intentionally invalid.
- Confirmed `/healthz` returned `{"status":"ok"}`.
- Confirmed `NEMO_RELAY_PLUGIN_CONFIG_FILE` works.
- Confirmed startup errors for `--plugin-config` plus `--plugin-config-file`.
- Confirmed startup errors for `[plugins].config` plus `--plugin-config-file`.

#### Where should the reviewer start?

Start with `crates/cli/src/config.rs`, especially the config-source validation and explicit plugin TOML loading path.

Then review `crates/cli/tests/coverage/config_tests.rs` for the intended precedence and conflict behavior.

#### Related Issues: (use one of the action keywords Closes / Fixes / Resolves / Relates to)

[...]

PR #310, opened 2026-06-25
feat(python): add nemo-relay-plugin worker SDK package

willkill07:
#### Overview

Add the pure-Python `nemo-relay-plugin` worker SDK package and its packaging/test support.

Stack dependency: this PR cannot be merged until parent stacked PR #309 is merged: https://github.com/NVIDIA/NeMo-Relay/pull/309

#### Details

- Add `python/plugin` with high-level worker plugin APIs, generated gRPC stubs, runtime helpers, continuations, and streaming helpers.
- Add Python SDK tests and a Python gRPC worker example.
- Update Python test, coverage, packaging, CI wheel upload, attribution, and pre-commit configuration for the new package.

Validation run across the completed stack:

- `cargo test -p nemo-relay-types`
- `cargo test -p nemo-relay-plugin`
- `cargo test -p nemo-relay-worker-proto`
- `cargo test -p nemo-relay-worker`
- `cargo check -p nemo-relay-cli`

Start with `python/plugin/src/nemo_relay_plugin/_api.py`, then review `python/tests/plugin/test_worker_sdk.py`.

- Relates to: #302
- Relates to: #230

## Summary by CodeRabbit

* **New Features**
  * Added a Python plugin SDK for gRPC v1 dynamic worker plugins, plus a working Python gRPC worker example.
  * CI can now build and publish an additional Linux-only Python plugin SDK wheel artifact.
* **Documentation**
  * Added authoring and usage guides for the SDK and the example plugin.
* **Tests**
  * Expanded plugin end-to-end, example integration, public API docstring coverage, and packaging/build (sdist rebuild) tests.
* **Bug Fixes**
  * Improved platform-aware CI/test skipping and coverage handling for generated protobuf sources.

### DIFF
--- a/python/plugin/README.md
+++ b/python/plugin/README.md
@@ -1,0 +1,49 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# nemo-relay-plugin
+
+Python authoring SDK for NeMo Relay out-of-process dynamic worker plugins.
+
+Install this package in the Python environment used by a worker manifest with
+`load.runtime = "python"`, then expose a `module:function` entrypoint that calls
+`serve_plugin`.
+
+```python
+from nemo_relay_plugin import Json, PluginContext, WorkerPlugin, serve_plugin
+
+
+class PolicyPlugin(WorkerPlugin):
+    plugin_id = "acme.policy"
+
+    def register(self, ctx: PluginContext, config: Json) -> None:
+        async def tag_tool_request(tool_name: str, args: Json) -> Json:
+            await ctx.runtime.emit_mark("acme.policy.tool_request", {"tool_name": tool_name})
+            if isinstance(args, dict):
+                return {**args, "policy": "checked"}
+            return {"value": args, "policy": "checked"}
+
+        ctx.register_tool_request_intercept("tag_tool_request", tag_tool_request)
+
+
+async def main() -> None:
+    await serve_plugin(PolicyPlugin())
+```
+
+Set `load.entrypoint` to `your_module:main` in `relay-plugin.toml`. Relay
+imports that function and awaits the returned coroutine when it starts the
+worker process.
+
+The SDK owns gRPC serving, JSON envelope conversion, callback dispatch,
+continuations, host runtime calls, and local scope-stack binding. Its private
+protobuf bindings are generated from the canonical Relay schema while the
+package is built; they are included in installed wheels but are not committed
+to the source repository. Installing the published wheel never requires
+`protoc` or `grpcio-tools`.
+
+Windows ARM64 is not currently supported because `grpcio` does not publish a
+usable wheel for that platform. The NeMo Relay workspace skips installation and
+tests for this SDK on Windows ARM64 rather than creating a package without its
+required gRPC runtime.
--- a/python/plugin/README.md
+++ b/python/plugin/README.md
@@ -43,6 +43,19 @@ package is built; they are included in installed wheels but are not committed
 to the source repository. Installing the published wheel never requires
 `protoc` or `grpcio-tools`.
 
+## Callback Concurrency
+
+The gRPC AsyncIO server can keep multiple RPCs in flight. Callback execution is
+cooperative: asynchronous callbacks overlap only when they yield control at an
+`await`. Synchronous callbacks and synchronous stream iterators run on the
+worker event-loop thread. Blocking I/O, `time.sleep`, or long-running CPU work
+in those callbacks stalls all worker RPCs. Wrap blocking work in an
+asynchronous callback and offload it with `asyncio.to_thread()` or another
+appropriate executor.
+
+The SDK does not configure `maximum_concurrent_rpcs`, so gRPC does not enforce
+an application-level RPC admission limit.
+
 Windows ARM64 is not currently supported because `grpcio` does not publish a
 usable wheel for that platform. The NeMo Relay workspace skips installation and
 tests for this SDK on Windows ARM64 rather than creating a package without its
--- a/python/plugin/src/nemo_relay_plugin/_api.py
+++ b/python/plugin/src/nemo_relay_plugin/_api.py
@@ -1,0 +1,1683 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""High-level Python API for NeMo Relay ``grpc-v1`` worker plugins.
+
+The module exposes the authoring contract for out-of-process Python plugins.
+Callbacks can be synchronous or asynchronous unless a method documents a more
+specific return type. Callback exceptions are returned to the Relay host as
+structured worker errors.
+
+Public data types:
+    Json: Any JSON-serializable Python value.
+    Event: A Relay event represented as a JSON object.
+    LlmRequest: A Relay LLM request represented as a JSON object.
+    AnnotatedLlmRequest: An annotated Relay LLM request represented as a JSON
+        object.
+    DiagnosticLevel: Severity of a configuration diagnostic.
+    ConfigDiagnostic: Structured configuration warning or error.
+    ScopeType: Semantic category for a Relay execution scope.
+    WorkerSdkError: SDK, host-call, or worker protocol error.
+
+Public authoring types:
+    WorkerPlugin: Base validation and registration contract for a plugin.
+    PluginContext: Component-scoped callback registration context.
+    PluginRuntime: Host runtime handle for event and scope operations.
+    ToolNext: Continuation for a tool execution intercept.
+    LlmNext: Continuation for a unary LLM execution intercept.
+    LlmStreamNext: Continuation for a streaming LLM execution intercept.
+
+Public callback aliases used in registration annotations:
+    SubscriberCallback: Event subscriber callback.
+    ToolSanitizeCallback: Tool request or response sanitizer callback.
+    ToolConditionalCallback: Tool execution guardrail callback.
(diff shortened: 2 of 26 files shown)


Salonijain27 (approved):

python/plugin/src/nemo_relay_plugin/_api.py
```diff
@@ -0,0 +1,1683 @@
...
+            with _bind_invocation_scope(request):
+                stream = await _maybe_await(handler(payload.model_name, llm_request, next_call))
+                async for value in _as_async_iter(stream):
+                    yield pb.StreamChunk(value=_json_envelope(JSON_SCHEMA, value))
+        except Exception as exc:  # noqa: BLE001 - callback failure is protocol data.
+            yield pb.StreamChunk(error=_sdk_error_to_worker(exc))
+
+    async def CancelInvocation(self, request: Any, context: Any) -> Any:
+        await self._authorize(request, context)
+        return pb.WorkerAck(accepted=False, message="cancel is not implemented by the Python worker SDK")
```
bbednarski9, line 1263: Cancellation is explicitly not accepted by python workers as shown here. Is this intentional? Probably OK for a Beta but will not have complete support for python plugins. Long-running plugin processes therefore would differ to the standard gRPC timeout (30 seconds?). This could be a gap that we either need to patch or document

python/plugin/src/nemo_relay_plugin/_api.py
```diff
@@ -0,0 +1,1683 @@
...
+    await _unlink_unix_socket(worker_endpoint)
+    host_channel = grpc.aio.insecure_channel(host_target)
+    runtime = PluginRuntime(
+        activation_id=activation_id,
+        auth_token=auth_token,
+        host_stub=pb_grpc.RelayHostRuntimeStub(host_channel),
+    )
+    shutdown_event = asyncio.Event()
+    service = _WorkerService(plugin, runtime, shutdown_event)
+    server = grpc.aio.server()
```
bbednarski9, line 1146: This is a single server configured without a concurrency/backpressure contract, i.e. how many concurrent gRPC processes are we willing to tollerate concurrently. Implementing this could be significant, but if the server invokes multiple sync gRPC requests against the worker it would be blocking. If rust is able to originate concurrent gPRC requests, then we dont get the same behavior from python plugins

python/plugin/src/nemo_relay_plugin/_api.py
willkill07: We can probably add this in, but should be follow-up since we don't even support it on the Rust side.

python/plugin/src/nemo_relay_plugin/_api.py
willkill07: > If rust is able to originate concurrent gPRC requests, then we dont get the same behavior from python plugins

That's not how this works. The gRPC worker process is owned by the NeMo Relay runtime. The worker's "server" is the NeMo Relay runtime itself.

EDIT: I conflated process ownership with the gRPC direction. Relay owns the worker process, but the Python worker exposes the PluginWorker server and Relay invokes it as a client. Async callbacks can overlap, while blocking synchronous callbacks can stall the worker’s AsyncIO event loop. We should document that execution model here and track configurable concurrency/backpressure separately.

python/plugin/src/nemo_relay_plugin/_api.py
```diff
@@ -0,0 +1,1683 @@
...
+            plugin_id=plugin_id,
+            plugin_kind=plugin_id,
+            allows_multiple_components=bool(getattr(self._plugin, "allows_multiple_components", False)),
+            worker_protocol=WORKER_PROTOCOL,
+            sdk_name="nemo-relay-plugin",
+            sdk_version=_SDK_VERSION,
+            runtime_name="python",
+            runtime_version=platform.python_version(),
+            supported_surfaces=_all_surfaces(),
+        )
```
bbednarski9, line 1189: As we look towards similar support for Go and C, does this feel a bit brittle to you as well? Perhaps we have the handshake negotitate a versioned worker semantic contract - rather than only reporting protocol and surface metadata (literal grpc values)? I know it's a bit more code but we could use this surface for extending to multi-lang

Im thinking mostly about the host defining the failure policy, but could be extended to registration surfaces and cancellation policies, and even serialization (json app data not gprc)

python/plugin/src/nemo_relay_plugin/_api.py
```diff
@@ -0,0 +1,1683 @@
...
+            llm_request = _decode_required_envelope(payload.request, "llm request", LLM_REQUEST_SCHEMA)
+            next_call = LlmStreamNext(self._runtime, request.continuation_id)
+            with _bind_invocation_scope(request):
+                stream = await _maybe_await(handler(payload.model_name, llm_request, next_call))
+                async for value in _as_async_iter(stream):
+                    yield pb.StreamChunk(value=_json_envelope(JSON_SCHEMA, value))
+        except Exception as exc:  # noqa: BLE001 - callback failure is protocol data.
+            yield pb.StreamChunk(error=_sdk_error_to_worker(exc))
+
+    async def CancelInvocation(self, request: Any, context: Any) -> Any:
```
bbednarski9, line 1261: Cancellation is explicitly not accepted by python workers as shown here. Is this intentional? Probably OK for a Beta but will not have complete support for python plugins. Long-running plugin processes therefore would differ to the standard gRPC timeout (30 seconds?). This could be a gap that we either need to patch or document

python/plugin/README.md
```diff
@@ -0,0 +1,49 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# nemo-relay-plugin
```
bbednarski9, line 6:
```suggestion
# nemo-relay-plugin (Python)
```

python/plugin/src/nemo_relay_plugin/_api.py
willkill07: A bit brittle, yes. But I don't _want_ to support Go and C for plugins. Folks should write it in Rust through a dynamic embeddable plugin if they need the speed. Otherwise, they can write a plugin in Python. Adding options just for the sake of having them isn't a good approach for the long-term.

python/plugin/README.md
willkill07: Unnecessary. This is a published python package and (Python) is implicit.

python/plugin/src/nemo_relay_plugin/_api.py
willkill07: Already mentioned above. Resolving.

python/plugin/src/nemo_relay_plugin/_api.py
bbednarski9: Alright cool, I wasnt sure if we were going to support Go and C for plugins so this is fine

python/plugin/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+Python authoring SDK for NeMo Relay out-of-process dynamic worker plugins.
+
+Install this package in the Python environment used by a worker manifest with
```
bbednarski9, line 10: Looking at worker.rs (lines 686-695) python uses a fixed interpreter. When `load.runtime = "python"` the worker spawns interpreter from NEMO_RELAY_PYTHON envar. Therefore there is no per-plugin interpreter/venv field. Two plugins with incompatible dependencies cannot independently select virtual environments unless they use a custom command entrypoint.

Flagging in case we want to: Document? Fill gap? Raise issue?

python/plugin/src/nemo_relay_plugin/_api.py
bbednarski9: Created issue #334 to track this. Can close if we decide not to do it.

python/plugin/README.md
willkill07: You’re right.

`plugins add` currently registers the manifest but does not provision a virtual environment.
All `runtime = "python"` workers use the gateway-wide `NEMO_RELAY_PYTHON` value, falling back to `python3`. Modeled `environment_ref` is not populated or consumed yet.

I will have a follow-up PR that makes this work as intended, with proper isolation for each plugin. I encountered this through manual testing.

bbednarski9:
Are we missing a happy-path unit test for python worker spin up in worker_plugin_tests.rs?

bbednarski9 (approved):
Approving as comments are either follow-ups or missing test cases. Can re-approve if there's a follow-up commit.

python/plugin/src/nemo_relay_plugin/_api.py
willkill07: Added documentation as a follow-up commit

bbednarski9:
Codex results from E2E testing:

Retested the latest PR head, `b889940`, using Python 3.13 and the example at `examples/python-grpc-worker-plugin`.

Passed:
- The editable `nemo-relay-plugin` SDK and example packages built and installed successfully.
- The example module and worker entrypoint imported successfully.
- In a Rust-hosted integration harness, the Python worker completed process startup, health check, authenticated handshake, configuration validation, and declarative registration.
- The equivalent Rust worker integration test, including host-runtime mark emission, passed.

Failed:
- The documented `nemo-relay plugins add ./relay-plugin.toml` command exits with code 3 because the example manifest is missing `source.artifact` and the corresponding integrity metadata required by CLI validation.
- The documented `nemo-relay gateway` command exits with code 2 because the current CLI has no `gateway` subcommand.
- The first real tool-request callback fails when the example awaits `ctx.runtime.emit_mark(...)`. The worker returns `StatusCode.INTERNAL` with `Stream removed (Received RST_STREAM with error code 1 (RST_STREAM))`. Because this happens before the callback returns, the mutated/tagged tool request is not returned to the Rust host.

The latest commit only adds callback-concurrency documentation, so these results are unchanged from the prior head. Overall, the example is not currently nominal end to end.

PR #314, opened 2026-06-25
fix(cli): parse codex `plugin list` text output instead of --json

### LINKED ISSUE #313: [Bug]: codex integration error: unexpected argument '--json' found
### Affected area

Plugins

### Current behavior

Codex CLI is codex-cli 0.136.0, and its codex plugin list has no --json flag (only -c, -m/--marketplace, --enable, --disable)

However, when doing nemo-relay global install, it tries to call codex with the --json flag, resulting an error

### Expected behavior

No errors with global install

### Steps to reproduce

cargo install nemo-relay-cli
nemo-relay plugins edit
nemo-relay install all
nemo-relay doctor --plugin all

### Environment

Relay is 0.4.0 on macos
Codex CLI is codex-cli 0.136.0

### Impact

Will implement a workaround myself

zhongxuanwang-nv:
#### Overview

`nemo-relay doctor --plugin codex` (and `--plugin all`) fails on Codex CLI 0.136.0 with `error: unexpected argument '--json' found`. The codex plugin host-registration probe calls `codex plugin list --json`, but current Codex `plugin list` has no `--json` flag (only `-c`, `-m/--marketplace`, `--enable`, `--disable`).

#### Details

- `codex_plugin_registered` now runs `codex plugin list` (no `--json`) and parses the text table, mirroring the existing `codex_marketplace_registered` text parsing. A small helper `codex_plugin_line_installed` matches the row whose first column is `nemo-relay-plugin@nemo-relay-local` and whose status column starts with `installed` (so `installed, enabled` matches but `not installed` does not).
- Claude Code is unchanged and still uses `--json`, since its CLI supports it.
- Updated `plugin_install` tests to mock the Codex text output, added not-installed and wrong-marketplace negative cases, and moved the JSON parse-error assertion to the Claude path (Codex no longer has a JSON parse path).

Validation (isolated worktree off `main`):
- `cargo test -p nemo-relay-cli plugin_install` — 34 passed
- `cargo clippy -p nemo-relay-cli --all-targets` — clean
- `cargo fmt -- --check` — clean
- pre-commit hooks passed on commit

`crates/cli/src/plugin_install/host.rs` — `codex_plugin_registered` and the new `codex_plugin_line_installed` helper.

- Closes #313

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Codex plugin registration detection by reading the standard (non-JSON) plugin list table output instead of relying on JSON parsing.
  * Strengthened matching to require the correct plugin name and marketplace pairing, reducing incorrect registrations.
  * Updated behavior when the plugin list output format can’t be parsed, so errors are surfaced as expected.

* **Tests**
  * Revised plugin installation and coverage tests to mock the text table output and validate installed, not installed, and mismatched marketplace scenarios.

### DIFF
--- a/crates/cli/tests/coverage/plugin_install_tests.rs
+++ b/crates/cli/tests/coverage/plugin_install_tests.rs
@@ -559,21 +559,17 @@ fn host_command_helpers_cover_dry_run_missing_failure_and_reporting() {
     );
 
     let runner = MockRunner::default()
-        .with_executable("codex", "/bin/codex")
-        .with_capture_output("/bin/codex plugin list --json", "not json")
-        .with_capture_output("/bin/codex plugin marketplace list", "");
+        .with_executable("claude", "/bin/claude")
+        .with_capture_output("/bin/claude plugin list --json", "not json");
     assert!(
-        host_registration_report(PluginHost::Codex, &normal, &runner)
+        host_registration_report(PluginHost::ClaudeCode, &normal, &runner)
             .unwrap_err()
             .contains("failed to parse")
     );
 
     let runner = MockRunner::default()
         .with_executable("codex", "/bin/codex")
-        .with_capture_output(
-            "/bin/codex plugin list --json",
-            json!({"installed": []}).to_string(),
-        )
+        .with_capture_output("/bin/codex plugin list", "PLUGIN  STATUS  VERSION  PATH\n")
         .with_capture_output("/bin/codex plugin marketplace list", "MARKETPLACE ROOT\n");
     let error = validate_host_registration(PluginHost::Codex, &normal, &runner).unwrap_err();
     assert!(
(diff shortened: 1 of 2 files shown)


crates/cli/tests/coverage/plugin_install_tests.rs
```diff
@@ -559,21 +559,17 @@ fn host_command_helpers_cover_dry_run_missing_failure_and_reporting() {
...
 
     let runner = MockRunner::default()
-        .with_executable("codex", "/bin/codex")
-        .with_capture_output("/bin/codex plugin list --json", "not json")
-        .with_capture_output("/bin/codex plugin marketplace list", "");
+        .with_executable("claude", "/bin/claude")
+        .with_capture_output("/bin/claude plugin list --json", "not json");
     assert!(
-        host_registration_report(PluginHost::Codex, &normal, &runner)
+        host_registration_report(PluginHost::ClaudeCode, &normal, &runner)
```
willkill07: Why are we switching from Codex to Claude Code here?

willkill07 (changes requested):

crates/cli/tests/coverage/plugin_install_tests.rs
zhongxuanwang-nv: Thanks Will! Since now we no longer parse JSON for codex and the only remaining caller of the function `parse_json_command_output` is Claude Code which still uses --json, the assertion moved to Claude to keep that coverage. But indeed this looks confusing and I could've put it in the Claude Code focused test instead. Just addressed it :) Will push the code after I finish some local testing!

willkill07 (approved):

PR #301, opened 2026-06-24
refactor: codec-owned provider-surface detection via a built-in registry

zhongxuanwang-nv:
#### Overview

Implements RELAY-362, the follow-up requested in review of #291: move provider-surface detection out of the hard-coded `if/else` chains in `codec::resolve` into **codec-owned descriptors** iterated through a small built-in registry. `resolve.rs` is now provider-agnostic (no provider field-name literals), new built-in surfaces become additive, and a codec-owned **provider hint** can disambiguate the one ambiguous request shape (an Anthropic request without a top-level `system`). The change is additive, Rust-core-only, and behavior-preserving: every existing `codec::resolve` test passes unchanged and the adaptive path is untouched.

- [x] I confirm this contribution is my own work, or I have the right to submit it under this project's license.
- [x] I searched existing issues and open pull requests, and this does not duplicate existing work.

#### Details

- **`codec::resolve`** — added a `pub(crate) SurfaceDescriptor` (a hint-aware request detector, a response detector, and decode entry points) and a by-value `static REGISTRY` ordered by request-detection priority (Responses > Anthropic > Chat). The public functions are now provider-agnostic policy only: first registry match for requests, unique match for responses, fail-open `normalize_*`. `ProviderSurface` and every existing public signature are unchanged.
- **New `pub fn detect_request_surface_with_hint(body, Option<&str>)`** — `detect_request_surface` delegates to it with `None` for exact parity. A recognized hint can only *upgrade* the ambiguous `messages`-only shape; registry priority order keeps it from overriding a strong `input`/`instructions`/`system` signal.
- **Codec-owned detection** — each built-in codec module (`openai_chat`, `openai_responses`, `anthropic`) owns its detectors (as closures) and decode entry points in a `SURFACE_DESCRIPTOR` const. Only Anthropic's request detector is hint-aware (it claims a system-less `messages` body when the hint is `"anthropic"`); the OpenAI detectors ignore the hint. Adding a future built-in surface (e.g. Gemini) means registering a descriptor, not editing detection logic.
- **Not wired into adaptive (deferred follow-up)** — the hint is available but not yet called. Wiring it into `resolve_request_surface` / `build_semantic_request_view` would change ACG pass-through behavior, so it is intentionally left to a separate follow-up. The shape-only ACG path and its source-guard test are untouched.
- **Tests** — added hint coverage (parity with a `None` hint, the Anthropic upgrade, "never overrides a strong signal", non-object input); the existing detection/normalization tests are unchanged.

[...]

PR #315, opened 2026-06-25
feat: provider-aware ACG request-surface resolution

zhongxuanwang-nv:
#### Overview

Wire the (previously unwired) provider-hint capability from #301 into the ACG (adaptive cache governor) request-surface resolution. ACG now passes the configured `provider` as a disambiguation hint, so a system-less Anthropic request — `messages` with no top-level `system`, which is shape-identical to OpenAI Chat — resolves to `AnthropicMessages` under provider `"anthropic"` and receives Anthropic cache-hint translation instead of passing through unoptimized. Adaptive-only change; the core hint API and the `provider` config surface are unchanged.

#### Details

- **`request_surfaces::resolve_request_surface_from_request`** now takes a `provider_hint: Option<&str>` and calls core `detect_request_surface_with_hint`. Both consumers — the live `build_semantic_request_view` path and the plugin-facade `resolve_request_surface` path — pass the configured provider, so surface resolution stays consistent across them.
- **`build_semantic_request_view(request, provider)`** threads the configured provider (already in scope in `translate_request`) into resolution.
- **Behavior:** only the ambiguous `messages`-only-under-`"anthropic"` case changes. Shape stays authoritative — a strong `input`/`instructions`/`system` signal is never overridden by the hint, so genuine provider/surface mismatches still pass through unchanged (`unwrap_or(request)`). `"openai"`/`"passthrough"` are unaffected.
- **Learn↔translate key alignment:** the learning key is derived from the normalized `AnnotatedLlmRequest`; the observation path keys off the provider codec's annotation (`event.annotated_request()`). The hint makes translate-time decode use the same (Anthropic) codec, so the learned profile is found at translate time — verified by a dedicated assertion.
- **Tests:** updated the request-surface matrix (system-less + `"anthropic"` now resolves Anthropic; added a strong-signal Responses mismatch case that still errors), added a behavior test and an **end-to-end** test that seeds a learned profile and asserts `cache_control` is actually applied, and rewrote the source guard to the new contract (shape-first; decode by the resolved surface; provider only as a hint — still never provider-first).
- **`detect_request_surface` (no-hint):** this PR removes its last in-tree caller. It is intentionally kept in core as the shape-only public counterpart to `detect_request_surface_with_hint` (both added in #301); no core change here.

Local validation: `cargo test -p nemo-relay-adaptive` (all pass), `cargo clippy -p nemo-relay-adaptive --all-targets` clean, `cargo fmt --check` clean. Core is untouched; full cross-language matrix left to CI. (Note: `runtime::features::tests::registration_context_registers_all_supported_callback_types` is a pre-existing flaky test unrelated to this change.)

`crates/adaptive/src/acg/request_surfaces/mod.rs` (the hint-aware shared resolver) and `crates/adaptive/src/acg_component.rs` (`build_semantic_request_view` threading the provider). Then the end-to-end test `acg_component_translate_request_optimizes_system_less_anthropic_with_provider_hint` in `acg_component_tests.rs` — it proves the feature actually fires (`cache_control` applied) and that the learn/translate keys align.

- Relates to RELAY-362 (deferred follow-up to #301)

## Summary by CodeRabbit

* **New Features**
  * Request detection now uses the selected provider as a hint, improving how similar request formats are interpreted for both request decoding and semantic request views.
* **Bug Fixes**
  * Improved provider-specific translation behavior for Anthropic and OpenAI request styles, including better handling of ambiguous system-less Anthropic requests and fewer misclassifications.
* **Tests**
  * Updated integration and unit tests to validate provider-aware surface resolution and corrected expectations for provider routing and failure cases.

### DIFF
--- a/crates/adaptive/src/acg_component.rs
+++ b/crates/adaptive/src/acg_component.rs
@@ -108,8 +108,11 @@ fn decode_request_for_surface(
     }
 }
 
-fn build_semantic_request_view(request: &LlmRequest) -> Result<SemanticRequestView> {
-    let request_surface = resolve_request_surface_from_request(request)
+fn build_semantic_request_view(
+    request: &LlmRequest,
+    provider: &str,
+) -> Result<SemanticRequestView> {
+    let request_surface = resolve_request_surface_from_request(request, Some(provider))
         .map_err(|error| AdaptiveError::Internal(error.to_string()))?;
     let annotated_request = decode_request_for_surface(request_surface, request)?;
 
@@ -387,7 +390,7 @@ fn translate_request(
     hot_cache: &Arc<RwLock<HotCache>>,
 ) -> Option<LlmRequest> {
     // Response codecs stay on the observability path after provider execution.
-    let semantic_request_view = match build_semantic_request_view(request) {
+    let semantic_request_view = match build_semantic_request_view(request, provider) {
         Ok(view) => view,
         Err(error) => {
             acg_debug::emit(
--- a/crates/adaptive/src/acg/request_surfaces/mod.rs
+++ b/crates/adaptive/src/acg/request_surfaces/mod.rs
@@ -16,7 +16,7 @@ pub(crate) mod openai_responses;
 use std::collections::HashSet;
 
 use nemo_relay::api::llm::LlmRequest;
-use nemo_relay::codec::resolve::{ProviderSurface, detect_request_surface};
+use nemo_relay::codec::resolve::{ProviderSurface, detect_request_surface_with_hint};
 use serde_json::Value;
 
 use crate::acg::prompt_ir::PromptIR;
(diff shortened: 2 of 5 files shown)


crates/adaptive/src/acg_component.rs
```diff
@@ -387,7 +390,7 @@ fn translate_request(
     hot_cache: &Arc<RwLock<HotCache>>,
 ) -> Option<LlmRequest> {
     // Response codecs stay on the observability path after provider execution.
-    let semantic_request_view = match build_semantic_request_view(request) {
+    let semantic_request_view = match build_semantic_request_view(request, provider) {
```
willkill07, line 393: I think this is a regression? Ideally we want to _drop_ the explicit provider mention within ACG. This not only brings it back, but overrides automatic detection.

willkill07 (changes requested):

zhongxuanwang-nv:
> I think this is a regression? Ideally we want to _drop_ the explicit provider mention within ACG. This not only brings it back, but overrides automatic detection.

Thanks Will! I agree, I opened this PR per the follow-up deferred from [PR-291](https://github.com/NVIDIA/NeMo-Relay/pull/291) (passing the configured `provider` so a system-less (a message without top-level `system` param) Ant request would resolve to Ant)

But it's true that we already have automatic shape detection, and this PR would make configured provider override that, and we also want the resolution to be driven by shape and be provider-agnostic.

If we in the future would like to drive ACG by provider codecs more broadly, or we need to optimize system-less Ant traffic, we could potentially bring this back!

PR #318, opened 2026-06-26
refactor: share agent and provider extraction strategies

yczhang-nv:
#### Overview

Establish explicit, shared extraction-strategy boundaries before code is moved onto them: provider schema extraction (codec-owned), agent payload extraction (harness-owned), and exporter projection. This is a refactor — runtime ownership, event shapes, and binding-facing behavior are unchanged. The only public surface added is the additive `normalize_request_with_hint`.

This branch has been merged with current `main` after PR #333 removed Cursor support; the agent-side extraction split now covers the supported coding-agent integrations: Claude Code, Codex, and Hermes.

#### Details

**What changed**

Provider schema extraction (`crates/core`):
- Renamed the internal codec registry surface to provider-extraction names: `SurfaceDescriptor` → `ProviderSurfaceDescriptor`, `REGISTRY` → `BUILTIN_PROVIDER_SURFACES`, per-codec `SURFACE_DESCRIPTOR` → `PROVIDER_SURFACE`, and the detector fn aliases to `RequestSurfaceDetector` / `ResponseSurfaceDetector`. All remain `pub(crate)`.
- Documented `LlmResponseCodec::decode_response` as the provider response-extraction interface (model name, finish reason, tool calls, usage, cost, provider-specific fields, replayable response data).
- Added hint-aware request normalization: `normalize_request_with_hint(request, hint)` (the existing `normalize_request` now delegates with `None`). `EventNormalizationExt` threads the managed provider-route event name (e.g. `anthropic.messages`) as the codec hint, so an Anthropic Messages body without a top-level `system` no longer misroutes to the OpenAI Chat codec. The Anthropic detector accepts both `"anthropic"` and `"anthropic.messages"` hints.

Agent payload extraction (`crates/cli`):
- Added shared low-level JSON-path primitives in `crates/cli/src/json_path.rs`, consumed by both adapter and alignment code (existing names preserved via re-export).
- Split agent payload extraction into per-host `AgentPayloadExtractor` implementations for Claude Code, Codex, and Hermes; event construction (`common_*_event`, `classify`) stays shared and takes `&dyn AgentPayloadExtractor`.
- Split provider request extraction into per-route `ProviderRequestExtractor` implementations (OpenAI Responses / Chat Completions / Models, Anthropic Messages / Count-Tokens), selected by gateway route.
- Compatibility fallbacks (synthetic session IDs, synthetic tool-call IDs, `unknown_tool`, generic subagent ID, `Value::Null`) stay at the adapter/session boundary; extractors return `None` for absent fields. Path/PII fields (cwd, transcript_path, project_dir, user email) are not promoted into shared metadata.

Docs:
- New "Extraction Strategy Boundaries" section in `docs/about-nemo-relay/concepts/codecs.mdx` documenting the provider / agent / exporter split and fallback semantics for missing, sparse, lossy, and summary-only payloads.

**Why**

Multiple components pull structured information from heterogeneous agent and provider payloads. Defining the strategy interfaces (and separating agent-payload extraction from provider-schema extraction) up front lets later work migrate the duplicated, best-effort extraction chains onto shared seams without widening the public API.

**How it was tested**

Latest validation after merging current `main`:
- `cargo fmt --all`
- `cargo test -p nemo-relay-cli adapters::tests`
- `PYO3_PYTHON=/Users/yuchenz/Desktop/Work/Project/NeMo-Flow/.venv/bin/python cargo clippy --workspace --all-targets -- -D warnings`
- `just test-rust`
- `git diff --check`

Additional validation performed earlier on this branch:
- `uv run pre-commit run --all-files` — SPDX/fmt/clippy/audit + public-docstring gate.

Validation scope for review:
- `validate-change` was used to select the required matrix. Because this PR touches `crates/core` event normalization (`EventNormalizationExt` and `normalize_request_with_hint`), pre-merge verification should include the broader runtime/language matrix, not Rust-only coverage: `cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings`, `just test-rust`, `just test-python`, `just test-node`, `just test-go`, and `just test-wasm`.
- Docs changes should also be covered by `just docs` before merge. Binding behavior is expected to remain compatible because the core API change is additive, event shapes are unchanged, and renamed codec registry symbols remain `pub(crate)`; the full binding matrix is still the required check set for this `crates/core` change.

**Breaking changes**

None from this PR. `normalize_request_with_hint` is additive; `normalize_request` keeps its signature and behavior (delegates with no hint).

`crates/cli/src/adapters/mod.rs` (the `AgentPayloadExtractor` implementations and the per-agent path tables), then `crates/cli/src/json_path.rs` and the adapter coverage tests, then `crates/cli/src/alignment/mod.rs` (`ProviderRequestExtractor` split), and finally `crates/core/src/codec/resolve.rs` (descriptor rename + hint threading) with `crates/core/src/api/event.rs` for the hint wiring.


crates/cli/src/adapters/mod.rs
```diff
@@ -31,13 +31,182 @@ pub(super) struct ClassificationRules<'a> {
...
+            .or_else(|| string_at(payload, &["event"]))
+            .or_else(|| string_at(payload, &["type"]))
+            .or_else(|| string_at(payload, &["name"]))
+            .or_else(|| string_at(payload, &["extra", "hook_event_name"]))
+            .or_else(|| string_at(payload, &["extra", "event_name"]))
+            .or_else(|| string_at(payload, &["extra", "eventName"]))
+            .or_else(|| string_at(payload, &["extra", "event"]))
+            .or_else(|| string_at(payload, &["extra", "type"]))
+            .or_else(|| string_at(payload, &["extra", "name"]))
+    }
```
willkill07: I was hoping there would be some sort of abstraction for each supported integration such that this type of extraction would be a separate `impl` for each Trait.

Example:

```rust
trait AgentInfo {
  fn event_name() -> &str;
  fn metadata() -> Value;
  fn session_id() -> Option<String>;
  fn subagent_id() -> Option<String>;
  fn llm_hint_event() -> LlmHintEvent;
  fn tool_event() -> ToolEvent;
}
```

Then based off of the payload, have an appropriate `impl for AgentInfo`

This type of chaining is what we are trying to avoid. Moving it doesn't solve the underlying problem. The solution should be rooted in:
**if we have to add another type of Agent Harness and/or LLM Provider, how can we easily extract information?**

crates/cli/src/adapters/mod.rs
```diff
@@ -31,13 +31,182 @@ pub(super) struct ClassificationRules<'a> {
...
+            result: tool_result(payload, &normalized_event),
+            status: tool_status(payload, &normalized_event),
+        }
+    }
+}
+
+// TODO(extraction): These free-function shims preserve existing adapter call
+// sites while host-specific extractors land. Move callers onto
+// `AgentPayloadExtractor` implementations and remove the shims after that
+// migration completes.
```
willkill07: We should not have any TODOs. If this is a refactor PR, then removal/replacement should occur with it.

willkill07 (changes requested):
We need to lean more into Traits which model expected behavior/APIs. And have implementations of those Traits be dependent on either (1) the agent harness supported (2) the LLM provider being used and/or (3) a combination of both.

We should avoid these best-effort chains for information extraction.

yczhang-nv:
@coderabbitai review

yczhang-nv:
> We need to lean more into Traits which model expected behavior/APIs. And have implementations of those Traits be dependent on either (1) the agent harness supported (2) the LLM provider being used and/or (3) a combination of both.
> 
> We should avoid these best-effort chains for information extraction.

Thanks @willkill07 for the elaboration! PR updated with the refactor direction you pointed out.  The shared best-effort chains are gone. Agent payloads now go through an `AgentPayloadExtractor` trait with one implementation per harness. And the provider side has a `ProviderRequestExtractor` trait with per-route impls (OpenAI Responses/Chat, Anthropic Messages/CountTokens) selected by `GatewayRouteKind`. Each impl owns its own schema paths instead of a single shared chain.

crates/cli/src/adapters/mod.rs
yczhang-nv: Got it. The interface is now defined in `AgentPayloadExtractor`. As a result, adding a new harness or provider is now "add one impl + its path set."

crates/cli/src/adapters/mod.rs
yczhang-nv: The agent integrations are also included in this PR to avoid TODOs.

crates/cli/src/adapters/mod.rs
```diff
@@ -30,68 +35,493 @@ pub(super) struct ClassificationRules<'a> {
...
+    &["session_id"],
+    &["sessionId"],
+    &["session", "id"],
+    &["conversation_id"],
+    &["conversationId"],
+    &["parent_session_id"],
+    &["task_id"],
+    &["extra", "session_id"],
+    &["extra", "task_id"],
+];
```
willkill07: is this really true that all three are the same? is there not a canonical representation based on harness? or is it only based on LLM provider?

crates/cli/src/adapters/mod.rs
```diff
@@ -30,68 +35,493 @@ pub(super) struct ClassificationRules<'a> {
...
+    &["event"],
+    &["type"],
+    &["name"],
+    &["extra", "hook_event_name"],
+    &["extra", "event_name"],
+    &["extra", "eventName"],
+    &["extra", "event"],
+    &["extra", "type"],
+    &["extra", "name"],
+];
```
willkill07: Same here....

crates/cli/src/adapters/mod.rs
```diff
@@ -30,68 +35,493 @@ pub(super) struct ClassificationRules<'a> {
...
+    &["tool_call_id"],
+    &["toolCallId"],
+    &["tool_use_id"],
+    &["call_id"],
+    &["extra", "tool_call_id"],
+    &["extra", "call_id"],
+    &["tool", "id"],
+    &["tool_input", "id"],
+    &["id"],
+];
```
willkill07, line 359: These also seem the same

willkill07 (commented):
I like the shape a lot more. Just some nits on how to meaningfully extract information based on Agent. If things are provider-specific, then we should be offloading that extraction based on the provider rather than generalizing on the harness.

crates/cli/src/adapters/mod.rs
yczhang-nv: Collapsed into one canonical `SESSION_ID_PATHS`

crates/cli/src/adapters/mod.rs
yczhang-nv: Collapsed into one canonical `EVENT_NAME_PATHS`

crates/cli/src/adapters/mod.rs
yczhang-nv: Collapsed into one canonical `TOOL_CALL_ID_PATHS`

yczhang-nv:
> If things are provider-specific, then we should be offloading that extraction based on the provider rather than generalizing on the harness.

@willkill07 Those paths are harness-specific: `session-id` and `event-name` are canonical (same across harnesses, only the header policy differs); `subagent-id` and `tool-call/arg` precedence genuinely vary by harness. 

The latest commit extracted the canonical paths, and ensured each harness overrides only the hooks where it genuinely deviates, making the code easier to understand and maintain.

willkill07 (approved):

lvojtku (approved):

PR #325, opened 2026-06-29
feat: configure dynamic plugins through CLI

willkill07:
#### Overview

Add a static, manifest-declared JSON Schema boundary so dynamic plugins can be configured through `nemo-relay plugins edit` without loading native code or starting workers. The editor now presents one top-level row per built-in or dynamic plugin and remains usable in small terminals.

#### Details

- Extend manifest v1 with an optional `[config_schema]` section and enforce its bidirectional relationship with the `config_schema` capability.
- Load and validate JSON Schema Draft 7 and Draft 2020-12 documents using local same-document references only, map supported shapes to native controls, and retain raw JSON fallbacks with full schema validation.
- Treat schema-declared `writeOnly` strings as secrets by masking menus and previews, using hidden input, and preserving existing values unless explicitly replaced or cleared.
- Rework `plugins edit` into built-in and dynamic plugin submenus with per-field and whole-plugin reset behavior, plus terminal-height-aware scrolling, truncation, and Page Up/Down and Home/End navigation.
- Preserve host policy, dynamic manifest references, unknown TOML fields and sections, TOML-native values, and absent-versus-empty dynamic configuration. Reuse the lossless document writer in pricing commands.
- Add manifest, schema, persistence, editor, viewport, lifecycle, and pricing regression tests; update CLI help, documentation, the Rust native plugin example, and the plugin-building skill.
- Keep the native ABI, worker protocol, SDK callbacks, runtime validation authority, and manifest version unchanged. Existing valid manifests remain compatible; the `config_schema` capability and section must now be declared together.

Validation performed:

- `just test-rust`
- `just test-python` (382 passed)
- `just test-go`
- `just test-node` (243 passed)
- `just test-wasm` (Rust checks and 80 JavaScript tests passed)
- `just docs`
- `just docs-linkcheck`
- `cargo deny check`
- Commit-time pre-commit hooks passed. The all-files run passed every hook except `ty`, which traversed a local untracked generated `python/plugin/` tree that is not part of this branch; the same type check excluding that directory exited successfully with two pre-existing warnings.

Start with `crates/cli/src/plugins/schema.rs` for the schema boundary and control mapping, then `crates/cli/src/plugins/dynamic_editor.rs` for editing and secret semantics, and `crates/cli/src/plugins/config_io.rs` for lossless TOML patching.

- Closes: RELAY-312

## Summary by CodeRabbit

* **New Features**
  * Added schema-backed interactive editing for dynamic plugins, including secret-aware input/redaction, schema-driven validation, and merged TOML preview/save.
  * Dynamic plugin manifests can declare local `config_schema`; CLI lifecycle commands now validate configuration against it.
  * Improved plugin editor UI with terminal-size-aware paging plus Page Up/Down and Home/End navigation, and expanded edit flows to include dynamic entries.
* **Bug Fixes**
  * Plugin configuration updates now preserve existing `plugins.toml` host/other sections (including during pricing updates).
* **Documentation**
  * Updated Quick Start and plugin editor docs with detailed dynamic schema behavior and editor shortcuts; added schema authoring guidance and examples.
* **Tests**
  * Expanded lifecycle and editor coverage for config-schema handling, validation failures, redaction behavior, and menu/paging navigation.


crates/cli/src/plugins/dynamic_editor.rs
```diff
@@ -0,0 +1,839 @@
...
+            remove_value_at_path(config, path);
+        }
+        self.touched = true;
+    }
+
+    fn reset_field(&mut self, path: &[String], field: &DynamicConfigField) {
+        match &field.default {
+            Some(default) => self.set_field(path, default.clone()),
+            None => self.remove_field(path),
+        }
```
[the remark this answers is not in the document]
willkill07: Fixed the absent-config case: clearing or resetting an unset optional field now preserves an absent `config`, with a regression test. For an existing config, resetting a no-default field intentionally removes it per the documented reset contract; required-field validation remains authoritative at save.

crates/cli/src/plugins/schema.rs
```diff
@@ -0,0 +1,1416 @@
...
+        for pattern in &self.secret_patterns {
+            pattern.redact_for_edit(&mut redacted, 0, &mut secrets, &occupied, &mut next_token);
+        }
+        (redacted, secrets)
+    }
+
+    /// Restores only the exact per-value tokens emitted by [`Self::redact_for_edit`].
+    pub(super) fn restore_edit_secrets(&self, edited: &Value, secrets: &SecretEditValues) -> Value {
+        restore_secret_tokens(edited, secrets)
+    }
```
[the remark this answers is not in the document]
willkill07: Fixed: each token now carries its secret schema pattern; restoration rejects tokens copied outside matching secret locations and duplicate occurrences. Array reordering remains supported where the schema uses wildcard items. Added regressions.

crates/cli/src/plugins/schema.rs
```diff
@@ -0,0 +1,1416 @@
...
+        schema: &Value,
+        pointer: &str,
+    ) -> Result<(), CliError> {
+        let Some(object) = schema.as_object() else {
+            return Ok(());
+        };
+        for key in ["$ref", "$dynamicRef"] {
+            if let Some(reference) = object.get(key).and_then(Value::as_str)
+                && !reference.starts_with('#')
+            {
```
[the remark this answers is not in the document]
willkill07: Fixed by rejecting `$dynamicRef` and `$dynamicAnchor` before editor synthesis; only standard same-document `$ref` is supported and documented. Added tests.

crates/core/src/plugin/dynamic/manifest.rs
```diff
@@ -459,6 +499,53 @@ fn reject_duplicate_capabilities(capabilities: &[DynamicPluginCapability]) -> Re
...
+            return Err(PluginError::InvalidConfig(
+                "[config_schema] requires capabilities.items containing config_schema".into(),
+            ));
+        }
+        (false, None) => return Ok(()),
+        (true, Some(config_schema)) => {
+            let path = required_trimmed_string(Some(&config_schema.path), "config_schema.path")?;
+            if has_uri_scheme(path.trim()) {
+                return Err(PluginError::InvalidConfig(
+                    "config_schema.path must be a local filesystem path, not a URI".into(),
```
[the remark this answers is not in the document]
willkill07: Fixed: config schema paths now reject UNC, verbatim UNC, and `//` network-share prefixes before I/O. Added manifest regression coverage.

crates/cli/src/plugins/schema.rs
```diff
@@ -0,0 +1,1543 @@
...
+}
+
+#[derive(Debug, Clone)]
+enum SecretInstanceSegment {
+    Property(String),
+    Index(usize),
+}
+
+fn pattern_matches(pattern: &str, property: &str) -> bool {
+    regex::Regex::new(pattern)
```
bbednarski9: Doesn't this fail silently if the regex schema load raises and exception? Maybe we can validate them once at schema load an improve perf as well

crates/cli/src/plugins/config_io.rs
```diff
@@ -26,6 +26,229 @@ pub(crate) enum TargetScope {
...
+                    (Some(raw), Some(original)) => patch_json_value(raw, original, updated)?,
+                    _ => {
+                        raw.insert(key.clone(), json_to_toml(updated.clone())?);
+                    }
+                }
+            }
+            Ok(())
+        }
+        (toml::Value::Array(raw), Value::Array(original), Value::Array(updated))
+            if raw.len() == original.len() && original.len() == updated.len() =>
```
bbednarski9: Interesting AI review comment:

Array length changes drop TOML-native values (config_io.rs:229-241). patch_json_value only preserves native TOML values for objects (key-wise) and same-length arrays. When an array's length changes, it falls through to whole-array replacement via json_to_toml(updated), so a TOML datetime nested inside a resized array degrades to a string. This is a narrow edge (datetimes inside dynamic-plugin config arrays are unusual) and is partially implied by the "unchanged values are preserved" doc, but it's untested and silent. Recommendation: add a test asserting the behavior and either document the limitation explicitly or patch element-wise up to min(len).

crates/core/src/plugin/dynamic/manifest.rs
```diff
@@ -334,6 +346,34 @@ impl DynamicPluginManifest {
         Ok(())
     }
 
+    /// Resolves the declared configuration schema path without accessing it.
+    ///
+    /// Relative schema paths are resolved from the parent directory of the
+    /// canonical `relay-plugin.toml` path. Absolute local paths are returned
+    /// unchanged.
+    pub fn resolve_config_schema_path(
```
bbednarski9, line 354: As this is a pub, are we circumventing the URI/UNC rejection here?

crates/core/src/plugin/dynamic/manifest.rs
```diff
@@ -459,6 +499,57 @@ fn reject_duplicate_capabilities(capabilities: &[DynamicPluginCapability]) -> Re
...
+                return Err(PluginError::InvalidConfig(
+                    "config_schema.path must be a local filesystem path, not a URI".into(),
+                ));
+            }
+        }
+    }
+    Ok(())
+}
+
+fn is_unc_path(value: &str) -> bool {
```
bbednarski9, line 537: Can we add a comment or warning that this also occasionally would block a valid POSIX path? I dont think we need to open it up, but maybe worth a note

crates/cli/src/plugins.rs
```diff
@@ -346,6 +405,16 @@ fn prompt_menu(
             Key::ArrowDown | Key::Char('j') => {
                 selected = (selected + 1) % items.len();
             }
+            Key::PageUp => {
```
bbednarski9: Now that we're using this PageUp/PageDown logic in a few places, maybe we can put these into a helper with centralized testing? It's nit and could be a follow-up PR

bbednarski9 (changes requested):
Super cool functionality. Excited for this. But a few comments on generalizability and testing robustness

willkill07:
@coderabbitai review

willkill07:
@coderabbitai review

willkill07:
@coderabbitai review

Salonijain27 (approved):

bbednarski9:
reviewing now

bbednarski9:
Explicitly missing --config is silently accepted.
nemo-relay run --config /missing/config.toml --dry-run exits 0 and uses defaults instead of reporting a missing file.

bbednarski9:
Unreadable config errors omit the source path.
Permission failures report only Permission denied, without identifying which config file failed.

Tested this with the following and confirmed:
```
cd /Users/bbednarski/Projects/code-review/NeMo-Relay-pr325

BIN="$PWD/target/debug/nemo-relay"
TMP="$(mktemp -d)"
trap 'chmod 644 "$TMP/config.toml" 2>/dev/null; rm -rf "$TMP"' EXIT

cat > "$TMP/config.toml" <<'EOF'
[upstream]
openai_base_url = "http://example.invalid"
EOF

chmod 000 "$TMP/config.toml"

HOME="$TMP/home" \
XDG_CONFIG_HOME="$TMP/xdg" \
"$BIN" run \
  --agent codex \
  --config "$TMP/config.toml" \
  --dry-run

echo "exit=$?"
```

Also worth checking if this is the same behavior for plugins.toml

bbednarski9:
Percent-encoded JSON Schema $ref paths fail.
References such as #/$defs/a%20b are rejected instead of URI-decoded.

crates/cli/src/plugins/schema.rs
```diff
@@ -0,0 +1,1308 @@
...
+    pub(super) fn fields(&self) -> &[DynamicConfigField] {
+        &self.editor.fields
+    }
+
+    /// Validates one object configuration and reports the first failing instance pointer.
+    pub(super) fn validate(&self, config: &Value) -> Result<(), CliError> {
+        self.validator.validate(config).map_err(|error| {
+            let pointer = error.instance_path().to_string();
+            CliError::Config(format!(
+                "dynamic plugin '{}' configuration at JSON pointer '{}' is invalid: {error}",
```
bbednarski9: we should redact secrets here so we dont leak to the console... I think that got missed

crates/cli/src/plugins/schema.rs
```diff
@@ -0,0 +1,1308 @@
...
+        Ok(resolved) => resolved,
+        Err(error) => {
+            return Err(format!(
+                "secret schema reference could not be resolved: {error}"
+            ));
+        }
+    };
+    if schema_type(resolved) == Some("string") && annotation_bool(schema, resolved, "writeOnly") {
+        output.push(SecretPattern(instance_path.to_vec()));
+        return Ok(());
```
bbednarski9, line 1442: Codex warning: The scanner only records writeOnly when the resolved node directly has type string. Draft 2020-12 `$ref` siblings are discarded by resolution, and annotations split across applicators or nullable string types are missed, causing plaintext preview and editable initial values. Either implement effective annotation collection or reject unsupported writeOnly shapes.

bbednarski9:
This seems worth further investigation:

The remaining P1 issue is the exact Draft 2020-12 structure previously reported:
```
{
  "$defs": { "base": { "type": "object" } },
  "properties": {
    "auth": {
      "$ref": "#/$defs/base",
      "properties": {
        "password": { "type": "string", "writeOnly": true }
      }
    }
  }
}
```
Resolution replaces auth with base, so the adjacent properties subtree is never visited during secret discovery. Additionally, malformed non-string values at known secret paths remain visible because `redact()` (line 1149)]= only masks when value.is_string().

bbednarski9 (approved):

bbednarski9 (approved):

bbednarski9 (approved):

PR #327, opened 2026-06-29
feat(plugin)!: support pending marks from LLM intercepts

bbednarski9:
#### Overview

Finalize one canonical LLM request-intercept outcome across the Rust runtime, built-in and adaptive plugins, native ABI v1, `grpc-v1` workers, public C FFI, Go, Python, Node.js, and WebAssembly.

Request intercepts can rewrite the provider request, carry an optional normalized annotation, and schedule ordered marks for the managed LLM lifecycle:

```json
{
  "request": {"headers": {}, "content": {}},
  "annotated_request": null,
  "pending_marks": []
}
```

`request` is required. `annotated_request` defaults to `null`, and `pending_marks` defaults to an empty list. Each pending mark contains only its name, optional category and category profile, data, and metadata; Relay continues to own event UUIDs, parent UUIDs, and timestamps.

The finalized contract also defines one provider-body source of truth. Without a request codec, `outcome.request.content` is authoritative. With a codec, `outcome.annotated_request` is required and authoritative, `outcome.request.content` is read-only context, and `outcome.request.headers` remains writable.

#### Why

Request intercepts run before Relay creates the managed LLM handle. A mark emitted directly from an intercept therefore cannot reliably attach to that future LLM scope. Returning pending mark specifications lets the lifecycle owner emit them at the correct boundary without leaking control data into provider requests, annotations, codecs, sanitizers, or execution intercepts.

Codec-aware interception also previously allowed two conflicting provider-body representations: an intercept could change both the raw request content and its normalized annotation, while Relay later encoded only the annotation. Making authority explicit prevents raw content edits from being silently discarded.

#### Details

- Make `LlmRequestInterceptOutcome` the only Rust callback result and keep one `register_llm_request_intercept` registration family for global, scope-local, plugin-context, and adaptive paths.
- Propagate each accepted request and annotation to the next intercept while appending pending marks in effective middleware order.
- Without a request codec, use `outcome.request.content` as the provider body.
- With a request codec, require `outcome.annotated_request`, encode the provider body from it, and allow header changes only through `outcome.request.headers`.
- Reject raw `request.content` mutations or missing annotations at the offending codec-path intercept, before later middleware, LLM lifecycle creation, mark emission, or provider invocation.
- Preserve marks from an intercept that breaks the chain; discard all accumulated marks if any intercept fails.
- Return the complete outcome from standalone request-intercept helpers. These helpers expose pending marks but do not emit them because they do not own an LLM lifecycle.
- After successful interception, create the LLM handle and capture one subscriber snapshot before emitting lifecycle events.
- Emit LLM start at `T`, every pending mark at `T + 1µs` in returned order with the LLM UUID as parent, and LLM end at or after `T + 1µs`.
- Apply the same behavior to streaming and non-streaming managed execution, including provider errors and stream finalization.
- Keep pending marks separate from provider-visible requests and annotations.

#### Boundary contracts

- **Native ABI v1:** return one host-owned outcome JSON string. Remove the private annotation-envelope transport and append required outcome-contract version fields to both host and plugin descriptor tables so stale binaries fail before callback invocation.
- **`grpc-v1`:** return one `JsonEnvelope` using schema `nemo.relay.LlmRequestInterceptOutcome@1`.
- **Public C FFI:** return one owned `char **out_outcome_json` and add `nemo_relay_llm_request_intercept_outcome_json_new`.
- **Go:** return `(LLMRequestInterceptOutcome, error)` and expose request, outcome, and pending-mark DTOs.
- **Python:** return `LLMRequestInterceptOutcome` and export `PendingMarkSpec`.
- **Node.js and WebAssembly:** return `{ request, annotated?, pendingMarks? }`. Binding-owned pending-mark DTOs use `categoryProfile`; canonical event and outcome JSON retains `category_profile`.
- **Rust native and worker SDKs:** expose only the canonical callback and registration method.

#### Breaking changes

This intentionally finalizes unpublished contracts in place:

- Rust and Python tuple results are removed.
- C and Go split outputs are removed.
- Mark-specific parallel registration variants are removed.
- The native annotation metadata envelope and fallback parser are removed.
- Native ABI host and plugin tables require the finalized outcome-contract field.
- The `grpc-v1` request-intercept result is replaced by the canonical outcome envelope.
- Codec-path intercepts must return an annotation and may no longer mutate raw `request.content`; malformed outcomes fail before lifecycle creation.
- Node.js and WebAssembly pending-mark objects use `categoryProfile` instead of the Rust/wire name `category_profile`.

All development native plugins and workers must rebuild against this version.

1. `crates/types/src/api/event.rs` and `crates/types/src/api/llm.rs` for the canonical data contract.
2. `crates/core/src/api/runtime/state.rs`, `crates/core/src/api/shared.rs`, `crates/core/src/api/llm.rs`, and `crates/core/src/stream.rs` for chaining, codec authority, validation, and lifecycle behavior.
3. `crates/plugin/src/lib.rs`, `crates/core/src/plugin/dynamic/native.rs`, and `crates/core/src/plugin/dynamic/worker.rs` for native and worker boundaries.
4. `crates/ffi`, `go/nemo_relay`, `crates/python`, `crates/node`, and `crates/wasm` for binding contracts and DTO conversion.
5. `crates/core/tests/integration/middleware_tests.rs`, `crates/core/tests/integration/pipeline_tests.rs`, `crates/plugin/tests/typed_callbacks.rs`, and the binding tests for lifecycle, codec-authority, and boundary coverage.

The full contract, request-authority diagram, and migration notes are tracked in [companion documentation PR #341](https://github.com/NVIDIA/NeMo-Relay/pull/341), which should merge immediately after this PR.

#### Testing

- `cargo test --workspace --all-targets`
- Python codec and worker SDK coverage passes, including malformed codec-path outcomes and canonical worker envelopes.
- Node.js LLM suite: **38 passed**, including `categoryProfile` input/output conversion and codec-authority rejection.
- Go: all `go/nemo_relay/...` packages passed, including codec-authority coverage; `go vet ./...` passes.
- Native SDK: **52 passed**.
- Worker SDK: **9 passed**; worker protocol tests: **6 passed**.
- C FFI: unit and integration suites passed, including owned outcome allocation and malformed/null input coverage.
- WebAssembly native Rust tests: **13 passed**, including camelCase pending-mark DTO round trips and rejection of the wire-only `category_profile` spelling.
- Repository formatting, strict Clippy, Ruff, Prettier, type, lockfile, FFI-header, and applicable pre-commit checks pass.

`wasm-pack` and the `wasm32-unknown-unknown` Rust target were not available for the package-level Wasm suite. Environment-dependent socket and external-network tests were not used to validate these binding changes.

#### Related Issues

- Relates to #296

## Summary by CodeRabbit

* **New Features**
  * LLM request intercepts can now return a unified outcome that includes the rewritten request, optional annotated request, and pending marks.
  * Pending marks are now emitted alongside LLM lifecycle events and supported across SDKs and plugins.

* **Bug Fixes**
  * Improved consistency of LLM event timing and parent/child relationships.
  * Added stricter validation so intercepts that modify raw request content or omit required annotations are rejected when needed.


crates/types/src/api/llm.rs
```diff
@@ -20,10 +22,48 @@ bitflags! {
...
 pub struct LlmRequest {
     /// Provider-specific request headers.
     pub headers: serde_json::Map<String, Json>,
     /// Provider-specific request body.
     pub content: Json,
 }
+
+/// Result of an LLM request intercept that can schedule lifecycle marks.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct LlmRequestInterceptOutcome {
```
willkill07, line 35: I believe if we add:

```rust
impl From<(LlmRequest, Option<AnnotatedLlmRequest>)> for LlmRequestInterceptOutcome {
   fn from(tuple: (LlmRequest, Option<AnnotatedLlmRequest>)) -> Self {
        LlmRequestInterceptOutcome {
            request: tuple.0,
            annotated_request: tuple.1,
            pending_marks: vec![],
        }
    }
}

impl From<(LlmRequest, AnnotatedLlmRequest)> for LlmRequestInterceptOutcome {
   fn from(tuple: (LlmRequest, AnnotatedLlmRequest)) -> Self {
        (tuple.0, Some(tuple.1)).into()
    }
}

impl From<LlmRequest> for LlmRequestInterceptOutcome {
   fn from(request: LlmRequest) -> Self {
        (request, None).into()
    }
}
```

Then we (almost) get full backwards compatibility and new convenience converters. At the callsite:

```rust
fn callable() -> LlmRequestInterceptOutcome {
  let request: LlmRequest = /* some call */;
  ...
  request.into()
}
```

crates/core/src/api/runtime/callbacks.rs
```diff
@@ -163,17 +163,17 @@ pub type LlmConditionalFn = Arc<dyn Fn(&LlmRequest) -> Result<Option<String>> +
...
 /// # Errors
 /// The callback can return any [`FlowError`](crate::error::FlowError) to abort
 /// the request-intercept chain.
 pub type LlmRequestInterceptFn = Arc<
-    dyn Fn(
-            &str,
-            LlmRequest,
-            Option<AnnotatedLlmRequest>,
-        ) -> Result<(LlmRequest, Option<AnnotatedLlmRequest>)>
+    dyn Fn(&str, LlmRequest, Option<AnnotatedLlmRequest>) -> Result<LlmRequestInterceptOutcome>
```
willkill07, line 176: I was always on the fence about returning a tuple. Concretizing this to a type is a huge win

crates/core/src/api/registry.rs
```diff
@@ -547,8 +547,7 @@ global_guardrail_registry_api!(
     LlmConditionalFn
 );
 global_intercept_registry_api!(
-    /// Register a global LLM request intercept.
-    /// Request intercepts can rewrite or annotate the outgoing LLM request.
+    /// Register a global LLM request intercept that can schedule lifecycle marks.
```
willkill07: We lose the behavioral description on rewrite/annotation

crates/core/src/api/registry.rs
```diff
@@ -653,9 +652,7 @@ scope_guardrail_registry_api!(
     LlmConditionalFn
 );
 scope_intercept_registry_api!(
-    /// Register a scope-local LLM request intercept.
-    /// Request intercepts can rewrite or annotate LLM requests inside the
-    /// owning scope.
+    /// Register a scope-local LLM request intercept that can schedule lifecycle marks.
```
willkill07: We lose the behavioral description on rewrite/annotation

crates/plugin/src/lib.rs
```diff
@@ -28,6 +30,8 @@ use serde_json::Map;
 
 /// Native plugin ABI version supported by this crate.
 pub const NEMO_RELAY_NATIVE_ABI_VERSION: u32 = 1;
+/// Final canonical LLM request-intercept outcome contract version.
+pub const NEMO_RELAY_NATIVE_LLM_INTERCEPT_OUTCOME_CONTRACT_VERSION: u32 = 1;
```
willkill07: Is this really necessary?

crates/core/src/bindings.rs
willkill07, line 1: We don't have any shared language binding types in core. should be duplicated across WebAssembly and Node crates.

crates/core/src/lib.rs
```diff
@@ -54,6 +54,8 @@
 //! All middleware is priority-ordered (ascending) and registered by name for
 //! easy addition and removal at runtime.
 pub mod api;
+#[doc(hidden)]
+pub mod bindings;
```
willkill07:
```suggestion
```

crates/node/src/api/mod.rs
```diff
@@ -2259,6 +2259,9 @@ pub fn register_llm_request_intercept(
     name: String,
     priority: i32,
     break_chain: bool,
+    #[napi(
```
willkill07, line 2262: thank you for adding argument types to this!

crates/node/src/callable.rs
```diff
@@ -19,11 +19,13 @@ use nemo_relay::api::runtime::{
...
     ToolExecutionNextFn, ToolInterceptFn, ToolSanitizeFn,
 };
+use serde::Deserialize;
 use serde_json::Value as Json;
 use tokio_stream::StreamExt;
 
 use nemo_relay::api::event::Event;
-use nemo_relay::api::llm::LlmRequest;
+use nemo_relay::api::llm::{LlmRequest, LlmRequestInterceptOutcome};
+pub(crate) use nemo_relay::bindings::js::{JsPendingMarkSpec, js_pending_marks};
```
willkill07: should be moved into this crate

willkill07 (approved):
Approving with nits on:
- structuring of crate-dependent types.
- unnecessary ABI checks.
- suggested `.into()` helpers.

crates/types/src/api/llm.rs
bbednarski9: Great suggestion thanks. Including this diagram for documentation
image

crates/plugin/src/lib.rs
bbednarski9: No I dont think so. We can roll this into ABI v1 with this release and consider it stable from hereon

willkill07 (approved):

PR #337, opened 2026-06-30
fix: manage Python dynamic plugin environments

willkill07:
#### Overview

Fix the Python dynamic plugin lifecycle regressions discovered by end-to-end testing after #310. Python plugins are now installed into Relay-managed virtual environments, activated with those environments, and removed safely with their owned resources.

#### Details

- Require `source.manifest_root` for Python worker manifests installed through `plugins add`.
- Provision a deterministic per-plugin virtual environment, install the manifest root non-editably, persist `environment_ref`, and roll back partial installs.
- Require lifecycle-managed `environment_ref` for Python worker activation. Direct or manually configured Python worker loading without lifecycle state is intentionally rejected.
- Report environment readiness through lifecycle validation, inspect output, list state, and required startup preflight.
- Tombstone before removal, guard cleanup to the expected Relay-managed directory, support retry after partial cleanup, and provision a fresh environment when reviving a plugin.
- Configure a fixed valid gRPC authority for Unix-domain Python host channels so callbacks can await host-runtime calls such as `emit_mark`.
- Correct the Python worker example manifest and documentation, including the integrity digest and `nemo-relay --bind 127.0.0.1:4040` daemon command.
- Add lifecycle, direct-loading rejection, Unix authority, real callback round-trip, and hermetic end-to-end coverage.

Validation:

- `cargo test -p nemo-relay-cli --no-fail-fast`
- `just test-rust`
- `just test-python-plugin`
- `just test-python`
- `just test-node`
- `just test-go`
- `just test-wasm`
- `just docs`
- Python, Node.js, WebAssembly, and Python plugin package checks

Start with `crates/cli/src/plugins/lifecycle/environment.rs` for environment ownership and provisioning, then `crates/core/src/plugin/dynamic/worker.rs` for the strict activation contract. The real `emit_mark` callback regression is in `crates/core/tests/integration/worker_plugin_tests.rs`.

- Relates to #310

## Summary by CodeRabbit

* **New Features**
  * Added lifecycle-managed per-plugin Python environments for dynamic workers, including environment tracking from add → activation/enable → inspect/validate → runtime launches.
  * Added `environment_state` reporting in lifecycle command responses.
* **Bug Fixes**
  * Activation/enable now correctly blocks when the managed runtime environment is missing/unavailable, and surfaces clearer environment-related validation/last-error messages.
  * Removal now deletes the correct managed environment (when present) and updates validation status accordingly.
* **Documentation**
  * Updated the Python SDK setup and the Python gRPC worker example to reflect registration-based managed environment behavior.
* **Tests**
  * Expanded unit, integration, and end-to-end tests for provisioning, rollback, refusal, and cleanup scenarios.


crates/cli/src/plugins/lifecycle/environment.rs
```diff
@@ -0,0 +1,258 @@
...
+        )
+    })?;
+    if !manifest_root.is_dir() {
+        return Err(format!(
+            "Python plugin source.manifest_root {} is not a directory",
+            manifest_root.display()
+        ));
+    }
+
+    let environment = managed_environment_path(state_path, &manifest.plugin.id)?;
```
zhongxuanwang-nv, line 81: nitty thing, here and :166 all would hash the un-trimmed `plugin.id`, but `worker.rs:749` hashes the `spec.plugin_id` which is the trimmed `metadata.id`. Should we explicitly trim it here?

crates/core/tests/integration/worker_plugin_tests.rs
```diff
@@ -761,7 +810,73 @@ fn python_worker_uses_configured_interpreter() {
...
-    assert!(error.contains("failed to spawn python worker"), "{error}");
+    assert!(
+        error.contains("configured Python worker environment interpreter")
+            && error.contains("does not exist"),
+        "{error}"
+    );
+}
+
+#[tokio::test]
+async fn python_worker_host_runtime_mark_and_mutated_request_round_trip() {
```
zhongxuanwang-nv, line 884: A nitty thing, should these be wired into the CI test too? Because currently they are only reachable by `test-python-plugin` --> `test-python-plugin-e2e`, but no workflow under `.github/workflows` invokes the `test-python-plugin`

crates/cli/src/plugins/lifecycle/environment.rs
```diff
@@ -0,0 +1,258 @@
...
+        manifest_root.as_os_str().to_owned(),
+    ];
+    if let Err(error) = runner.run(environment_python.as_os_str(), &install_args) {
+        let _ = remove_directory_if_present(&environment, "clean up");
+        return Err(format!(
+            "failed to install Python plugin from {} into {}: {error}",
+            manifest_root.display(),
+            environment.display()
+        ));
+    }
```
zhongxuanwang-nv, line 133: Also a nitty thing, `verify_integrity` (trust.rs:235-256) hashes only source.artifact (one file), but here then runs installs the whole tree, that the digest/signature may not cover. Under a signature-required policy a correctly-signed worker.py would pass while unsigned build code + arbitrary deps run. Should we document this or is this meant to gate untrusted-but-signed plugins?

zhongxuanwang-nv (approved):
Thanks Will! :)

zhongxuanwang-nv (approved):

zhongxuanwang-nv (approved):

zhongxuanwang-nv (approved):

PR #345, opened 2026-07-01
fix: report daemon bind and exporter status

willkill07:
#### Overview

Add informative startup output to daemon mode so `nemo-relay --bind ...` reports the bound gateway address and configured exporter destinations. This addresses RELAY-400.

#### Details

Daemon startup now renders the same compact status frame used by transparent agent launches. It reports the listener's actual address, including the assigned port for `--bind ...:0`, and lists configured observability destinations. Existing exporter discovery is reused, and the output is written to stderr with terminal-aware color handling.

Start with `crates/cli/src/server.rs`, especially the startup status rendering and its call immediately after listener binding. The focused regression test is in `crates/cli/tests/coverage/server_tests.rs`.

Validation completed:

- Focused startup-status test passed.
- CLI tests passed: 575 unit tests and 49 integration tests.
- Targeted pre-commit checks, formatting, and cargo check passed.
- `just test-rust` passed for core, adaptive, CLI, and other exercised targets but reported five unrelated existing FFI test failures.

- Relates to: RELAY-400 (Linear)

## Summary by CodeRabbit

* **New Features**
  * Added a startup status banner that shows the bound gateway address and any configured exporter endpoints when the service launches.
  * Live status output is now rendered in a bordered frame, and startup/status formatting adapts to terminal color support while respecting `NO_COLOR`.
* **Bug Fixes**
  * Startup output now clearly indicates when no exporter destinations are configured.
* **Tests**
  * Added unit tests to verify startup banner formatting for both configured and unconfigured exporter scenarios.

### DIFF
--- a/crates/cli/src/server.rs
+++ b/crates/cli/src/server.rs
@@ -3,6 +3,7 @@
 
 use std::future::Future;
 use std::pin::Pin;
+use std::net::SocketAddr;
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
 
@@ -67,6 +68,7 @@ pub(crate) async fn serve_with_dynamic(
             CliError::Io(err)
         }
     })?;
+    print_startup_status(listener.local_addr()?, &config);
     serve_listener_with_dynamic_inner(
         listener,
         config,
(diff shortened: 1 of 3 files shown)


crates/cli/src/server.rs
```diff
@@ -76,6 +78,37 @@ pub(crate) async fn serve_with_dynamic(
...
+                "  {}{}",
+                if index == 0 {
+                    "Exporters      "
+                } else {
+                    "               "
+                },
+                destination
+            ));
+        }
+    }
```
zhongxuanwang-nv, line 107: A nitty thing, when calling `exporter_destinations` here, there might be a previous issue that would always print a local  `output_directory/filename_template` even though `atif.storage` is non-empty and the runtime skips the local file and uploads only to remote backends, when it should just not print anything or print the remote desinations (sanitized)?

crates/cli/tests/coverage/server_tests.rs
```diff
@@ -170,6 +170,40 @@ fn test_config() -> GatewayConfig {
...
+    let output = render_startup_status("127.0.0.1:4567".parse().unwrap(), &config, false);
+
+    assert!(output.contains("NeMo Relay"));
+    assert!(output.contains("Gateway        http://127.0.0.1:4567"));
+    assert!(output.contains("OpenTelemetry http://127.0.0.1:4318/v1/traces"));
+}
+
+#[test]
+fn startup_status_reports_not_configured_when_no_exporters() {
+    let output = render_startup_status("127.0.0.1:4567".parse().unwrap(), &test_config(), false);
```
zhongxuanwang-nv, line 202: Nitty thing, should we also test the port 0 to see whether it's actually using a OS-assigned port instead?

zhongxuanwang-nv (approved):
Thanks Will! All super nitty things!

crates/cli/tests/coverage/server_tests.rs
willkill07: This test is testing rendering rather than universal functionality.

zhongxuanwang-nv: Right! yeah, i would take this comment back :)

crates/cli/src/server.rs
willkill07: > there might be a previous issue that would always print a local output_directory/filename_template

I need clarification or better wording on this.

> [if] atif.storage is non-empty and the runtime skips the local file and uploads only to remote backends, when it should just not print anything or print the remote desinations (sanitized)?

I'd expect that if there are any destinations configured, it will render them. This behavior has already existed prior to this PR.

zhongxuanwang-nv: > I need clarification or better wording on this.

Yes! Here `server.rs:84` reuses `exporter_destinations`, whose ATIF branch (launcher.rs:657-666) always prints a local `output_directory/filename_template/` But when we only uploads the ATIF to the remote backends, the banner might incorrectly show a local path that's never written and hides the real remote target. I could also reproduce this and create a follow-up PR, because this is probably out-of-scope for this one.

> > [if] atif.storage is non-empty and the runtime skips the local file and uploads only to remote backends, when it should just not print anything or print the remote desinations (sanitized)?

> I'd expect that if there are any destinations configured, it will render them. This behavior has already existed prior to this PR.

Yes, I think the rendering is definitely fine; the issue I mentioned is probably out-of-scope for this PR, I can also create a follow-up PR on this

willkill07: Yes. Followup PR makes sense.

PR #346, opened 2026-07-01
fix!: streamline plugin configuration

willkill07:
#### Overview

Streamline CLI plugin configuration around `plugins.toml` and add a hidden path override for controlled runtime launches.

#### Details

- Added hidden `--plugin-config-path` support with `NEMO_RELAY_PLUGIN_CONFIG_PATH` for daemon and `run` configuration.
- Removed the `--plugin-config` CLI inputs, including hook forwarding.
- Removed plugin configuration from `config.toml`; plugin activation now comes from `plugins.toml`.
- Updated setup behavior, documentation, and regression coverage for path precedence and removed configuration forms.
- Breaking changes: users must move plugin configuration out of `config.toml` and replace `--plugin-config` usage with a plugin file.

Validation:

- `just test-rust`
- `uv run pre-commit run --files ...`
- `just docs`

Start with `crates/cli/src/config.rs` for source precedence and removal of inline plugin configuration, then review `crates/cli/tests/coverage/config_tests.rs` and `crates/cli/tests/cli_tests.rs` for the new path override and CLI behavior.

- Closes RELAY-401

## Summary by CodeRabbit

* **New Features**
  * Plugin configuration is now loaded from `plugins.toml`, with support for an override that changes the load location.
* **Bug Fixes**
  * `config.toml` no longer supports a `[plugins]` section; plugin configuration must live in `plugins.toml`.
  * Hook-forward requests no longer include plugin configuration in HTTP headers.
  * Setup merge mode now drops the existing top-level `[plugins]` section and focuses on updating the relevant `[agents.*]` blocks.
* **Documentation**
  * Updated plugin discovery and CLI usage docs to remove `--plugin-config` and reflect the new configuration behavior.


dagardner-nv (approved):

crates/cli/src/setup/model.rs
```diff
@@ -175,22 +175,11 @@ pub(super) fn write_or_merge(
         .parse()
         .map_err(|err| CliError::Config(format!("could not parse existing config: {err}")))?;
     let agent_key = agent_key_and_command(agent).0;
-    // `plugins` is not wizard-owned (users may hand-edit it). Preserve on omission.
-    merge_section(&mut existing, doc, "plugins");
     merge_agents_entry(&mut existing, doc, agent_key);
```
mnajafian-nv, line 181: Should we strip any legacy `[plugins]` block here before writing the merged config back? `load_shared_config` now rejects top-level `plugins` in `config.toml`, so a user who re-runs setup in merge mode with an older file could otherwise save successfully and then fail on the next launch. It may be worth making this path self-heal that migration case.

crates/cli/src/config.rs
```diff
@@ -884,7 +858,13 @@ fn config_paths(explicit: Option<&PathBuf>) -> Vec<PathBuf> {
 // Returns the plugin config search path. An explicit gateway config path scopes plugins.toml to the
 // same directory so `--config path/to/config.toml` can be extended by `path/to/plugins.toml` without
 // reading unrelated implicit project/user/global plugin files.
-fn plugin_config_paths(explicit: Option<&PathBuf>) -> Vec<PathBuf> {
+fn plugin_config_paths(
+    explicit: Option<&PathBuf>,
+    plugin_config_path: Option<&PathBuf>,
+) -> Vec<PathBuf> {
+    if let Some(path) = plugin_config_path {
```
mnajafian-nv, line 889: One thing I wanted to sanity-check is the contract for this explicit override branch. When `plugin_config_path` is set, do we want to require that file to exist rather than treating it like an optional discovered candidate? As written, a typo in `NEMO_RELAY_PLUGIN_CONFIG_PATH` looks like it can fall through to “no plugin config” and silently disable plugin activation for that run.

mnajafian-nv (approved):
LGTM overall. Thanks for driving this cleanup. Conditional approval pending review of the inline suggestions.

crates/cli/src/config.rs
willkill07: Since it's undocumented and only intended for interaction with Fabric, I don't have a strong preference here.

crates/cli/src/setup/model.rs
willkill07: addressing with commit for self-healing (not full extraction to plugins.toml though)

mnajafian-nv (approved):
LGTM!

lvojtku (approved):
Approved with no comments

bbednarski9 (approved):

PR #357, opened 2026-07-02
feat: add setup readiness diagnostics

willkill07:
#### Overview

Add unified `nemo-relay doctor` setup-readiness diagnostics for agents, discovered plugin configuration, and persistent Codex/Claude Code host-plugin installations.

#### Details

- Share host-plugin readiness collection between default doctor and focused plugin doctor, including generated manifest validation, registration, routing/hooks, remediation, and informational lazy-sidecar status.
- Report discovered `plugins.toml` layers, effective-source provenance, semantic source errors, merged validation, and secret-safe JSON fields while retaining schema version 1.
- Add stable `Plugin configuration` and `Host plugins` output sections, diagnostics tests, and CLI documentation.

Start with `crates/cli/src/doctor.rs` for the unified report and rendering, then `crates/cli/src/plugin_install/mod.rs` for shared host checks and `crates/cli/src/config.rs` for provenance.

- Closes: RELAY-316

## Summary by CodeRabbit

* **New Features**
  * Enhanced `nemo-relay doctor` plugin diagnostics with per-layer plugin configuration and an overall “Plugin resolution” check.
  * Added structured host-plugin readiness reporting in both human output and `doctor-json`, including per-check results and remediation guidance.
* **Bug Fixes**
  * `plugins.toml` validation errors now report the originating `plugins.toml` source file.
  * `doctor` now fails when persistent host plugins are installed but host readiness is not OK.
* **Documentation**
  * Expanded plugin installation “Diagnose” guidance (`--plugin` / `--install-dir`) and forced-reinstall remediation details.
  * Reformatted multiple “Why Use It?” / “What You Get” README bullet lists for consistent styling.

### DIFF
--- a/crates/cli/README.md
+++ b/crates/cli/README.md
@@ -26,28 +26,28 @@ with the installed `nemo-relay` command rather than link against the crate.
 
 ## Why Use It?
 
-- 🧭 **Observe existing coding agents**: Run Claude Code, Codex, or Hermes
+- **Observe existing coding agents**: Run Claude Code, Codex, or Hermes
   Agent through a local NeMo Relay gateway without changing the agent
   itself.
-- 🛠️ **Configure hooks interactively**: Use the setup wizard to write project or
+- **Configure hooks interactively**: Use the setup wizard to write project or
   user config and install the hook files needed by supported agents.
-- 📡 **Export local sessions**: Write ATIF trajectory files, ATOF event JSONL
+- **Export local sessions**: Write ATIF trajectory files, ATOF event JSONL
   streams, or OpenInference spans from one shared config model.
-- 🩺 **Diagnose setup readiness**: Check config layers, `plugins.toml` discovery,
+- **Diagnose setup readiness**: Check config layers, `plugins.toml` discovery,
   agent binaries, persistent host-plugin installs, hook status, observability
   outputs, and shell completions with `nemo-relay doctor`.
 
 ## What You Get
 
-- ✅ **`nemo-relay` binary**: The executable installed by the `nemo-relay-cli`
+- **`nemo-relay` binary**: The executable installed by the `nemo-relay-cli`
   Cargo package.
-- ✅ **First-run setup**: Bare `nemo-relay` launches setup when no config exists,
+- **First-run setup**: Bare `nemo-relay` launches setup when no config exists,
   then runs doctor once config is present.
-- ✅ **Agent shortcuts**: `nemo-relay claude`, `nemo-relay codex`, and
+- **Agent shortcuts**: `nemo-relay claude`, `nemo-relay codex`, and
   `nemo-relay hermes` start observed agent runs.
-- ✅ **Config-driven launch**: `nemo-relay run` resolves config, environment, and
+- **Config-driven launch**: `nemo-relay run` resolves config, environment, and
   CLI overrides for deterministic non-interactive use.
-- ✅ **Hook forwarding server**: A local gateway accepts agent hook events and
+- **Hook forwarding server**: A local gateway accepts agent hook events and
   provider-shaped OpenAI or Anthropic requests.
 
 ## Installation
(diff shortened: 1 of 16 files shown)


crates/cli/README.md
```diff
@@ -33,8 +33,9 @@ with the installed `nemo-relay` command rather than link against the crate.
   user config and install the hook files needed by supported agents.
 - 📡 **Export local sessions**: Write ATIF trajectory files, ATOF event JSONL
   streams, or OpenInference spans from one shared config model.
-- 🩺 **Diagnose the machine**: Check config layers, agent binaries, hook status,
-  observability outputs, and shell completions with `nemo-relay doctor`.
+- 🩺 **Diagnose setup readiness**: Check config layers, `plugins.toml` discovery,
```
lvojtku: Please remove all emojis from documentation.

lvojtku (changes requested):
Please remove all emojis from the docs.

crates/cli/src/plugin_install/mod.rs
```diff
@@ -325,24 +492,212 @@ fn doctor_host_json_value(
...
+    }
+
+    let host_cli_check = require_host_cli(host, options, runner);
+    readiness.push(
+        "Host CLI",
+        host_cli_check
+            .as_ref()
+            .map(|_| format!("{} is available", host_cli(host)))
+            .map_err(Clone::clone),
+    );
```
zhongxuanwang-nv, line 592: A nitty thing, here it does a full JSON equality of installed `plugin.json` against the `plugin_manifest(host)` which embeds env!("CARGO_PKG_VERSION") (marketplace.rs:105); 

If we don't reinstall after CLI upgrades, the the stored version differs from the running binary's, so the "Generated plugin" check (mod.rs:486) fails and exit_code returns 1 (doctor.rs:1299) for both nemo-relay doctor and bare nemo-relay, so a healthy install will report an error. Is this expected that the users should always reinstall?

crates/cli/src/plugin_install/mod.rs
```diff
@@ -305,18 +464,26 @@ fn doctor_host(
...
+        let marker = if check.ok { "ok" } else { "failed" };
+        println!("{}: {marker} ({})", check.name, check.details);
+    }
+    readiness.ok().then_some(()).ok_or_else(|| {
+        format!(
+            "{} plugin doctor checks failed; run `nemo-relay install {} --force` to repair the installation",
+            host_label(host),
+            host_arg(host)
+        )
+    })
```
zhongxuanwang-nv, line 486: Should we also check Claude's generated `hooks/hooks.json` using `plugin_hooks`, similar to how the marketplace manifest and the plugin manifest are checked?

crates/cli/src/doctor.rs
```diff
@@ -155,8 +175,10 @@ pub(crate) async fn collect_report(
             resolution,
             configured_agents,
             &resolved.dynamic_plugins,
+            &plugin_diagnostics,
         ),
         agents: collect_agents(target_agent, &resolved).await,
+        host_plugins: crate::plugin_install::collect_default_host_plugin_readiness(),
```
zhongxuanwang-nv, line 181: Should we also consider adding a timeout for this, considering in `collect_agents`'s `probe_version` function, it is wrapped in `tokio::time::timeout`, so a stalled host CLI may hang the `nemo-relay doctor`?

zhongxuanwang-nv (approved):
Verified these potential nitty things with powerful AIs, and generally LGTM! Thanks Will!

crates/cli/src/plugin_install/mod.rs
willkill07: we can do a followup that inspects hooks. The reason why i didn't is that we don't have markers injected (yet) and without them I'd be hesitant to read files not owned by Relay

crates/cli/src/doctor.rs
willkill07: i can't see how it could stall(?)

crates/cli/src/plugin_install/mod.rs
willkill07: It is already fixed in the current branch. `without_version` (line 656 of mod.rs) removes the volatile version before manifest comparison, and a regression test covers an older generated plugin manifest.

lvojtku (approved):

PR #364, opened 2026-07-06
feat(plugin): add owned dynamic host activation

bbednarski9:
#### Overview

This promotes dynamic-plugin activation from CLI-private orchestration into a core-owned lifecycle for harness-native embedding hosts. It deliberately does not replace the existing static plugin lifecycle: static-only callers continue to use normal plugin initialization, while `PluginHostActivation` requires at least one explicit dynamic plugin specification.

#### Lifecycle contract

- `initialize_plugins*` / `clear_plugin_configuration` remain the static-only API.
- `PluginHostActivation::activate(base_config, dynamic_specs)` is the owned dynamic API and rejects an empty `dynamic_specs` list with an actionable error.
- The base config may include registered static components. Those initialize first; components contributed by loaded plugins are appended afterward, all in one transaction.
- A process has one active plugin-configuration owner. Calling dynamic activation after static initialization is rejected; mixed users must pass static components in the dynamic activation's base config before initialization.
- Manifest and worker-environment references are explicit paths resolved by the embedding host. Relay reads them during startup activation; this API does not discover installation state or continuously revalidate mutable files.

#### Details

- Add public `PluginHostActivation` and `DynamicPluginActivationSpec` APIs for native and worker plugins.
- Load plugin runtimes, append their configured components, and initialize the complete configuration transactionally.
- Reject duplicate dynamic IDs before loading and protect built-in plugin IDs from preclaimed implementations.
- Track registry registrations by unique ownership ID so teardown removes only the adapter it installed; a replacement is preserved and reported as a conflict.
- Run plugin mutations on a process-wide, cancellation-resistant Relay executor with a stable Tokio runtime. Callers must not rely on activation callbacks running on their thread or runtime.
- Clear callbacks and flush subscribers before deregistering owned plugin kinds, stopping workers, and unloading libraries.
- Retain runtimes and the host owner when cleanup cannot prove safe unloading; otherwise release ownership while surfacing teardown errors.
- Provide explicit `clear()` plus best-effort `Drop` cleanup that logs failures.
- Refactor the CLI server to retain the legacy static lifecycle for static-only configurations and use the owned host only when dynamic specifications are present. CLI discovery and sidecar behavior are unchanged.

Start with `crates/core/src/plugin/dynamic/host.rs` for the public lifecycle and executor contract, then `crates/core/src/plugin.rs` for ownership and transactional rollback. `crates/cli/src/server.rs` shows the static/dynamic dispatch. Native and worker integration tests cover mixed static/dynamic activation, conflicts, rollback, in-flight callbacks during teardown, cancellation, explicit clear, and drop cleanup.

#### Validation

- `cargo clippy -p nemo-relay --features worker-grpc --all-targets -- -D warnings`
- Native plugin-host integration suite: 13 passed
- Worker shutdown/recovery regression passed
- Built-in ownership regression passed
- Static activation cancellation regression passed
- Static-only CLI lifecycle tests passed

None.

## Summary by CodeRabbit

* **New Features**
  * Added an owned activation/teardown lifecycle for dynamic native and worker plugins via an activation handle.
  * Introduced tracked plugin ownership and stronger protections for built-in plugin kinds to prevent conflicting activations.
* **Bug Fixes**
  * Made plugin initialization rollback/deregistration panic-safe with more reliable, observable recovery on partial failures.
  * Improved dynamic host clearing and shutdown safety, including safer worker shutdown behavior and clearer error reporting.
* **Tests**
  * Expanded CLI coverage and core integration/unit tests for ownership, builtin reservation, clear semantics, cancellation/rollback, and worker lifecycle.
* **Documentation**
  * —

### DIFF
--- a/crates/core/src/plugin.rs
+++ b/crates/core/src/plugin.rs
@@ -49,15 +49,34 @@ pub use nemo_relay_types::plugin::{ConfigDiagnostic, DiagnosticLevel};
 pub mod dynamic;
 pub use dynamic::*;
 
-type PluginMap = HashMap<String, Arc<dyn Plugin>>;
+type PluginMap = HashMap<String, RegisteredPlugin>;
+
+struct RegisteredPlugin {
+    registration_id: u64,
+    owner: PluginRegistrationOwner,
+    plugin: Arc<dyn Plugin>,
+}
+
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum PluginRegistrationOwner {
+    Builtin,
+    External,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum PluginDeregistrationOutcome {
+    Removed,
+    Missing,
+    Replaced,
+}
 
 static PLUGIN_HANDLERS: LazyLock<RwLock<PluginMap>> = LazyLock::new(|| RwLock::new(HashMap::new()));
 static ACTIVE_PLUGIN_CONFIGURATION: LazyLock<Mutex<Option<ActivePluginConfiguration>>> =
     LazyLock::new(|| Mutex::new(None));
 static PLUGIN_MUTATION_OWNER: LazyLock<Mutex<PluginMutationOwner>> =
     LazyLock::new(|| Mutex::new(PluginMutationOwner::Idle));
+static NEXT_PLUGIN_REGISTRATION_ID: AtomicU64 = AtomicU64::new(1);
 static NEXT_PLUGIN_HOST_OWNER_ID: AtomicU64 = AtomicU64::new(1);
-static BUILTIN_PLUGIN_REGISTRATION: OnceLock<Result<()>> = OnceLock::new();
 static PLUGIN_MUTATION_EXECUTOR: OnceLock<PluginMutationSender> = OnceLock::new();
 
 type PluginMutationJob = Pin<Box<dyn Future<Output = ()> + Send + 'static>>;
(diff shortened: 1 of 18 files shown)


mnajafian-nv (commented):
Thanks, this is moving in a good direction. I left two focused inline comments on validation/activation alignment and the retained-owner activity contract.

crates/core/src/plugin.rs
```diff
@@ -908,10 +985,11 @@ pub fn list_plugin_kinds() -> Vec<String> {
 /// The returned plugin is shared by [`Arc`], so callers receive a cheap clone.
 pub fn lookup_plugin(plugin_kind: &str) -> Option<Arc<dyn Plugin>> {
     let _ = ensure_builtin_plugins_registered();
```
mnajafian-nv: One thing I’d tighten before merge: this drops builtin-registration failure on the read path, while activation now treats the same condition as a hard error. With a preclaimed builtin kind, that can let validation or lookup-based checks proceed and only fail later during activation. I’d rather keep those paths aligned by surfacing the failure here, or add a regression that makes the intended mismatch explicit if that behavior is deliberate.

crates/core/src/plugin/dynamic/host.rs
```diff
@@ -0,0 +1,425 @@
...
+    /// Clear registered callbacks before unloading libraries and workers.
+    pub fn clear(mut self) -> Result<()> {
+        self.clear_inner()
+    }
+
+    fn clear_inner(&mut self) -> Result<()> {
+        if !self.active {
+            return Ok(());
+        }
+        self.active = false;
```
mnajafian-nv, line 228: One contract detail seems worth making explicit before merge: `self.active` flips to `false` before teardown is proven safe, but later paths intentionally retain the runtimes and host claim when callbacks cannot be cleared or unload is unsafe. In that state a binding can report inactive even though a new activation would still conflict on retained ownership. I’d either keep `active` true until the claim is actually released, or document and test that `is_active()` means 'this handle is no longer reusable' rather than 'the process is ready for another activation.'

willkill07 (approved):

bbednarski9:
CLI lifecycle selection:

```mermaid
flowchart TD
    A["Relay CLI startup"] --> B{"Any plugin configuration"}
    B -->|"No"| C["Run without plugin activation"]
    B -->|"Yes"| D{"Dynamic plugin specs present"}
    D -->|"No"| E["Use existing static initialization"]
    D -->|"Yes"| F["Create PluginHostActivation"]
    E --> G["Run gateway"]
    F --> G
    G --> H["Close sessions and flush subscribers"]
    H --> I{"Activation type"}
    I -->|"Static"| J["Clear static configuration"]
    I -->|"Dynamic"| K["Clear dynamic host"]
```

bbednarski9:
Dynamic activation:

```mermaid
sequenceDiagram
    participant Caller
    participant Executor as Lifecycle executor
    participant Owner as Process owner
    participant Loader
    participant Registry
    participant Config as Configuration runtime

    Caller->>Executor: Activate base config and specs
    Executor->>Executor: Validate nonempty and unique IDs
    Executor->>Owner: Acquire exclusive host lease
    Owner-->>Executor: Return owner ID
    Executor->>Registry: Ensure built-in plugin kinds
    Executor->>Loader: Load native libraries
    Executor->>Loader: Start worker processes
    Loader->>Registry: Register adapters with registration IDs
    Executor->>Config: Append dynamic components after static components
    Executor->>Config: Validate and initialize components

    alt Initialization succeeds
        Config-->>Caller: Activation guard and report
    else Initialization fails cleanly
        Config->>Registry: Roll back component registrations
        Loader->>Registry: Deregister adapters
        Loader->>Loader: Unload libraries and workers
        Owner->>Owner: Release lease
        Config-->>Caller: Activation error
    else Rollback cannot prove safety
        Loader->>Loader: Retain loaded runtimes
        Owner->>Owner: Retain lease
        Config-->>Caller: Incomplete rollback error
    end
```

bbednarski9:
Safe teardown:

```mermaid
flowchart TD
    A["Clear activation"] --> B["Flush subscribers and remove component callbacks"]
    B --> C{"Callbacks proven removed"}
    C -->|"No"| D["Retain runtimes and owner for process lifetime"]
    D --> E["Return teardown error"]
    C -->|"Yes"| F["Deregister plugin adapters by registration ID"]
    F --> G{"Adapters safe to unload"}
    G -->|"No"| D
    G -->|"Yes"| H["Shut down worker processes"]
    H --> I{"Worker shutdown safe"}
    I -->|"No"| D
    I -->|"Yes"| J["Unload libraries and runtimes"]
    J --> K["Release process owner"]
    K --> L["Return success or recoverable teardown error"]
```

willkill07 (approved):

PR #366, opened 2026-07-06
feat(node): initialize with dynamic plugins

bbednarski9:
#### Overview

This adds an owned Node binding for explicit native and worker dynamic-plugin activation. The shared process-wide activation lifecycle landed in #364, and discovery-aware configuration layering landed in #418. This PR now contains only the Node binding and its tests.

Static-only harness-native applications keep using `plugin.initialize()` / `plugin.clear()`. `initializeWithDynamicPlugins(...)` requires at least one dynamic specification. For harness-native activation, the supplied config is layered over discovered `plugins.toml` configuration; static components from that effective config initialize before components appended by the dynamic plugins.

#### Details

- Add `initializeWithDynamicPlugins(config, specs)` with a dedicated JavaScript/TypeScript specification shape for native and worker plugins.
- Return an owned activation exposing `report`, `active`, async `close()`, and `[Symbol.asyncDispose]()` for deterministic `await using` cleanup.
- Define `active` as handle state: it becomes false when teardown begins and does not by itself guarantee that process-wide ownership has been released after a teardown failure.
- Share one detached close operation across repeated and concurrent callers; every caller observes the same teardown result.
- Run cleanup off the JavaScript thread and retain defensive finalization when callers omit explicit close.
- Preserve existing static `plugin.initialize()` and `plugin.clear()` behavior. Empty dynamic activation is rejected without claiming process ownership, so static initialization remains available.
- Keep complete core validation, manifest, load, ownership, and teardown diagnostics in rejected promises.
- Add neutral native and worker fixture coverage for discovered and explicit static configuration, managed LLM/tool callbacks, callback absence after close, partial-load rollback, conflicts, repeated/concurrent close, structured disposal, cancellation-safe cleanup, and garbage collection.

Start with `crates/node/src/api/mod.rs` for N-API ownership and cleanup, then `crates/node/plugin.d.ts` and `crates/node/plugin.js` for the `initializeWithDynamicPlugins` public contract and deliberate absence of an old-name alias. The lifecycle cases are in `crates/node/tests/dynamic_plugin_tests.mjs`.

#### Validation

- `npm run build-debug`
- Dynamic-plugin Node tests — 9 passed
- Full Node suite — 263 passed
- `npm run check:docstrings --workspace crates/node`
- `cargo test -p nemo-relay-node`
- `cargo clippy -p nemo-relay-node --all-targets -- -D warnings`
- Generated native/wrapper export consistency and stale-name checks passed
- Changed-file Prettier checks passed
- `git diff --check`

Relates to #364.
Relates to #418.

## Summary by CodeRabbit

- **New Features**
  - Added a Node.js API to activate and manage dynamically resolved plugins (`initializeWithDynamicPlugins`).
  - Supports native and worker dynamic plugins with per-plugin identifiers, optional environment targeting, and per-plugin JSON config.
  - Returns an owned activation handle exposing `active` status and a configuration validation `report`.
  - Provides deterministic, idempotent lifecycle teardown via `close()` and async disposal (`Symbol.asyncDispose`).

- **Bug Fixes**
  - Improved dynamic plugin cleanup for idempotent and concurrent shutdown, including recovery after rejected activations.

- **Tests**
  - Added comprehensive dynamic plugin host tests covering lifecycle, disposal, concurrency, and finalization behavior (native + worker).

### DIFF
--- a/crates/node/plugin.d.ts
+++ b/crates/node/plugin.d.ts
@@ -1,6 +1,8 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+/// <reference lib="esnext.disposable" />
+
 import type { EventSanitizeFields, Json } from './index';
 
 /** Policy behavior for unsupported configuration. */
--- a/crates/node/plugin.d.ts
+++ b/crates/node/plugin.d.ts
@@ -61,7 +61,10 @@ export interface DynamicPluginActivationSpec {
 export interface DynamicPluginActivation {
   /** Validation report produced by the successful activation. */
   readonly report: ConfigReport;
-  /** Whether this object still owns the dynamic plugin host. */
+  /**
+   * Whether this activation handle has not begun teardown. `false` does not
+   * guarantee another process-wide activation can start after failed teardown.
+   */
   readonly active: boolean;
   /** Clear callbacks before unloading libraries and workers. Idempotent. */
   close(): Promise<void>;
--- a/crates/node/plugin.d.ts
+++ b/crates/node/plugin.d.ts
@@ -327,7 +327,7 @@ export declare function validate(config: PluginConfig): ConfigReport;
  */
 export declare function initialize(config: PluginConfig): Promise<ConfigReport>;
 /**
- * Load and activate explicitly resolved dynamic plugins.
+ * Initialize with explicitly resolved dynamic plugins.
  *
  * The returned object owns loaded libraries and worker processes. Keep it
  * alive while plugin callbacks may run and call `close()` for deterministic
@@ -339,7 +339,7 @@ export declare function initialize(config: PluginConfig): Promise<ConfigReport>;
  * @remarks File-configured static components initialize before dynamic
  * components. Use `initialize()` for a static-only configuration.
  */
-export declare function activateDynamicPlugins(
+export declare function initializeWithDynamicPlugins(
   config: PluginConfig,
   specs: DynamicPluginActivationSpec[],
 ): Promise<DynamicPluginActivation>;
--- a/crates/node/plugin.d.ts
+++ b/crates/node/plugin.d.ts
@@ -326,9 +326,10 @@ export declare function initialize(config: PluginConfig): Promise<ConfigReport>;
  * alive while plugin callbacks may run and call `close()` for deterministic
  * teardown. Garbage collection is a defensive fallback only.
  *
- * @param config - Base plugin configuration activated alongside dynamic components.
- * @param specs - Explicit manifest and component configuration for each plugin.
+ * @param config - Base configuration activated first, including any static components.
+ * @param specs - Non-empty explicit manifest and component configuration for each plugin.
  * @returns The owned activation and its validation report.
+ * @remarks Use `initialize()` for a static-only configuration.
  */
 export declare function activateDynamicPlugins(
   config: PluginConfig,
--- a/crates/node/plugin.d.ts
+++ b/crates/node/plugin.d.ts
@@ -45,6 +45,28 @@ export interface PluginConfig {
   policy?: ConfigPolicy;
 }
 
+/** Execution lane for a dynamically loaded Relay plugin. */
+export type DynamicPluginKind = 'rust_dynamic' | 'worker';
+
+/** Explicitly resolved dynamic plugin load and component configuration. */
+export interface DynamicPluginActivationSpec {
+  pluginId: string;
+  kind: DynamicPluginKind;
+  manifestRef: string;
+  environmentRef?: string | null;
+  config?: Record<string, Json>;
+}
+
+/** Owns one process-wide dynamic plugin host activation. */
+export interface DynamicPluginActivation {
+  /** Validation report produced by the successful activation. */
+  readonly report: ConfigReport;
+  /** Whether this object still owns the dynamic plugin host. */
+  readonly active: boolean;
+  /** Clear callbacks before unloading libraries and workers. Idempotent. */
+  close(): Promise<void>;
+}
+
 /** A mark Relay materializes under a managed lifecycle. */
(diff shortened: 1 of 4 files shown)


crates/node/plugin.js
```diff
@@ -78,6 +78,19 @@ function initialize(config) {
   return lib.initializePlugins(config);
 }
 
+/**
+ * Load and activate explicitly resolved dynamic plugins.
+ *
+ * @param {object} config - Base plugin configuration document.
+ * @param {Array<object>} specs - Native-library or worker plugin load specifications.
```
AteebNoOne: Given that specs crosses the boundary to the native layer, Array<object> is a bit broad for consumers relying on intellisense. Consider defining a specific @typedef for the plugin load specifications (e.g., { kind: string, path: string }) so IDEs can catch malformed configs before they hit the Rust runtime.

crates/node/plugin.js
AteebNoOne, line 128: ince deterministic teardown is required for these dynamic plugins, should we implement Symbol.asyncDispose on the returned activation object under the hood? Doing so would allow Node developers to use the modern await using syntax for guaranteed cleanup, even if the runtime throws an error before close() is explicitly called.

AteebNoOne (commented):
Thanks for putting this together! I took a look at the JavaScript bindings and left a quick suggestion on the JSDoc typing to help with intellisense. Looking forward to seeing this land!

crates/node/plugin.d.ts
```diff
@@ -45,6 +45,28 @@ export interface PluginConfig {
...
+  manifestRef: string;
+  environmentRef?: string | null;
+  config?: Record<string, Json>;
+}
+
+/** Owns one process-wide dynamic plugin host activation. */
+export interface DynamicPluginActivation {
+  /** Validation report produced by the successful activation. */
+  readonly report: ConfigReport;
+  /** Whether this object still owns the dynamic plugin host. */
```
mnajafian-nv: I’d tighten this public wording to match the behavior the tests already lock in. `active` becomes `false` as soon as `close()` starts, before the returned promise resolves and before teardown is necessarily complete. Framing this as 'teardown has not started for this handle' or calling out that `false` does not by itself mean a replacement activation can succeed yet would make the Node contract easier to use correctly.

mnajafian-nv (commented):
Thanks, this is in good shape overall. I left one Node-surface comment on the public activation-handle contract.

crates/node/plugin.js
```diff
@@ -78,6 +78,20 @@ function initialize(config) {
...
+ * @param {Array<object>} specs - Non-empty native-library or worker plugin specifications.
+ * @returns {Promise<object>} An owned activation with `report`, `active`, and `close()`.
+ * @remarks Keep the returned activation alive while its callbacks may run and
+ * call `close()` for deterministic teardown. Use `initialize()` for a
+ * static-only configuration.
+ */
+function activateDynamicPlugins(config, specs) {
+  return lib.activateDynamicPlugins(config, specs);
+}
+
```
zhongxuanwang-nv: I think this is the same thing that happened in PR-365, that here they might skip the base `plugins.toml` file

zhongxuanwang-nv (approved):
I think overall this is good. Thanks Bryan!

crates/node/plugin.js
bbednarski9: thanks for flagging this one. I opened #418 to address this

crates/node/plugin.js
bbednarski9: Addressed in 553c2d6d. `plugin.js` now defines a dedicated `DynamicPluginActivationSpec` JSDoc typedef matching the public activation shape and uses it for `specs`, so plain-JavaScript consumers get field and kind IntelliSense before crossing the native boundary. The Node docstring check and changed-file formatting checks pass.

crates/node/tests/dynamic_plugin_tests.mjs
```diff
@@ -0,0 +1,474 @@
...
+      } finally {
+        try {
+          process.kill(workerPid, 'SIGCONT');
+        } catch (error) {
+          if (error.code !== 'ESRCH') {
+            throw error;
+          }
+        }
+        await Promise.all([firstClose, secondClose]);
+      }
```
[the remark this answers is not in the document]
bbednarski9: Addressed in 553c2d6d. Cleanup still records SIGCONT and close failures and preserves operation-error priority, but both prioritized throws now occur after the `finally` block, eliminating the `noUnsafeFinally` finding. The complete dynamic-plugin Node suite passes (9/9).

bbednarski9:
Review scope follow-up: the binding-parity observations are intentional sequencing, not an omitted surface in this PR. Python dynamic activation is covered by #365 and FFI/Go by #368, both built on the shared core activation API. This PR remains scoped to the Node binding. The base `plugins.toml` concern is addressed through #418 and consumed here via the discovery-aware activation path.

crates/node/plugin.js
```diff
@@ -78,6 +78,20 @@ function initialize(config) {
   return lib.initializePlugins(config);
 }
 
+/**
+ * Load and activate explicitly resolved dynamic plugins.
+ *
+ * @param {object} config - Base configuration, including any static components.
+ * @param {Array<object>} specs - Non-empty native-library or worker plugin specifications.
+ * @returns {Promise<object>} An owned activation with `report`, `active`, and `close()`.
```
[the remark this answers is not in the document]
bbednarski9: Addressed in 553c2d6d. The wrapper now declares `DynamicPluginActivationSpec` with `pluginId`, the native/worker kind union, manifest/environment references, and component config, and `specs` is typed as `DynamicPluginActivationSpec[]`. This keeps the plain-JavaScript IntelliSense shape aligned with `plugin.d.ts`; docstring and formatting checks pass.

willkill07:
/ok to test [553c2d6](https://github.com/NVIDIA/NeMo-Relay/pull/366/commits/553c2d6d00e146ab6d4cdc397a06cdd67b2ac508)

willkill07 (approved):

PR #371, opened 2026-07-07
feat(sanitizers)!: add mark and scope event sanitizer registries [1/6]

willkill07:
#### Overview

Adds the shared Rust runtime, plugin, native ABI, and worker protocol infrastructure for sanitizing mark and scope event observability fields without changing existing APIs or ABI v1 offsets.

> [!WARNING]
> **BREAKING CHANGE:** **[Behavioral]** Sanitizer Failures are now fail-closed.
> 
> If a sanitizer is invoked within the path of any event and fails, the event
> being sanitized will be dropped by the Relay runtime. This may result in
> missing observability/event data.
>
> If fail-open behavior is desired, then failure logic must be wrapped inside
> of a sanitizer itself.

#### Details

- Adds `EventSanitizeFields` for `data`, `category_profile`, and `metadata`.
- Adds global and scope-local mark, scope-start, and scope-end registries.
- Extends Rust plugin contexts, the appended native host table, and worker registration surfaces 30–32.
- Applies scope event sanitizers to every category after specialized tool/LLM sanitizers.
- Covers explicit, pending, plugin-runtime, and streaming marks.
- Adds exhaustive ordering, inheritance, cleanup, rollback, ABI, worker, and exporter tests.
- Contains no documentation changes; documentation is isolated in a final independent PR.
- Breaking changes: none.

Validation: `just test-rust`, CI-style Rust coverage (2515 tests), Python/Node/Go binding matrices, formatting, Clippy, and pre-commit.

Start with `crates/types/src/api/event.rs`, `crates/core/src/api/registry.rs`, and the event sanitizer integration tests in `crates/core/tests/integration/api_surface_tests.rs`.

- Relates to: RELAY-409

## Summary by CodeRabbit

* **New Features**
  * Added event field sanitization via `EventSanitizeFields`, with new sanitizer guardrail registration APIs for mark events and scope (start/end) events (global and scope-local).
  * Extended native plugin and worker SDK support with new event-sanitizer surfaces and callback wiring.

* **Bug Fixes**
  * Event sanitization is now consistently applied across scope/mark lifecycle, pending marks, tool/LLM lifecycle, and streaming chunk/end emissions.
  * Events are only dispatched when sanitization succeeds.

* **Tests**
  * Expanded integration/unit coverage for ordering, rewrites, deregistration, and plugin/worker sanitizer metadata.


crates/worker/src/lib.rs
```diff
@@ -192,6 +195,68 @@ impl PluginContext {
...
+        surface: RegistrationSurface,
+        callback: F,
+    ) where
+        F: Fn(&Event, EventSanitizeFields) -> EventSanitizeFields + Send + Sync + 'static,
+    {
+        self.push_registration(name, surface, priority, false);
+        self.handlers
+            .event_sanitizers
+            .insert(name.into(), Arc::new(callback));
+    }
```
zhongxuanwang-nv: nitty thing — here it's using only the `name` as the key, but all three surfaces (MarkSanitizeGuardrail, ScopeSanitizeStartGuardrail, ScopeSanitizeEndGuardrail) share the same `event_sanitizers` map to determine the callbacks, which may have a collision when a worker plugin registers the same name on two of the surfaces, so the second one overwrites the first? 

Should we instead use `(surface, name)` as the key or create per-surface maps?

crates/core/src/plugin/dynamic/native.rs
```diff
@@ -1420,6 +1472,62 @@ fn wrap_event_subscriber(
...
+    instance: Arc<NativePluginInstance>,
+    cb: NemoRelayNativeEventSanitizeCb,
+    user_data: *mut c_void,
+    free_fn: NemoRelayNativeFreeFn,
+) -> EventSanitizeFn {
+    let user_data = make_user_data(instance, user_data, free_fn);
+    Arc::new(move |event, fields| {
+        let fallback = fields.clone();
+        call_event_sanitize_callback(cb, user_data.ptr, event, &fields).unwrap_or(fallback)
+    })
```
zhongxuanwang-nv, line 1485: Nitty thing, it here falls back to the unsanitized fields on any sanitizer failures, `unwrap_or(fields.clone())`. Wondering if this should be accepted, or should we just output generic message instead?

zhongxuanwang-nv (approved):
Used powerful AIs to help; everything looks great!

crates/core/src/plugin/dynamic/native.rs
willkill07: intentional. Fail-open behavior is the only possible remediation.

crates/worker/src/lib.rs
willkill07: This API change would over-specialize in comparison to other middleware. I don't think this is necessary, as the same issue already comes up with other middleware.

crates/core/src/plugin/dynamic/worker.rs
```diff
@@ -812,6 +812,29 @@ impl WorkerPluginInstance {
...
                     )?;
                 }
+                RegistrationSurface::MarkSanitizeGuardrail
+                | RegistrationSurface::ScopeSanitizeStartGuardrail
+                | RegistrationSurface::ScopeSanitizeEndGuardrail => {
+                    let instance = Arc::new(self.clone_for_callback());
+                    let callback_name = name.clone();
+                    let callback = Arc::new(move |event: &Event, fields: EventSanitizeFields| {
+                        instance
+                            .invoke_event_sanitize(&callback_name, surface, event)
```
bbednarski9, line 822: Add a worker integration test with two event sanitizers where the second modifies only one field; verify the first sanitizer’s changes survive.

crates/worker/src/lib.rs
```diff
@@ -192,6 +195,68 @@ impl PluginContext {
             .insert(name.into(), Arc::new(callback));
     }
 
+    fn register_event_sanitizer<F>(
+        &mut self,
+        name: &str,
```
bbednarski9, line 202: If we are keying these just by name for the worker-level registrations, then within the same surface if we register mark and scope sanitizers with the same name, then the latter could replace the prior?

crates/core/src/plugin/dynamic/worker.rs
willkill07: Adding

crates/worker/src/lib.rs
willkill07: mark and scope are separate.

It doesn't prevent overwriting with the same type + name though

bbednarski9:
Nice, as expected the new unit test failed. I think we'll need to patch that

bbednarski9 (changes requested):
Just that unit test failing

bbednarski9 (approved):

PR #382, opened 2026-07-07
feat(observability)!: make mark projection configurable for trace exporters

### LINKED ISSUE #381: Preserve custom mark events in ATIF and visual trace representations
## Summary

NeMo Relay emits custom mark events to subscribers and telemetry exporters, but downstream trajectory and trace views do not consistently preserve them as visible, semantically meaningful events.

The event data is present in the raw event stream and can be present in OTLP span events, yet generic ATIF consumers may only see a `system` step with `extra.event_payload`, while native ATIF import and visual trace views do not expose the mark as an event or step that users can easily inspect.

This makes it difficult to verify event ordering, parentage, and payloads for integrations that rely on custom marks.

## Observed behavior

- A mark with a name, payload, timestamp, and parent scope is emitted by Relay.
- ATOF preserves the mark as a raw event.
- Generic ATIF serialization preserves the payload and ancestry, but represents the mark as a system step without an observation or tool/event representation.
- Native ATIF import into downstream trace tooling may accept the trajectory while omitting the custom mark from the visible trace structure.
- OTLP/OpenInference exporters attach a mark to an active parent span as a span event; this preserves the data, but it may not appear as a distinct node in visual trace trees.
- The result differs depending on the export format and downstream consumer, despite originating from the same Relay event.

## Expected behavior

Relay should provide a generic, documented representation for custom mark events that preserves:

- mark name and category/profile;
- payload and metadata;
- event timestamp;
- parent UUID and ancestry;
- ordering relative to surrounding scope, tool, and LLM events.

The representation should remain generic and must not depend on any particular plugin or mark name. It should also preserve the distinction between a point-in-time mark and a scoped tool or LLM operation; marks should not be silently converted into synthetic tool calls unless that is an explicit consumer projection.

## Possible directions

- Define a generic ATIF mark-step convention that downstream consumers can render directly.
- Add an exporter/renderer adapter mechanism for mark categories without hard-coding product-specific names in Relay core.
- Update the native ATIF import path or downstream integration guidance to preserve custom mark steps as visible events.
- Ensure OTLP/OpenInference consumers can inspect span events and retain the same parentage and ordering guarantees.

## Acceptance criteria

- Add a neutral fixture mark with structured data and a parent scope.
- Verify that ATOF, ATIF, OTLP, and OpenInference retain the mark name, payload, timestamp, parentage, and ordering.
- Verify that native ATIF import exposes the mark as a discoverable event/step rather than silently reducing it to opaque metadata.
- Verify that visual trace consumers can inspect the mark without requiring product-specific knowledge.
- Add regression coverage for marks emitted before and after parent scope completion, including the orphan-mark fallback behavior.
- Do not introduce private marker fields or product-specific special cases into the generic Relay API.

bbednarski9:
> [!WARNING]
> **BREAKING CHANGE:** **[Behavioral]** ATIF consumers will no
> longer receive synthetic `system` steps or populated `extra.event_payload`
> for Relay mark events. Use ATOF for the canonical mark stream or
> OTLP/OpenInference for trace-correlated mark telemetry. The Rust
> `AtifStepExtra.event_payload` field remains for source compatibility, but
> exporter-produced ATIF steps leave it unset.

## Summary

Adds an opt-in OTLP trace projection that represents Relay mark events as visible zero-duration child spans while preserving marks as canonical ATOF events.

The default remains exporter-native handling. Setting `mark_projection = "tool"` on an OpenTelemetry or OpenInference exporter makes eligible marks independently visible in trace-tree viewers while retaining their original identity, payload, category, profile, timestamp, and Relay ancestry.

ATIF remains step-oriented: it exports agent, LLM, tool-call, and observation steps, but does not fabricate ATIF tool calls or system steps for point-in-time marks. Use ATOF when the canonical mark stream is required.

Closes #381.

## Problem

Relay correctly records marks emitted by interceptors, including pending marks whose ordering is resolved by the managed LLM and tool execution paths. OTLP exporters normally attach a mark as a span event while its parent span is active. Marks emitted after a completed parent cannot be appended to that ended span, so the existing fallback is a zero-duration span that reuses the completed parent's trace context.

Trace-tree viewers do not consistently surface span events as independent nodes. A mark can therefore survive Relay and OTLP export but still appear to be missing in the viewer.

OpenTelemetry's log-based EventRecord model is a better semantic fit for independently exported point-in-time events, but Phoenix does not currently ingest OTLP Logs: [`POST /v1/logs` returns 405](https://github.com/Arize-ai/phoenix/issues/7212), and [OTel Logs support was previously tracked as an enhancement](https://github.com/Arize-ai/phoenix/issues/7355). Until trace-correlated EventRecords are accepted and rendered by common backends, an opt-in synthetic span remains the compatibility path that makes marks visible without changing runtime payloads or requiring plugins to emit fake lifecycle events.

### OTLP mark projection

- Adds the public `MarkProjection` enum with `inherit`, `event`, and `tool` values.
- Adds `mark_projection` independently to OpenTelemetry and OpenInference configuration.
- Adds configurable `mark_exclude_names`, defaulting to `llm.chunk`, so high-volume or backend-specific marks retain native event handling.
- Keeps `inherit` as the default for output compatibility.
- Keeps `event` as an explicit request for exporter-native event handling.
- In `tool` mode, emits eligible marks as zero-duration child spans named `mark:<mark name>`.
- Preserves mark UUID, parent UUID, payload, metadata, category, category profile, timestamp, and projection type as span attributes.
- OpenTelemetry emits an internal span with Relay tool-scope compatibility attributes.
- OpenInference additionally emits `openinference.span.kind = "TOOL"` for viewers that use OpenInference span kinds to render trace-tree nodes.

### Parentage and ordering

- Projection happens only in exporters; Relay lifecycle and callback ordering are unchanged.
- Marks emitted while the parent is active are parented to that span.
- Pending marks emitted after tool end reuse the completed tool span context, preserving trace ID and tool parentage while keeping tool-end-before-mark ordering.
- Pending request-intercept marks retain LLM parentage and LLM-start-before-mark ordering.
- Excluded marks use exporter-native handling rather than being dropped.

### ATIF and ATOF boundaries

- ATOF remains the canonical representation of every mark.
- ATIF omits marks because ATIF v1.7 models trajectory steps, tool calls, and observations rather than an independent point-in-time event type.
- Existing `AtifStepExtra.event_payload` remains available for Rust source compatibility, but exporter-produced steps leave it unset.
- No mark payload or private marker field is injected into in-flight requests, responses, or tool results.

### Bindings and documentation

- Exposes OTLP projection configuration through Rust, Python, Node.js, and Go helpers.
- Validates enum values and reports ATIF projection fields as unknown rather than silently accepting unsupported configuration.
- Documents exporter-native behavior, exclusions, synthetic-span compatibility, and the ATIF/ATOF format boundary.

## Configuration

```toml
[components.config.openinference]
enabled = true
endpoint = "http://localhost:6006/v1/traces"
mark_projection = "tool"
mark_exclude_names = ["llm.chunk"]
```

Existing configurations remain unchanged because `inherit` is the default.

## Validation

- Complete Rust workspace test suite passed, including native and worker dynamic-plugin integration tests.
- Strict workspace Clippy passed with warnings denied.
- `cargo fmt --all --check` and `git diff --check` passed.
- Focused observability suite: 276 passed.
- Managed-tool integration verifies:
  - tool end precedes its pending mark;
  - the projected mark is parented to the completed tool span;
  - ATIF contains no fabricated mark tool step;
  - mark category, category profile, metadata, and payload survive OTLP projection.
- OpenTelemetry and OpenInference tests cover active parents, completed parents, orphan fallback, exclusion by event name and hook alias, data-less marks, metadata fidelity, and `llm.chunk` preservation.
- Python observability tests: 10 passed.
- Node.js observability tests: 6 passed.
- Go binding suite passed across all packages.
- Go serialization coverage verifies that `nil` inherits the default mark exclusions while an explicit empty list disables them.
- Manual Phoenix validation confirmed that synthetic mark spans are visible in the trace tree with the expected parentage and attributes; the same environment accepted `/v1/traces` and returned 405 for `/v1/logs`.

## Compatibility rationale

The synthetic span is intentionally opt-in and exporter-local. It is not presented as the canonical OpenTelemetry representation of a point-in-time event. It is retained as a pragmatic visualization bridge until the target backend ecosystem can ingest and render trace-correlated OTLP EventRecords directly.


crates/core/src/observability/mod.rs
```diff
@@ -22,6 +23,34 @@ pub mod openinference;
...
 
+/// Export representation for point-in-time mark events.
+///
+/// Marks remain canonical ATOF events regardless of this setting. Exporters
+/// apply the selected projection only when translating those events into a
+/// downstream trajectory or trace format.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Serialize, Deserialize)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+#[serde(rename_all = "snake_case")]
+pub enum MarkProjection {
```
willkill07, line 34: Recommendation: add `Inherit` as a valid discrimator.

Reason: tool scopes under LLM calls is odd at best. We should avoid specialization based on marks in code. See suggestion below.

crates/core/src/observability/mod.rs
```diff
@@ -22,6 +23,34 @@ pub mod openinference;
...
+    /// Preserve marks as ATIF system steps or OTEL span events.
+    #[default]
+    Event,
+    /// Render non-streaming marks as deterministic ATIF tool steps or
+    /// zero-duration OTEL child spans so trace-tree consumers can display them
+    /// directly. High-volume `llm.chunk` marks remain events.
+    Tool,
+}
+
+pub(crate) fn is_llm_chunk_mark(event: &crate::api::event::Event) -> bool {
```
willkill07: Rather than hard-code logic, we should have a event name exclude filter be configuration driven, with things like `"llm.chunk"` being excluded by default.

crates/core/src/observability/openinference.rs
```diff
@@ -442,6 +484,10 @@ impl OpenInferenceEventProcessor {
     }
 
     fn process_mark(&mut self, event: &Event) {
+        if self.mark_projection == MarkProjection::Tool && !is_llm_chunk_mark(event) {
```
willkill07: Then logic becomes more generalized here...

crates/core/src/observability/atif.rs
```diff
@@ -2592,6 +2620,10 @@ impl StepConversionState {
             return;
         }
         self.flush_observations();
+        if self.mark_projection == MarkProjection::Tool {
```
willkill07: And logic becomes more generalized here...

willkill07 (changes requested):
I think we can improve this a bit with:
- exclusion filters over hard-coding
- `Inherit` option rather than always falling back to `Tool`

Unsure if an inclusion filter also makes sense

crates/core/src/observability/mod.rs
bbednarski9: Agree with this, i dont think we need an include because it will add some undue complexity in resolving the final set and this is really just to remove llm.chunk

crates/core/src/observability/mod.rs
bbednarski9: Implemented. `MarkProjection` now includes `inherit` and defaults to it for output-compatible exporter-native behavior. `event` explicitly forces event output, while `tool` projects eligible marks and lets excluded marks resolve to native handling.

crates/core/src/observability/mod.rs
bbednarski9: Added `mark_exclude_names`, defaulting to `llm.chunk`. Matching is configuration-driven and covers both literal event names and hook-event aliases in metadata, so exporter logic no longer needs a hard-coded exclusion list for tool projection.

crates/core/src/observability/openinference.rs
bbednarski9: Updated the branch to resolve projection through the configured mode and exclusion matcher: eligible marks use the tool projection, while excluded marks follow `inherit` and remain span events in OpenInference.

crates/core/src/observability/atif.rs
bbednarski9: Updated the branch to use the same generalized mode/exclusion resolution in ATIF: eligible marks become deterministic tool steps, excluded marks use `inherit` and remain native system-step output, with the existing `llm.chunk` suppression preserved for inherited handling.

bbednarski9:
configured mode = tool
        |
        +-- mark matches exclusion?
        |       yes → inherit
        |       no  → tool projection
        |
configured mode = event
        |
        +-- event representation
        |
configured mode = inherit
        |
        +-- exporter-native behavior

docs/observability-plugin/opentelemetry.mdx
```diff
@@ -51,6 +51,8 @@ NeMo Relay trace spans to the configured OTLP endpoint.
 | Field | Default | Notes |
 |---|---|---|
 | `enabled` | `false` | Must be `true` to construct and register the subscriber. |
+| `mark_projection` | `event` | `event` attaches marks to active parent spans. `tool` emits zero-duration child spans for trace-tree visibility. |
+| `mark_exclude_names` | `["llm.chunk"]` | Mark names excluded from `tool` projection; excluded marks retain native event handling. Metadata `hook_event_name` aliases are also matched. |
```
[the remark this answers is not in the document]
bbednarski9: Are you suggesting i run this locally? or should i trigger a docs build with a specific comment on this PR?

docs/configure-plugins/observability/opentelemetry.mdx
```diff
@@ -71,6 +73,20 @@ The collector should receive OTLP trace export requests. The tracing backend
...
+that representation. Set `mark_projection = "tool"` only when the target
+backend needs non-streaming marks represented as visible child spans. Projected
+marks use `SpanKind::Internal`, carry
+`nemo_relay.mark.projection = "tool"`, and retain mark UUID, parentage,
+category/profile, and payload attributes. High-volume `llm.chunk` marks retain
+exporter-native handling by default and when excluded from tool projection.
+
+Add other event names to `mark_exclude_names` when a backend should keep those
+marks in exporter-native form rather than visible tool children. The exclusion
+list affects only tool projection; it does not remove mark payload or metadata.
```
lvojtku: The first sentence isn't grammatically correct.

Maybe

```suggestion
Add other event names to `mark_exclude_names` for a backend to keep those
marks in exporter-native form rather than visible tool children. The exclusion
list affects only tool projection; it does not remove mark payload or metadata.
```

lvojtku (approved):
Approved with comments. Please address this comment.

willkill07 (approved):

PR #369, opened 2026-07-06
feat(plugin)!: Add Switchyard Decision API routing plugin

bbednarski9:
> [!WARNING]
> **BREAKING CHANGE:** **[Rust source compatibility]** This PR adds `FlowError::Upstream` and new fields to the public `AtofEndpointConfig`, `AtofEndpointSectionConfig`, and `EmitMarkEventParams` types. Downstream exhaustive `FlowError` matches must handle the new variant, and direct struct literals must initialize the new fields or migrate to the provided constructor/builder APIs. Builder-based mark emission and default Relay configurations remain compatible.

#### Overview

This adds Relay's first-party integration with the [Switchyard](https://github.com/NVIDIA-NeMo/Switchyard) Decision API. The plugin routes buffered and streaming LLM requests through an external Switchyard service, executes the selected Relay-owned backend, and records causal model-routing optimization evidence.

Switchyard owns provider-protocol translation through its `switchyard-translation` Rust library. Relay consumes that library only when the native Switchyard plugin is enabled; default Relay builds do not include either the plugin or the translation dependency.

- [x] I confirm this contribution is my own work, or I have the right to submit it under this project's license.
- [x] I searched existing issues and open pull requests, and this does not duplicate existing work.

##### Plugin and execution behavior

- Add the standalone `nemo-relay-switchyard` first-party plugin.
- Normalize stable Relay identity and build `switchyard.routing_request.v1` requests.
- Validate selected routes and optional counterfactual baselines against exact Relay-owned backend bindings.
- Support enforce and observe-only modes, six request-materialization modes, bounded retry-aware dispatch, protocol-specific trusted fallbacks, and no retry after the first upstream stream item.
- Record one model-routing optimization contribution only after the selected route commits. Failed decisions, failed provider attempts, and fallback dispatches do not claim routing savings.
- Preserve Switchyard routing metadata, including snapshot freshness, decision source, event count, and turn depth.
- Continue using Relay's existing ATOF exporter and lifecycle; the plugin does not install local routers, feature accumulators, or a duplicate event subscriber.

##### Translation ownership and optional dependency

Request, buffered-response, and SSE translation across OpenAI Chat Completions, OpenAI Responses, and Anthropic Messages use Switchyard's public `switchyard-translation` library instead of a Relay-local translation engine.

[...]

Issue #35, opened 2026-05-01
Testing Integration


mdemoret-nv:
This is the description

PR #389, opened 2026-07-08
feat: add extensible LLM optimization accounting

bbednarski9:
#### Overview

This adds a plugin-neutral Relay contract for recording arbitrary LLM optimizations, retaining their raw token effects, and calculating a combined baseline-versus-actual cost estimate when the managed LLM call closes.

#### Summary

- Add an open, string-backed `LlmOptimizationKind`, schema-tagged custom payloads, shared model transitions, and explicit baseline/effective/saved token counts.
- Extend request-intercept outcomes with ordered optimization contributions and normalized responses with a typed optimization summary.
- Add a per-call recorder for first-party execution intercepts without changing native ABI v1 host structs.
- Emit canonical `nemo_relay.llm.optimization` marks parented to the managed LLM handle while preserving plugin-specific marks.
- Calculate one combined baseline-versus-actual estimate at buffered completion, streaming finalization, or failure close.
- Export complete evidence to ATOF/ATIF and focused attributes to OTEL/OpenInference.
- Carry the additive JSON contract through worker/native JSON, FFI, Python, Node, and plugin-authoring surfaces.

#### Runtime behavior

The close path performs no network, file, refresh, subprocess, exporter-flush, or async work. It uses an `Arc` snapshot of the active in-memory pricing resolver and a short-held non-async mutex. Work is bounded to 64 attempted/accepted contributions, 16 KiB per complete serialized contribution envelope, and 256 KiB total retained contribution data per call. A breached bound seals accounting for that call, retains the accepted prefix, produces a typed partial summary, and never changes a provider result.

Token counts remain first-class evidence. Downstream systems can reprice stored summaries from baseline/effective models, observed usage, explicit token savings, and pricing provenance. Checked arithmetic exposes overflow, missing observed fields, and inconsistent totals as limitations rather than clamping or fabricating evidence. Missing usage, model identity, pricing, calculable cost totals, or currency agreement produces a partial summary with null unavailable cost fields.

Exactly one applied `model_routing` contribution may define the counterfactual baseline. Relay prefers that contribution's effective model because it is the exact target committed by the execution interceptor; provider response aliases and the requested model are fallbacks only when no applied routing target exists. Multiple applied routing contributions produce `multiple_routing_contributions` and are excluded from route arithmetic while all original evidence remains visible. Non-applied decisions remain visible but have no accounting authority. When there is no routing contribution, the terminal effective model is also the baseline so compression-only savings remain calculable.

Canonical optimization marks run through the mark-sanitizer chain captured with the managed LLM scope, including streaming close/drop paths. The summary embedded in the LLM end event independently runs through Relay's lifecycle sanitizer policy. Sanitizer or subscriber-dispatch failures remain best effort, retain unacknowledged mark state through the next lifecycle boundary, and never alter the LLM result.

1. `crates/types/src/codec/optimization.rs` — open contribution and summary wire contracts.
2. `crates/core/src/api/optimization.rs` — bounded recorder and close-time arithmetic.
3. `crates/core/src/api/llm.rs` and `crates/core/src/stream.rs` — buffered/streaming lifecycle integration.
4. `crates/core/src/observability/{atif,otel,openinference}.rs` — exporter projections.
5. `crates/core/tests/integration/middleware_tests.rs` — ordering, mark parentage, and end-summary integration.

#### Review guidelines

- Verify custom kinds and payload schemas round-trip without a Relay code change.
- Treat raw token counts as authoritative evidence and pricing as a versioned estimate.
- Check that applied routing evidence uses only the actually dispatched effective model.
- Confirm close-time work is bounded, in-memory, synchronous CPU work with asynchronous exporter delivery.
- Confirm missing data yields a partial summary rather than affecting provider behavior.
- This PR contains no Switchyard, Visor, or other plugin-specific policy or fixtures.

#### Testing summary

- Full `cargo test --workspace` after rebasing onto current `main`: passed, including 825 core library, 75 FFI unit, 65 FFI integration, 63 PII redaction, 57 middleware, 26 worker integration, 15 stream integration tests, and all doctests.
- Python: 499/499; Python plugin: 113/113 with 96.94% coverage; Python worker host E2E: 1/1.
- Node: 246/246; all Go packages passed.
- Full-envelope and attempt budgets, concurrent isolation, atomic close, sanitizer/queue failure, captured streaming sanitizer scope, interrupted streams, exactly-one routing authority, checked token arithmetic, incomplete costs, routed-model repricing, and provider-cost preservation.
- One canonical all-fields fixture round-trips through shared Rust types, native/worker SDKs, C FFI v2, Python native/worker, Node, and Go while preserving a custom kind and unknown top-level field.
- ATIF non-USD/no-optimization branches and matching OTEL/OpenInference per-cost currency and independent provenance, including complete non-USD summaries.
- A real cumulative Claude Code/InferenceHub StageRouter run produced five committed routing contributions, four Visor compression contributions, 1,868 explicit saved tokens, five complete summaries, zero provider errors/fallbacks, and exact ATOF/ATIF/OTEL parity.
- `cargo clippy --workspace --all-targets -- -D warnings`, `cargo deny check`, formatting, Ruff, `ty`, Go vet, Node formatting/docstrings, lockfile, protobuf, FFI-header, and non-attribution pre-commit checks passed.

#### Compatibility and scope

- Existing interceptors that omit `optimization_contributions` remain wire-compatible.
- Unknown optimization kinds and future top-level fields round-trip unchanged.
- Native ABI v1 structs are unchanged; the existing C helper is unchanged and an additive `_v2` helper accepts contribution JSON.
- Python native/worker, TypeScript/Node, Go, Rust worker/native, and C surfaces share the same canonical conformance fixture. The language models remain idiomatic and hand-maintained; schema generation is intentionally out of scope.
- Request contributions materialize after LLM start; execution contributions materialize at close with deterministic LLM parentage.
- Core arithmetic computes only the combined pipeline effect and never assigns monetary savings to individual plugins.

#### Review feedback addressed

- Moved new unit and wire tests out of `src/` following repository convention.
- Kept `LlmOptimizationKind` open and string-backed so third-party kinds do not require a Relay release; standard constants retain exact wire values.
- Bound the full contribution envelope and intercept aggregation, not only custom payload JSON.
- Routed canonical marks through captured-scope sanitizers and covered sanitizer/dispatcher failure behavior.
- Made model-pricing estimates follow the committed route while preserving provider-reported values.
- Added independently inspectable baseline/actual currency and provenance to OTEL/OpenInference.
- Completed Python response access, Python worker, Rust worker/native, Node, Go, and additive C FFI contribution surfaces.
- Added requested concurrency, ATIF, incomplete-cost, stream-lifecycle, token-overflow, and regression coverage.
- Made an omitted `applied` field fail-safe and consistent across Rust, Go, and Python; only explicit applied evidence can affect accounting.
- Preserved unknown future Python evidence-quality strings and prevented stale reserved fields in flattened `extra` data from reappearing on the wire.
- Timestamped execution contributions at acceptance so canonical routing marks sort strictly after their originating decision and no later than LLM end.

#### Related PRs

- Enables the accounting layer used by the Switchyard Decision API plugin in [#369](https://github.com/NVIDIA/NeMo-Relay/pull/369).
- [NVIDIA-NeMo/Switchyard #35](https://github.com/NVIDIA-NeMo/Switchyard/pull/35) supplies an explicit StageRouter counterfactual baseline.

## Summary by CodeRabbit

* **New Features**
  * Added managed LLM optimization evidence capture and close-time `optimization_summary` for buffered and streaming calls.
  * Emit namespaced optimization marks/events and include finalized optimization summaries in end-of-call responses.
  * Extended request-intercept contracts to carry ordered `optimizationContributions` across Node, Python, Go, and native/FFI; added `optimization_summary`/`optimization_contributions` accessors and OpenInference/OTEL/ATIF optimization attribute exports.

* **Bug Fixes**
  * Improved stream termination/interrupt accounting and event delivery queueing to avoid incorrect/missing optimization evidence.

* **Tests**
  * Expanded unit/integration coverage with new optimization fixtures and serialization/ordering/sanitization assertions.


crates/core/src/api/optimization.rs
```diff
@@ -0,0 +1,728 @@
...
+    let mut present = false;
+    let total = values.into_iter().flatten().fold(0_u64, |total, value| {
+        present = true;
+        total.saturating_add(value)
+    });
+    present.then_some(total)
+}
+
+#[cfg(test)]
+mod tests {
```
willkill07: should not be under `src/`

crates/types/src/codec/optimization.rs
```diff
@@ -0,0 +1,353 @@
...
+    pub estimated_cost_saved: Option<f64>,
+    /// Currency for `estimated_cost_saved`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub currency: Option<String>,
+    /// Ordered, bounded source evidence used by the calculation.
+    pub contributions: Vec<LlmOptimizationContribution>,
+}
+
+#[cfg(test)]
+mod tests {
```
willkill07: all tests should be out-of-tree from `src/`

crates/types/src/codec/optimization.rs
```diff
@@ -0,0 +1,353 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct LlmOptimizationKind(String);
+
+impl LlmOptimizationKind {
+    /// A request transformation that reduces input tokens.
+    pub const INPUT_COMPRESSION: &'static str = "input_compression";
+    /// A routing decision that changes the model serving a request.
+    pub const MODEL_ROUTING: &'static str = "model_routing";
+
```
willkill07, line 31: Why is this not an enum?

`strum` crate exists if you need enum discriminator => string conversion

crates/core/src/api/llm.rs
```diff
@@ -356,6 +365,42 @@ fn emit_pending_request_marks(
...
+                })
+                .build(),
+            Some(EventCategory::custom()),
+            Some(
+                CategoryProfile::builder()
+                    .subtype("nemo_relay.llm.optimization")
+                    .build(),
+            ),
+        ));
+        NemoRelayContextState::emit_event(&event, subscribers);
```
willkill07: **[P1] Should optimization marks be run through the mark-sanitizer chain??** This directly emits the contribution event, unlike pending marks and lifecycle events which call `sanitize_event`. Because contributions can contain arbitrary `payload`, `extra`, model identifiers, and producer strings, registered mark sanitizers are bypassed and sensitive data can reach subscribers. Apply `sanitize_event` (or the captured-stack variant for streaming) before `emit_event`, and add a redaction regression test.

crates/core/src/api/llm.rs
```diff
@@ -490,15 +535,15 @@ fn llm_call_end_with_behavior(
...
     let mut decode_error = None;
-    let annotated_response = match annotated_response {
-        Some(annotated_response) => Some(annotated_response),
+    let mut annotated_response = match annotated_response {
+        Some(annotated_response) => Some((*annotated_response).clone()),
         None => match (response_codec.as_ref(), data.as_ref()) {
             (Some(codec), Some(response)) => match codec.decode_response(response) {
                 Ok(mut decoded) => {
                     if behavior.attach_estimated_cost {
                         attach_estimated_cost_for_provider(&mut decoded, Some(&handle.name));
```
willkill07, line 600: **[P1] Do not retain a Relay estimate calculated for the response alias after routing.** This prices `response.model` before optimization finalization. If an applied routing contribution identifies a different authoritative effective model, `finalize_optimization_summary` sees `usage.cost` and preserves this stale estimate instead of pricing the dispatched model. Preserve provider-reported costs, but compute or replace Relay-generated costs only after the effective model is resolved. The existing direct finalizer test bypasses this lifecycle path.

crates/python/src/py_types/codecs.rs
```diff
@@ -539,6 +539,7 @@ impl PyAnnotatedLLMResponse {
                 tool_calls: optional_py_json(tool_calls, "tool_calls")?,
                 finish_reason: optional_finish_reason(finish_reason)?,
                 usage: optional_py_json(usage, "usage")?,
+                optimization_summary: None,
```
willkill07, line 543: **[P1] Expose the runtime-produced optimization summary to Python.** Relay later populates this field on annotated end responses, but `PyAnnotatedLLMResponse` has no `optimization_summary` getter, so Python subscribers cannot read the feature. Add a JSON-converting getter and update the checked-in `_native.pyi` declaration and tests.

crates/ffi/src/api/mod.rs
```diff
@@ -308,6 +308,7 @@ pub unsafe extern "C" fn nemo_relay_llm_request_intercept_outcome_json_new(
         request: unsafe { &*request }.0.clone(),
         annotated_request,
         pending_marks,
+        optimization_contributions: Vec::new(),
```
willkill07: **[P2] Provide an FFI construction path that preserves contributions.** The canonical outcome constructor always serializes an empty vector, so raw C callers using this helper cannot produce the new contract even though Python and Node can. Changing this function signature would break the existing ABI; add an ABI-compatible companion or versioned constructor that accepts optional contributions JSON while retaining this symbol.

crates/core/src/api/optimization.rs
```diff
@@ -0,0 +1,728 @@
...
+
+    let (estimated_cost_saved, currency) = match (&baseline_cost, &actual_cost) {
+        (Some(baseline), Some(actual))
+            if baseline.currency.eq_ignore_ascii_case(&actual.currency) =>
+        {
+            (
+                baseline
+                    .total_or_component_sum()
+                    .zip(actual.total_or_component_sum())
+                    .map(|(baseline, actual)| baseline - actual),
```
willkill07: **[P2] Mark the summary partial when a cost object has no calculable total.** With same-currency `CostEstimate` values whose total and components are all absent, this returns `None` without adding a limitation, and the summary can be reported as `complete`. Add a `missing_cost_total` limitation when either `total_or_component_sum()` is unavailable; also consider an explicit limitation when the effective model identity is missing.

willkill07 (changes requested):
(Codex Assisted)

Requesting changes for runtime safety, accounting correctness, and binding parity.

I confirmed the existing unresolved threads for the incomplete contribution-size bound and the missing per-cost currency/provenance attributes; those should remain merge blockers and are not duplicated below.

Two additional binding gaps are outside the changed-file diff and therefore cannot be attached inline:

- `python/plugin/src/nemo_relay_plugin/_api.py`: the slotted `LlmRequestInterceptOutcome` has no `optimization_contributions` field and `to_json()` never emits it, so Python gRPC worker plugins cannot use the new contract.
- `python/nemo_relay/_native.pyi`: `LLMRequestInterceptOutcome` still advertises the old constructor and omits the new property, so typed Python callers reject the runtime-supported API.

Please add regression coverage for mark sanitization, routed-model lifecycle pricing, Python worker round-tripping, non-USD/mixed-provenance export, concurrent call isolation, and the overflow/partial-summary branches.

crates/core/src/api/llm.rs
bbednarski9: Addressed in `968fdda1`. Provider-reported costs remain authoritative, but Relay `model_pricing` estimates are recomputed after resolving the single authoritative applied routing transition, using its exact effective model/provider. Buffered and streaming lifecycle regressions prove a provider response alias cannot survive as the priced model.

crates/core/src/api/optimization.rs
bbednarski9: Addressed in `968fdda1`. The recorder/finalizer tests now live in `crates/core/tests/unit/optimization_tests.rs`; `src/api/optimization.rs` retains only the repository-style `#[cfg(test)] #[path = ...]` module declaration. The rebased full Rust workspace passes.

crates/types/src/codec/optimization.rs
bbednarski9: I retained a string-backed newtype intentionally: third-party plugins must be able to introduce and round-trip unknown optimization kinds without waiting for a Relay release. A closed enum, including one using `strum`, would close that wire domain. `968fdda1` documents this rationale, while standard constants/constructors provide enum-like ergonomics and out-of-tree tests assert the exact `input_compression` and `model_routing` wire values.

crates/python/src/py_types/codecs.rs
bbednarski9: Addressed in `e095fa64`. `PyAnnotatedLLMResponse.optimization_summary` now returns the complete JSON summary, `_native.pyi` exposes the property, and PyO3 coverage exercises both populated and absent values. Python passes 499/499 tests.

crates/ffi/src/api/mod.rs
bbednarski9: Addressed in `e095fa64` without changing ABI v1. The existing helper is preserved and delegates to additive `nemo_relay_llm_request_intercept_outcome_json_new_v2`, which accepts optional canonical contribution JSON. FFI tests cover malformed input and lossless all-fields/unknown-field round trips.

crates/core/src/api/llm.rs
bbednarski9: Addressed in `968fdda1`. Canonical optimization marks now use `sanitize_event_with_scope_stack` against the managed LLM captured scope for buffered and streaming close/drop paths. Sanitizer unavailability and dispatcher rejection do not advance the mark cursor. Regressions cover nested payload/extra redaction, captured streaming scope, unavailable sanitizers, and separate scope-end summary sanitization. The branch is also rebased onto the PII mark-sanitization work in `c83c8f27`, and its full workspace tests pass.

python/plugin/src/nemo_relay_plugin/_api.py
```diff
@@ -200,13 +233,277 @@ def to_json(self) -> dict[str, Json]:
...
+    producer: str
+    kind: str
+    applied: bool = True
+    id: str | None = None
+    sequence: int | None = None
+    model_transition: LlmOptimizationModelTransition | None = None
+    token_impact: LlmOptimizationTokenImpact | None = None
+    payload_schema: LlmOptimizationDataSchema | None = None
+    payload: Json | None = None
+    extra: dict[str, Json] = field(default_factory=dict)
```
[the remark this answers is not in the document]
bbednarski9: Addressed in `aca78de4`. Direct Python authoring and `from_json` now both treat omitted `applied` as `true`, preserving the established constructor behavior. The regression verifies both serialized forms; 107 worker-SDK tests pass.

crates/core/tests/integration/pipeline_tests.rs
```diff
@@ -1574,6 +1657,149 @@ async fn test_stream_response_codec_populates_annotated_response() {
...
+            .usage
+            .as_ref()
+            .and_then(|usage| usage.cost.as_ref())
+            .map(|cost| cost.source.clone()),
+        Some(CostSource::ProviderReported)
+    );
+
+    deregister_subscriber("routed_alias_pricing_sub").unwrap();
+    reset_active_pricing_resolver().unwrap();
+}
```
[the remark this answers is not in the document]
bbednarski9: Addressed in `aca78de4`. The pricing resolver is now reset by a `Drop` guard in every test that installs global pricing state, so panic or early return cannot leak the resolver into later tests. The routed-pricing integration suite passes.

go/nemo_relay/optimization.go
```diff
@@ -0,0 +1,172 @@
...
+		Kind:            known.Kind,
+		Applied:         known.Applied,
+		ModelTransition: known.ModelTransition,
+		TokenImpact:     known.TokenImpact,
+		PayloadSchema:   known.PayloadSchema,
+		Payload:         known.Payload,
+		Extra:           extra,
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
bbednarski9: Addressed in `aca78de4`. Go marshal and unmarshal now reject a non-null payload without `payload_schema`; explicit JSON null remains equivalent to absent payload. Focused Go tests and `go vet ./...` pass.

crates/core/tests/integration/stream_tests.rs
```diff
@@ -255,10 +264,164 @@ async fn test_stream_wrapper_drop_emits_end_event_for_partial_stream() {
...
+                && event.parent_uuid() == Some(drop_after_uuid)
+        })
+        .expect("route-commit contribution mark should survive an interrupted stream");
+    assert_eq!(
+        committed_mark.data().unwrap()["producer"],
+        "test.route_commit"
+    );
+
+    deregister_subscriber("stream_termination_accounting").unwrap();
+}
```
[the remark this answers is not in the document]
bbednarski9: Reviewed on `aca78de4`. These stream tests exercise different termination conditions and assertions; consolidating their setup would be a test-only refactor with no coverage or behavior change, so I kept the explicit cases.

crates/worker/tests/optimization_contract.rs
```diff
@@ -0,0 +1,51 @@
...
+    .unwrap();
+    let contribution: LlmOptimizationContribution =
+        serde_json::from_value(fixture.clone()).unwrap();
+
+    assert_ne!(contribution.kind, LlmOptimizationKind::input_compression());
+    assert_ne!(contribution.kind, LlmOptimizationKind::model_routing());
+    assert_eq!(
+        contribution.extra["future_top_level_field"],
+        fixture["future_top_level_field"]
+    );
```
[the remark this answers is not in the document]
bbednarski9: Addressed in `aca78de4`. The worker conformance test now asserts the exact custom kind `custom_energy_optimization`, in addition to its full semantic round trip.

crates/node/src/api/mod.rs
```diff
@@ -2263,7 +2263,7 @@ pub fn register_llm_request_intercept(
...
+        ts_arg_type = "(args: { name: string; request: Json; annotated: Json | null }) => { request: Json; annotated?: Json | null; pendingMarks?: Array<{ name: string; category?: string | null; categoryProfile?: Json; data?: Json; metadata?: Json }>; optimizationContributions?: Array<{ id?: string; sequence?: number; producer: string; kind: string; applied: boolean; model_transition?: { baseline?: { model: string; provider?: string }; effective?: { model: string; provider?: string } }; token_impact?: { baseline?: { prompt_tokens?: number; completion_tokens?: number; cache_read_tokens?: number; cache_write_tokens?: number; total_tokens?: number }; effective?: { prompt_tokens?: number; completion_tokens?: number; cache_read_tokens?: number; cache_write_tokens?: number; total_tokens?: number }; saved?: { prompt_tokens?: number; completion_tokens?: number; cache_read_tokens?: number; cache_write_tokens?: number; total_tokens?: number }; quality?: 'observed' | 'estimated'; estimation_method?: string }; payload_schema?: { name: string; version: string }; payload?: Json; [key: string]: Json | undefined }> }"
```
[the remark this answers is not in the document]
bbednarski9: Addressed in `aca78de4`. All three generated Node signatures now expose `kind: 'input_compression' | 'model_routing' | (string & {})`, preserving standard autocomplete and arbitrary custom kinds. A generated-declaration regression asserts all three occurrences.

crates/python/src/py_types/codecs.rs
```diff
@@ -617,6 +618,22 @@ impl PyAnnotatedLLMResponse {
...
+                    pyo3::exceptions::PyValueError::new_err(format!(
+                        "optimization summary serialization error: {error}"
+                    ))
+                })?;
+                json_to_py(py, &value)
+            }
+            None => Ok(py.None()),
+        }
+    }
+
```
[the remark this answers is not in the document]
bbednarski9: Addressed in `aca78de4`. `optimization_summary` now uses the shared JSON conversion/error hook, including a forced serialization-failure regression. The Python/PyO3 test suite passes.

crates/core/src/observability/otel.rs
```diff
@@ -707,9 +707,157 @@ fn end_attributes(event: &Event) -> Vec<KeyValue> {
...
+                .as_ref()
+                .and_then(|cost| cost.pricing_as_of.as_ref())
+        })
+    {
+        attributes.push(KeyValue::new(
+            "nemo_relay.llm.optimization.pricing_as_of",
+            as_of.clone(),
+        ));
+    }
+}
```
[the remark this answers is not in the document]
bbednarski9: Reviewed on `aca78de4`. The baseline and actual branches intentionally remain explicit because their omission/provenance rules are independently tested; this is duplication-only with no behavioral gap, so I left it as optional cleanup outside this feature revision.

python/plugin/src/nemo_relay_plugin/_api.py
```diff
@@ -200,13 +233,287 @@ def to_json(self) -> dict[str, Json]:
...
+            "applied",
+            "model_transition",
+            "token_impact",
+            "payload_schema",
+            "payload",
+        }
+        transition = value.get("model_transition")
+        impact = value.get("token_impact")
+        schema = value.get("payload_schema")
+        applied = value.get("applied", True)
```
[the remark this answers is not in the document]
bbednarski9: Fixed in 232d66cf. The wire default is now consistently fail-safe: omitted applied decodes as false in Python, matching existing Rust serde and Go zero-value behavior, so incomplete evidence cannot affect accounting. Direct Python construction uses the same default. Dedicated Rust, Go, and Python regressions decode the same omitted-field payload and assert non-applied behavior; the focused suites, Go vet, and formatting checks pass.

willkill07 (approved):

PR #391, opened 2026-07-08
feat: emit automatic skill-load marks [1/3]

willkill07:
#### Overview

Implements the shared RELAY-443 runtime behavior for automatic skill-load marks. This is the first commit in a three-PR cumulative stack; all three PRs target upstream `main` as requested.

#### Details

- Detects first-class skill tools, complete structured `SKILL.md` reads, and safe standalone shell readers in the Rust tool lifecycle.
- Emits eager, tool-parented `skill.load` marks with minimal payloads, per-call deduplication, and failure-path preservation.
- Normalizes Claude Code, Codex, and Hermes hook behavior, including inferred slash-command expansion marks.
- Adds comprehensive positive and rejected-branch coverage outside the `src` tree.
- Preserves current upstream observability semantics: ATOF is canonical for marks and ATIF omits point-in-time marks.
- Contains no documentation changes.

Validation: `just test-rust`, `just test-python`, `just test-go`, `just test-node`, `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `uv run pre-commit run --all-files`.

Breaking changes: none.

Start with `crates/core/src/api/skill_load.rs`, then review its lifecycle integration in `crates/core/src/api/tool.rs` and the external test matrix in `crates/core/tests/unit/skill_load_tests.rs`.

- Relates to: RELAY-443

## Summary by CodeRabbit

* **New Features**
  * Added skill-load (`skill.load`) mark emission for tool calls, including detected skill names and load sources.
  * Added support for Claude Code `UserPromptExpansion`, including inferred “skill load” marks and stable event metadata.
  * Preserved skill-load marks through sanitization and tool execution failures.
* **Bug Fixes**
  * Improved deduplication and idempotent hook handling; prevented extra skill-load marks when handling is already owned.
* **Documentation**
  * Updated Claude Code integration docs to include `UserPromptExpansion` and clarified the minimum supported version.
  * Enhanced diagnostics from the CLI doctor with a version-bound warning for hook compatibility.

### DIFF
--- a/crates/cli/src/installer.rs
+++ b/crates/cli/src/installer.rs
@@ -18,6 +18,7 @@ use crate::error::CliError;
 const HOOK_EVENTS: &[&str] = &[
     "SessionStart",
     "UserPromptSubmit",
+    "UserPromptExpansion",
     "PreToolUse",
     "PostToolUse",
     "PostToolUseFailure",
--- a/crates/cli/src/installer.rs
+++ b/crates/cli/src/installer.rs
@@ -10,11 +10,14 @@ use serde_json::{Value, json};
 use crate::config::{CodingAgent, GatewayMode, HookForwardCommand};
 use crate::error::CliError;
 
-// Claude Code's hook loader strictly whitelists event names — any unknown event causes the
-// entire hooks file to be rejected (no hooks register). Only events present in Claude Code's
-// whitelist as of 2.1.x belong here. Codex 0.129 has a smaller subset (SessionStart,
-// UserPromptSubmit, PreToolUse, PostToolUse, Stop, PreCompact, PostCompact, PermissionRequest)
-// and silently ignores events it doesn't recognize, so the union list is safe for both agents.
+// Claude Code validates plugin hooks.json against a strict event-name whitelist — one unknown
+// event rejects the entire plugin's hooks (no hooks register, silently). Both Claude vectors
+// (the transparent-run temp plugin and the marketplace plugin) are plugin hooks.json, so every
+// event here must exist in the oldest supported Claude Code. UserPromptExpansion sets that
+// floor: 2.1.116 (verified empirically; 2.1.114 rejects it — see `claude_hook_floor_warning`
+// in doctor.rs). Codex 0.129 has a smaller subset (SessionStart, UserPromptSubmit, PreToolUse,
+// PostToolUse, Stop, PreCompact, PostCompact, PermissionRequest) and silently ignores events
+// it doesn't recognize, so the union list is safe for both agents.
 const HOOK_EVENTS: &[&str] = &[
     "SessionStart",
     "UserPromptSubmit",
--- a/crates/core/src/api/skill_load.rs
+++ b/crates/core/src/api/skill_load.rs
@@ -1,0 +1,374 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Private detection helpers for eager skill-load observability marks.
+
+use std::collections::HashSet;
+
+use serde_json::Value;
+
+pub(crate) const HANDLED_METADATA_KEY: &str = "nemo_relay.skill_load_handled";
+pub(crate) const PRECOMPUTED_METADATA_KEY: &str = "nemo_relay.skill_loads";
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum SkillLoadSource {
+    SkillTool,
+    StructuredRead,
+    ShellRead,
+}
+
+impl SkillLoadSource {
+    pub(crate) const fn as_str(self) -> &'static str {
+        match self {
+            Self::SkillTool => "skill_tool",
+            Self::StructuredRead => "structured_read",
+            Self::ShellRead => "shell_read",
+        }
+    }
+
+    fn from_str(value: &str) -> Option<Self> {
+        match value {
+            "skill_tool" => Some(Self::SkillTool),
+            "structured_read" => Some(Self::StructuredRead),
+            "shell_read" => Some(Self::ShellRead),
+            _ => None,
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct SkillLoad {
+    pub(crate) name: String,
+    pub(crate) source: SkillLoadSource,
+}
+
+pub(crate) fn detect(tool_name: &str, args: &Value) -> Vec<SkillLoad> {
+    let normalized_tool = normalize_identifier(tool_name);
+    let source_and_names = if matches!(normalized_tool.as_str(), "skill" | "skillview") {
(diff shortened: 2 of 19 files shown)


crates/cli/src/installer.rs
```diff
@@ -18,6 +18,7 @@ use crate::error::CliError;
 const HOOK_EVENTS: &[&str] = &[
     "SessionStart",
     "UserPromptSubmit",
+    "UserPromptExpansion",
```
zhongxuanwang-nv, line 24: A potential nitty thing is: the comment at lines 13-17 says this list must only contain events in Claude Code's whitelist 'as of 2.1.x' because one unknown name makes Claude Code reject the whole hooks file. So in this case, an older Claude Code builds this addition may silently disable all relay hooks?

I added a follow-up PR based on this branch :)

zhongxuanwang-nv: https://github.com/willkill07/NeMo-Relay/pull/4

crates/core/src/api/tool.rs
```diff
@@ -250,11 +265,37 @@ fn tool_call_with_subscriber_snapshot(
...
-        (handle, event)
+        let marks = skill_loads
+            .into_iter()
+            .map(|skill_load| {
+                state.create_event(MarkEvent::new(
+                    BaseEvent::builder()
+                        .name("skill.load")
+                        .parent_uuid(handle.uuid)
+                        .timestamp(handle.started_at)
+                        .data(json!({"skill_name": skill_load.name}))
```
zhongxuanwang-nv, line 276: A nitty thing and probably also expected; on here and on line 245, detection uses pre-redacted args, and the emitted mark only passes the mark-sanitizer chain. Since PII redactor doesn't register mark sanitizer, here we may accidentally expose the SKILL.md's parent-dir name or raw skill arguments via the mark? Probably no remediations because adding a mark sanitizer would be adding extra work on a hot path?

zhongxuanwang-nv (approved):
Nitty things, thanks Will!

crates/core/src/api/skill_load.rs
```diff
@@ -0,0 +1,374 @@
...
+fn is_structured_reader(tool_name: &str) -> bool {
+    [
+        "read",
+        "readfile",
+        "readtextfile",
+        "readmultiplefiles",
+        "fileread",
+    ]
+    .iter()
+    .any(|reader| tool_name == *reader || tool_name.ends_with(reader))
```
mnajafian-nv: Could we tighten this matcher before merge? After `normalize_identifier()`, the `ends_with(...)` check can classify unrelated tool names like `thread` as structured readers. That would emit false `skill.load` marks if the args contain a `SKILL.md` path. I think this wants exact-name or separator-boundary matching, plus a regression test.

crates/cli/src/installer.rs
willkill07: I'm not sure how many people are actually on older versions. They usually gate new model access behind updates. Is this a real concern?

If deep backwards compatibility is truly gating, I claim it should be part of a follow up PR.

crates/core/src/api/skill_load.rs
willkill07: I'm not aware of any built-in tool called thread. I understand defensive programming, but this seems like over-engineering.

The reality is if there is a built-in tool which has an argument of a SKILL.md file, then it's probably reading it.

The endswith is intentional as a tool ending with `read` is most likely a file reading tool.

If/when this changes, I'll gladly eat my foot.

crates/core/src/api/tool.rs
willkill07: Skill load source is not a path.

It's an enum that states the type of tool source (explicit read, shell command, etc)

zhongxuanwang-nv: Oops, I should've investigated deeper into this!

crates/cli/src/installer.rs
zhongxuanwang-nv: I think if users downloaded claude code through their native installer, claude auto updates itself. However, Claude may not auto-update through other ways like homebrew. I think most users probably have the newer version; I mean I guess we don't have many alternatives besides warning them in `nemo-relay doctor` and documentations? 🤔

crates/core/tests/unit/atif_tests.rs
```diff
@@ -2780,6 +2780,36 @@ fn test_exporter_ignores_marks_with_hook_metadata() {
...
+        }))
+        .build();
+
+    {
+        let mut state = exporter.state.lock().unwrap();
+        state.events.push(tool_start);
+        state.events.push(mark);
+    }
+
+    let trajectory = exporter.export().unwrap();
```
mnajafian-nv, line 2809: Nice addition. This test still looks weaker than the contract the PR claims: with only a tool start plus a mark, `trajectory.steps.is_empty()` passes even if mark filtering is broken, because no tool end exists to form a step. Could we make this a full tool start / skill.load mark / tool end flow and then assert we get exactly one tool step with no leaked `skill.load` data or metadata?

crates/core/src/api/tool.rs
```diff
@@ -229,12 +230,26 @@ fn tool_call_with_subscriber_snapshot(
...
+        .and_then(Json::as_object)
+        .and_then(|metadata| metadata.get(skill_load::HANDLED_METADATA_KEY))
+        .and_then(Json::as_bool)
+        .is_some_and(|handled| handled);
+    let skill_loads = if handled_skill_loads {
+        Vec::new()
+    } else if let Some(skill_loads) = skill_load::precomputed(params.metadata.as_ref()) {
+        skill_loads
+    } else {
+        skill_load::detect(params.name, &params.args)
```
mnajafian-nv, line 245: One contract question here: detection now runs on raw `params.args`, while the emitted tool start event uses `sanitized_args`. That means a request sanitizer can redact the source path from the tool event but Relay will still derive and emit `skill_name` from the original args, and the new integration test locks that in. If that is intentional, could we document it as the observability contract? Otherwise I think detection should run on the same sanitized projection we expose to subscribers.

willkill07: You have to feed in the unsanitized input because a sanitizer could theoretically rename/redact all arguments to tool calls. It would be impossible to track or emit skill usage in that case. Downstream mark sanitizers can further redact the emitted skill marks.

There is no documentation change necessary because sanitizers still have control over data included in the emitted mark

crates/core/tests/unit/atif_tests.rs
willkill07: This is not weaker. This is ATIF specific behavior.

Let's also think about the preconditions established. If you don't have a tool call end, do you really have a complete trajectory? No.

I claim that since you don't have a complete trajectory, any inconsistencies are fine. If you think there's a way to reasonably get the behavior you'd expect, I'm all ears. However, that's out of scope for this PR anyways since it's a generic change that would need to be applied to the entire ATIF exporter functionality.

Therefore, it's implementation defined behavior and not a downstream reproducible contract needing to be advertised.

mnajafian-nv (approved):

PR #392, opened 2026-07-08
feat: add skill-load coverage for agent integrations [2/3]

willkill07:
#### Overview

Extends RELAY-443 coverage into the supported agent integrations. This is the second commit in a three-PR cumulative stack; all three PRs target upstream `main` as requested, so this PR temporarily includes the first commit until it merges.

#### Details

- Adds skill-load hook coverage for the Claude Code and Codex coding-agent integrations.
- Adds real LangChain, LangGraph, and Deep Agents coverage while preserving the existing Deep Agents configured-skills mark.
- Keeps all test bodies outside implementation `src` trees.
- Removes all OpenClaw changes; that integration work will be refactored separately.

Validation: `just test-rust`, `just test-python`, `just test-node`, and `uv run pre-commit run --all-files`.

Breaking changes: none.

Start with `crates/core/src/api/skill_load.rs`, then review the coding-agent hook configurations and Python integration coverage.

- Relates to: RELAY-443

## Summary by CodeRabbit

* **Tests**
  * Added integration coverage confirming `skill.load` marks are emitted when structured skills are read via supported middleware and tooling.
  * Verified `skill.load` event payloads include `skill_name` plus attribution metadata (`skill_load_source`, `tool_name`) and that the `parent_uuid` links back to the associated `read_file` start scope.
  * Extended coverage across DeepAgents middleware, LangChain middleware, LangGraph execution, and direct tool usage.


integrations/openclaw/src/hook-replay/tool.ts
```diff
@@ -63,8 +115,28 @@ export function replayAfterToolCall(
...
+      : event.error
+        ? { ...toolDisplayPayload(event, false), error: errorToJson(event.error), result: event.result ?? null }
+        : { ...toolDisplayPayload(event, false), result: event.result ?? null },
+  );
+  if (eagerToolCall) {
+    manager.emitCapturedUnderSession('after_tool_call', session, () => {
+      manager.nf.toolCallEnd(eagerToolCall.handle, endPayload, null, metadata, endMicros);
+      manager.state.counters.toolSpansReplayed += 1;
+    });
+    return;
```
mnajafian-nv: Could we confirm the intended behavior here for blocked skill reads? Returning early on the eager path seems to bypass `blockedToolDetails()`, so an eager `read_file` call that comes back blocked would close the tool span without emitting the existing `openclaw.tool_blocked` mark. If that’s intentional, could we lock it in with a test? Otherwise it seems like blocked handling may need to happen before this return.

integrations/openclaw/src/hook-replay/skill-load.ts
```diff
@@ -0,0 +1,222 @@
...
+    return [];
+  }
+
+  return [...new Set(names)].map((skillName) => ({ skillName, source }));
+}
+
+function isStructuredReader(toolName: string): boolean {
+  return ['read', 'readfile', 'readtextfile', 'readmultiplefiles', 'fileread'].some(
+    (reader) => toolName === reader || toolName.endsWith(reader),
+  );
```
mnajafian-nv: I think this matcher needs a narrower boundary here. After normalization, the `endsWith(...)` check can classify unrelated tool names like `thread` as structured readers. That would emit false `skill.load` marks if params contain a `SKILL.md` path. Exact-name or separator-boundary matching here, plus a regression test in `skill-load.test.ts`, would make this contract safer.

mnajafian-nv:
I’d drop the OpenClaw-specific skill-load handling from this stack. With the CLI/sidecar pivot, extending hook replay with eager/fallback state looks like unnecessary complexity. I’d keep the shared core and other integration changes, remove the corresponding OpenClaw changes from 392/393, and handle managed OpenClaw skill marks at the sidecar boundary if we still need them. That would also make my OpenClaw-specific inline comments moot.

mnajafian-nv (approved):
LGTM!

PR #398, opened 2026-07-09
feat(events)!: cull stale annotated LLM histories

willkill07:
#### Overview

> [!WARNING]
> **BREAKING CHANGE (Behavioral):** LLM start request event inputs and annotations can omit earlier turns after the owning agent's freshness has been consumed. The request used for provider execution remains unchanged.

Implement RELAY-445 with per-agent freshness. A new agent session, and the first LLM start after compaction, retains the complete sanitized request history. Later starts cull the annotation and codec-backed event input to the current user turn.

#### Details

- Track freshness for the implicit root and nested agent scopes independently, and remove explicit-agent state when its scope closes. Non-agent scopes inherit their nearest agent's freshness instead of creating a separate budget.
- Atomically consume freshness at LLM start. Concurrent starts need not have a stable emission order, but exactly one consumes a fresh state.
- Mark the owning agent fresh on canonical `compaction` events. Repeated marks are idempotent, and both `PreCompact` and `PostCompact` normalize to the shared canonical event name.
- For stale agents, retain system instructions, the latest user message, and every following assistant or tool message. Handle empty, instruction-only, and no-user histories without cloning an annotation when projection is a no-op.
- When a request codec is active, encode the projected annotation into the event-only `LlmRequest` so `event.input` and `category_profile.annotated_request` contain the same bounded history. If that event-only encode fails, emit a diagnostic, preserve the complete event input and annotation, and continue the provider call.
- Leave response payloads, middleware/provider requests, caller-visible values, and existing public API signatures unchanged. This implementation has no history baselines, ordering queues, or response projection.
- Keep ordinary mark events on the scope-stack read-lock path; acquire the write lock only for canonical compaction marks.
- Add real-codec coverage for managed and streaming calls, including fail-open encode errors; projection edge cases and no-op allocation behavior; implicit-root and nested-agent inheritance; concurrent starts; and repeated compaction marks.
- Move documentation and consumer-skill updates to the documentation-only follow-up PR #406.

Validation:

- `cargo fmt --all`
- `just test-rust` — 833 core unit tests plus integration tests passed
- `just test-python` — 511 passed
- `just test-node` — 254 passed
- `just test-go`

Start with `crates/core/src/api/runtime/scope_stack.rs` for the root/nested-agent freshness ownership, `crates/core/src/api/llm.rs` for fail-open event-only request projection, and `crates/core/tests/unit/llm_api_tests.rs` for the behavioral contract.

- Fixes: RELAY-445

### DIFF
--- a/crates/core/src/api/llm.rs
+++ b/crates/core/src/api/llm.rs
@@ -29,7 +29,7 @@ use crate::api::shared::{
     resolve_parent_uuid, run_request_intercepts_with_codec_and_recorder,
     sanitize_event_with_scope_stack, snapshot_event_subscribers,
 };
-use crate::codec::request::AnnotatedLlmRequest;
+use crate::codec::request::{AnnotatedLlmRequest, Message};
 use crate::codec::response::{AnnotatedLlmResponse, attach_estimated_cost_for_provider};
 use crate::codec::traits::{LlmCodec, LlmResponseCodec};
 use crate::error::{FlowError, Result};
@@ -298,6 +298,89 @@ fn create_llm_handle(params: CreateLlmHandleParams<'_>) -> Result<LlmHandle> {
     Ok(state.create_llm_handle(params))
 }
 
+fn request_turn_projection_needed<T>(
+    items: &[T],
+    is_user: &impl Fn(&T) -> bool,
+    is_instruction: &impl Fn(&T) -> bool,
+) -> bool {
+    let Some(last_index) = items.len().checked_sub(1) else {
+        return false;
+    };
+    match items.iter().rposition(is_user) {
+        Some(start) => items[..start].iter().any(|item| !is_instruction(item)),
+        None => items
+            .iter()
+            .enumerate()
+            .any(|(index, item)| index != last_index && !is_instruction(item)),
+    }
+}
+
+fn retain_current_request_turn<T>(
+    items: &mut Vec<T>,
+    is_user: impl Fn(&T) -> bool,
+    is_instruction: impl Fn(&T) -> bool,
+) -> bool {
+    if !request_turn_projection_needed(items, &is_user, &is_instruction) {
+        return false;
+    }
+    let last_index = items.len() - 1;
+    let Some(start) = items.iter().rposition(is_user) else {
+        let mut index = 0;
+        items.retain(|item| {
+            let retain = index == last_index || is_instruction(item);
+            index += 1;
+            retain
+        });
+        return true;
+    };
+    let mut current_turn = items.split_off(start);
+    items.retain(is_instruction);
+    items.append(&mut current_turn);
+    true
+}
+
+fn project_llm_request_to_current_user_turn(
+    request: &mut LlmRequest,
+    annotated_request: &mut Option<Arc<AnnotatedLlmRequest>>,
+    request_codec: Option<&dyn LlmCodec>,
+) {
+    let Some(annotation) = annotated_request.as_mut() else {
+        return;
+    };
+    if !request_turn_projection_needed(
+        &annotation.messages,
+        &|message| matches!(message, Message::User { .. }),
+        &|message| matches!(message, Message::System { .. }),
+    ) {
+        return;
+    }
+    let original_annotation = request_codec.map(|_| Arc::clone(annotation));
+    let projected = limit_annotated_request_history_to_current_user_turn(Arc::make_mut(annotation));
+    debug_assert!(projected);
+    if let Some(codec) = request_codec {
+        match codec.encode(annotation, request) {
+            Ok(encoded) => *request = encoded,
+            Err(error) => {
+                eprintln!(
+                    "nemo_relay: LLM request projection encode failed; preserving full event history: {error}"
+                );
+                *annotation = original_annotation
+                    .expect("codec-backed projection should preserve the original annotation")
+            }
+        }
+    }
+}
+
+fn limit_annotated_request_history_to_current_user_turn(
+    annotated_request: &mut AnnotatedLlmRequest,
+) -> bool {
+    retain_current_request_turn(
+        &mut annotated_request.messages,
+        |message| matches!(message, Message::User { .. }),
+        |message| matches!(message, Message::System { .. }),
+    )
+}
+
 fn emit_llm_start(
     handle: &LlmHandle,
     request: &LlmRequest,
@@ -339,9 +422,9 @@ fn emit_llm_start_with_subscribers(
             .map_err(|error| FlowError::Internal(error.to_string()))?;
         state.llm_sanitize_request_entries(&scope_locals)
     };
-    let sanitized_request =
+    let mut sanitized_request =
         NemoRelayContextState::llm_sanitize_request_snapshot_chain(request.clone(), &entries);
-    let annotated_request = match request_codec {
+    let mut annotated_request = match request_codec {
         Some(codec)
             if sanitized_request.headers != request.headers
                 || sanitized_request.content != request.content =>
@@ -350,6 +433,18 @@ fn emit_llm_start_with_subscribers(
         }
         _ => annotated_request,
     };
+    let scope_stack = handle.captured_scope_stack();
+    let agent_is_fresh = {
+        let mut scope_guard = scope_stack.write().expect("scope stack lock poisoned");
+        scope_guard.take_agent_freshness(handle.parent_uuid)
+    };
+    if !agent_is_fresh {
+        project_llm_request_to_current_user_turn(
+            &mut sanitized_request,
+            &mut annotated_request,
+            request_codec,
+        );
+    }
     let input = serde_json::to_value(&sanitized_request).unwrap_or(Json::Null);
     let event = {
         let context = global_context();
@@ -358,7 +453,7 @@ fn emit_llm_start_with_subscribers(
             .map_err(|error| FlowError::Internal(error.to_string()))?;
         state.build_llm_start_event(handle, Some(input), annotated_request)
     };
-    if let Some(event) = sanitize_event_with_scope_stack(event, handle.captured_scope_stack()) {
+    if let Some(event) = sanitize_event_with_scope_stack(event, scope_stack) {
         NemoRelayContextState::emit_event(&event, subscribers);
     }
     Ok(())
@@ -486,7 +581,10 @@ fn emit_optimization_marks_with(
 ///
 /// # Notes
 /// Sanitize-request guardrails affect only the emitted start-event payload, not
-/// the caller-owned [`LlmRequest`].
+/// the caller-owned [`LlmRequest`]. When the owning agent is not fresh, the
+/// emitted request annotation is limited to the current user turn. Managed
+/// calls with a request codec also apply that projection to the event input,
+/// without changing the request used for provider execution.
 pub fn llm_call(params: LlmCallParams<'_>) -> Result<LlmHandle> {
     let handle_params = CreateLlmHandleParams::builder()
         .name(params.name)
--- a/crates/core/src/api/runtime/scope_stack.rs
+++ b/crates/core/src/api/runtime/scope_stack.rs
@@ -9,6 +9,7 @@
 //! context into worker threads.
 
 use std::cell::RefCell;
+use std::collections::{HashMap, HashSet};
 use std::sync::{Arc, RwLock};
 
 use uuid::Uuid;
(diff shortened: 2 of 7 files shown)


crates/core/src/api/llm.rs
```diff
@@ -626,6 +728,10 @@ fn llm_call_end_with_behavior(
             ..AnnotatedLlmResponse::default()
         });
     }
+    if let Some(data) = data.as_mut() {
+        limit_llm_response_to_current_user_turn(data);
```
mnajafian-nv: Could we enforce this contract at a shared LLM end-event construction boundary? This projection covers `llm_call_end_with_behavior`, but `LlmStreamWrapper::emit_end_event` and `emit_llm_end_without_output` construct end events without calling either history limiter. A streaming finalizer, annotated streaming response, or failed call whose `handle.data` contains `messages` can therefore emit complete history even though the PR promises that every end event is bounded.

Would it be safer to introduce one observability projection helper used by manual, managed, streaming, and failure completion paths? A contract test applying the same multi-turn response to all four paths would keep them from diverging again.

crates/core/src/api/llm.rs
```diff
@@ -298,6 +299,93 @@ fn create_llm_handle(params: CreateLlmHandleParams<'_>) -> Result<LlmHandle> {
...
+        limit_json_history_to_current_user_turn(input);
+    }
+}
+
+fn limit_llm_response_to_current_user_turn(response: &mut Json) {
+    let Some(response) = response.as_object_mut() else {
+        return;
+    };
+    if let Some(messages) = response.get_mut("messages") {
+        limit_json_history_to_current_user_turn(messages);
```
mnajafian-nv: What invariant establishes that every top-level `messages` array is a conversation history? LLM responses and `data` are otherwise opaque JSON, so a custom integration can legitimately use this key for a non-conversation collection. If none of those objects has `role: user`, the current helper silently retains only the final item.

Could we require a recognized provider shape, codec annotation, or role-bearing message structure before applying the no-user fallback? Unknown shapes should remain unchanged, with a regression covering a non-role `messages` array.

crates/core/src/api/llm.rs
```diff
@@ -350,6 +438,15 @@ fn emit_llm_start_with_subscribers(
         }
         _ => annotated_request,
     };
+    let preserve_full_context = {
+        let scope_stack = handle.captured_scope_stack();
+        let mut scope_guard = scope_stack.write().expect("scope stack lock poisoned");
+        scope_guard.take_full_llm_context(handle.parent_uuid)
```
mnajafian-nv: Does this guarantee that the first subscriber-visible LLM start is the baseline when a propagated scope stack is used concurrently? The pending flag is removed under the scope lock, but the lock is released before projection, event sanitization, construction, and dispatcher enqueue. Two threads sharing the stack can therefore let the call that consumed the baseline enqueue after a bounded call.

The PR description calls this atomic consumption at LLM start emission, but only the reservation is atomic. Should baseline selection be coupled to an ordered emission sequence, or should each event carry an explicit `full` versus `current_turn` representation so consumers do not depend on inferred ordering? A concurrent-start regression using the supported cross-thread propagation path would prove the chosen contract.

crates/core/src/api/runtime/scope_stack.rs
```diff
@@ -26,7 +27,8 @@ use crate::registry::{RegistryEntry, SortedRegistry};
 /// close.
 pub struct ScopeStack {
     stack: Vec<ScopeHandle>,
-    scope_registries: std::collections::HashMap<Uuid, ScopeLocalRegistries>,
+    scope_registries: HashMap<Uuid, ScopeLocalRegistries>,
+    full_llm_context_pending: HashSet<Uuid>,
```
mnajafian-nv: What identity is the baseline intended to describe: an agent execution or an individual conversation history? One agent scope can contain independent planner, classifier, evaluator, and primary-model histories, but the first call consumes the agent's only baseline.

If one baseline per agent is intentional, could the contract state explicitly that complete context is not guaranteed for every LLM history and that exact reconstruction is out of scope? If reconstruction is expected, this state needs a stable history identity rather than only the owning agent UUID.

mnajafian-nv (commented):
I left four focused questions on contract consistency, concurrency, and compatibility. The post-sanitization projection direction looks sound; these comments are intended to make the breaking event contract consistent across every path before approval.

crates/core/src/api/runtime/scope_stack.rs
willkill07: The documentation says ATOF is not a per-event complete archive, but it also doesn't state what is in-scope or out of scope.

I'd rather *NOT* stipulate any formal contract aiming for reconstruction. But it's still possible even with this change.

crates/core/src/api/llm.rs
willkill07: addressing in follow-up commit

crates/core/src/api/llm.rs
willkill07: Resolving with the built-in codec patterns

crates/core/src/api/llm.rs
willkill07: Addressing

crates/core/src/api/llm.rs
```diff
@@ -358,8 +590,8 @@ fn emit_llm_start_with_subscribers(
             .map_err(|error| FlowError::Internal(error.to_string()))?;
         state.build_llm_start_event(handle, Some(input), annotated_request)
     };
-    if let Some(event) = sanitize_event_with_scope_stack(event, handle.captured_scope_stack()) {
-        NemoRelayContextState::emit_event(&event, subscribers);
+    if let Some(event) = sanitize_event_with_scope_stack(event, &scope_stack) {
+        emission_reservation.commit(event, subscribers, scope_stack);
```
mnajafian-nv: Could we keep request-local marks inside the same per-agent ordering boundary as the reserved start event?

The reservation serializes the start itself, but the later `emit_pending_request_marks` and `emit_optimization_marks` calls still use the normal async subscriber path. A later request can therefore queue its marks before an earlier baseline start has actually been dispatched, so a consumer may observe request-local marks before the start event that is supposed to anchor them.

Would it be safer to route those marks through the same ordered emission state, or otherwise delay them until the reserved start has been dispatched? A concurrency regression with two requests that both emit pending marks would make the intended ordering contract explicit.

crates/core/src/api/llm.rs
```diff
@@ -298,6 +301,222 @@ fn create_llm_handle(params: CreateLlmHandleParams<'_>) -> Result<LlmHandle> {
...
+        return limit_role_history_to_current_user_turn(messages).unwrap_or(false);
+    }
+    false
+}
+
+pub(crate) fn limit_llm_end_observation_to_current_user_turn(
+    data: &mut Option<Json>,
+    annotated_response: &mut Option<Arc<AnnotatedLlmResponse>>,
+) {
+    let source_annotation = annotated_response.as_deref().cloned().or_else(|| {
```
mnajafian-nv: Should this helper fall back to raw `data["messages"]` when an `annotated_response` is present but does not carry response history?

Right now it prefers the supplied annotation and returns early when `extra.messages` is absent. That means a summary-only or metadata-only annotation can leave `data.messages` unprojected, even though `build_llm_end_event` is now the shared contract point for every end path.

Would it be safer to use the annotation path only when it contains history, and otherwise continue with the raw `data` path? A regression where `annotated_response` contains only `optimization_summary` and `data.messages` is multi-turn would pin this down.

willkill07: Done

crates/core/src/api/llm.rs
willkill07: Done

crates/core/src/api/llm.rs
```diff
@@ -291,11 +339,244 @@ pub struct LlmCallEndParams<'a> {
...
+    items: &mut Vec<T>,
+    is_user: impl Fn(&T) -> bool,
+    is_instruction: impl Fn(&T) -> bool,
+) -> bool {
+    let Some(last_index) = items.len().checked_sub(1) else {
+        return false;
+    };
+    let Some(start) = items.iter().rposition(is_user) else {
+        let changed = last_index > 0;
+        items.drain(..last_index);
```
mnajafian-nv: Could we split the no-user fallback by history kind here? For request projection, a system/developer-only conversation is still the complete current context, but this branch trims it to the final instruction by draining everything before `last_index`. That seems to contradict the new start-event contract to preserve system/developer instructions. A regression with `messages: [system, developer]` and a Responses `input` that contains only instruction items would make the intended behavior explicit.

crates/core/src/api/llm.rs
```diff
@@ -291,11 +339,292 @@ pub struct LlmCallEndParams<'a> {
...
+    match value {
+        Json::Array(items) => items
+            .iter_mut()
+            .any(|item| replace_first_matching_history(item, original_history, projected_history)),
+        Json::Object(fields) => fields.values_mut().any(|field| {
+            replace_first_matching_history(field, original_history, projected_history)
+        }),
+        _ => false,
+    }
+}
```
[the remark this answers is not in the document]
mnajafian-nv: Agree this looks worth tightening. Since we extract history from a specific field, patching by deep equality could truncate the wrong copy if the same array appears elsewhere in the payload. Could we replace at the exact extracted path, or constrain this to known history keys?

mnajafian-nv (approved):
Great work! Conditional approval pending addressing the last comment!

crates/core/src/api/llm.rs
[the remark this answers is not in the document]
mnajafian-nv: @willkill07 ^^^

crates/core/src/api/runtime/scope_stack.rs
```diff
@@ -40,9 +42,11 @@ impl ScopeStack {
             .name("root")
             .scope_type(ScopeType::Agent)
             .build();
+        let root_uuid = root.uuid;
         Self {
             stack: vec![root],
-            scope_registries: std::collections::HashMap::new(),
+            scope_registries: HashMap::new(),
+            fresh_agents: HashSet::from([root_uuid]),
```
mnajafian-nv, line 51: Is it intentional for the implicit root to participate in freshness here? With the current behavior, two unrelated request or custom scopes under the same thread-local root share one freshness budget, so the second request can emit a bounded annotation even when the caller isolates work per request. If the contract is specifically per explicit agent session, it may be worth either excluding the implicit root here or locking the intended behavior in with a doc note and regression test.

docs/instrument-applications/instrument-llm-call.mdx
```diff
@@ -238,7 +238,10 @@ Native subscriber delivery is asynchronous. Flush subscribers before validating
 printed output. In Node.js, also wait one event-loop tick after
 `flushSubscribers()` so JavaScript callbacks can run.
 - LLM start input contains the request after request intercepts and
-  sanitize-request guardrails.
+  sanitize-request guardrails. The raw provider payload remains unchanged.
+- Annotated requests retain complete history while their agent session is fresh
```
mnajafian-nv: This looks good overall. One clarification that may help application users: freshness here is tied to agent scopes and compaction, not to request or custom scopes in general. A short note to that effect would make the runtime contract easier to follow.

crates/core/src/api/runtime/scope_stack.rs
willkill07: good catch. it should be included

mnajafian-nv (approved):
Great work, I love the simplification... Conditional approval upon reviewing inline suggestions

crates/core/tests/unit/llm_api_tests.rs
```diff
@@ -42,6 +51,580 @@ fn request() -> LlmRequest {
...
+    let events = Arc::new(Mutex::new(Vec::<Event>::new()));
+    let captured = events.clone();
+    register_subscriber(
+        "managed-streaming-freshness",
+        Arc::new(move |event| captured.lock().unwrap().push(event.clone())),
+    )
+    .unwrap();
+
+    let request = multi_turn_request();
+    let codec: Arc<dyn LlmCodec> = Arc::new(OpenAIChatCodec);
```
mnajafian-nv, line 247: Nice coverage on the Chat path. Please can you add one regression using a second built-in request codec, ideally OpenAI Responses, that proves the same contract there too: fresh starts keep full `event.input` and full annotation, stale starts bound both `event.input` and `annotated_request`, and provider execution still receives the original unbounded request. That would make the generalized `request codec` claim much easier to trust and would guard against shape-specific drift in `input` / `instructions` handling.

willkill07: I don't see the need to test another codec. that is orthogonal to what is umder test.

mnajafian-nv (approved):

mnajafian-nv (approved):
LGTM!

mnajafian-nv (approved):

PR #396, opened 2026-07-09
feat(codec): add a provider codec factory and route built-in consumers through it

zhongxuanwang-nv:
#### Overview

Add a single provider-codec factory in `nemo_relay::codec::resolve` and route the previously hand-rolled surface/name → codec dispatch across the built-in consumers through it, so the set of built-in provider codecs and their canonical names live in one place instead of being re-encoded per crate (several used `_ => None/Err` arms that silently failed to gain a new provider).

- [x] I confirm this contribution is my own work, or I have the right to submit it under this project's license.
- [x] I searched existing issues and open pull requests, and this does not duplicate existing work.

#### Details

Factory in `crates/core/src/codec/resolve.rs`, hung off `ProviderSurfaceDescriptor` next to each codec:
- `ProviderSurface::codec_name` / `from_codec_name` — canonical name ↔ surface.
- `request_codec` / `response_codec` / `streaming_codec` — construct the built-in `LlmCodec` / `LlmResponseCodec` / `StreamingCodec` (returned as `Box`; the codecs are stateless, so callers that share them convert with `Arc::from` at their own boundary).
- `supported_codec_names` — accepted names for config validation.

Consumers routed through the factory (all internal — private fns / `pub(crate)` enums — so not breaking):
- CLI gateway `codecs_for_route` via new `ProviderRoute::provider_surface`.
- Adaptive ACG `decode_request_for_surface` via new `RequestSurface::provider_surface` (error labels are now strum-derived from the surface variant names, per review).
- NeMo Guardrails `resolve_codec` / `LocalGuardrailsCodec` (enum retained for provider-specific streaming-text extraction).
- PII redaction: removed `BuiltinRequestResponseCodec` + `instantiate_builtin_codec`; split the codec into request/response codecs; `BuiltinCodecName` retained for response overlay.
- NeMo Guardrails and PII config validators now check `supported_codec_names()` instead of a hardcoded list.

Additive `pub` / `pub(crate)` only; no serialized, wire, config, event, or cache-key schema change.

Follow-ups (out of scope here): fully single-source the JSON-schema/editor display lists (currently order-fixed consts) with drift-guard tests; adopt the factory in the opt-in response-cache feature once it lands.

#### Where should the reviewer start?

[...]

PR #404, opened 2026-07-10
feat: add opt-in exact-match caching for LLM responses

zhongxuanwang-nv:
#### Overview

First implementation stage of a five-stage response-cache series (exact match → logical keys → tool-result cache → semantic → streaming semantic), with the standalone documentation in #405. This stage is independently reviewable.

Adds an opt-in, exact-match response cache for managed LLM calls as a `response_cache` section of the Adaptive plugin config. Repeated eligible calls can be served from an in-memory or Redis store instead of re-running the provider. The feature is off until the section is present; by default, only explicitly deterministic requests (`temperature = 0`) are cacheable. Runtime backend failures fail open to a live call, while invalid configuration is rejected.

#### Details

- **Keying** (`crates/adaptive/src/response_cache/key.rs`): keys use a SHA-256 fingerprint of the RFC 8785-canonicalized, normalized request. Provider surface detection and guarded raw-body fallback avoid incorrect reuse when a decode would be lossy. Provider, namespace, and schema version are folded into the key; only allowlisted headers participate.
- **Eligibility and storage** (`crates/adaptive/src/response_cache/intercept.rs`): incomplete, error, stateful, unparseable, or otherwise unsafe calls bypass or remain unstored. Nondeterministic requests bypass unless `cache_nondeterministic = true` is set explicitly.
- **Streaming**: live misses tee provider-native chunks while assembling a replayable aggregate. Only naturally completed, faithfully replayable streams are stored. Early close and cleanup errors remain idempotent, close upstream exactly once, and prevent cache writes.
- **Store** (`crates/adaptive/src/response_cache/store.rs`): bounded in-memory storage and feature-gated Redis storage with expiry and operation deadlines. Redis initialization/storage failures disable or bypass the optional cache rather than blocking managed calls.
- **Gateway integration** (`crates/cli/src/gateway`): upstream failures remain verbatim for the client but are not cacheable; post-success runtime rejections still surface as errors. Buffered and streaming cache hits use the correct content type.
- **Config and bindings**: Rust, Python, Node.js, and Go expose the same safe default (`cache_nondeterministic = false`) and validation behavior. `nemo-relay doctor` reports configuration and backend reachability.
- **Observability**: `response_cache` hit/miss/bypass marks include fingerprints, reasons, and savings without request or response bodies.
- **Review follow-up**: restored the close-aware streaming regressions dropped by an earlier branch replacement, added subscriber-visible nondeterministic bypass assertions, added Redis initialization fail-open coverage, and strengthened the benchmark determinism assertion.

Validation completed on the final branch head:

- `RUST_TEST_THREADS=1 just test-rust`
- `just test-python` — 542 passed
- `just test-node` — 286 passed
- `just test-go`
- `uv run pre-commit run --all-files` — all hooks pass on the final tree

1. `crates/adaptive/src/response_cache/key.rs` — key derivation and guarded normalization.
2. `crates/adaptive/tests/integration/response_cache_tests.rs` — end-to-end eligibility, replay, and stream-lifecycle behavior.
3. `crates/cli/src/gateway` — the upstream-failure and cached-response boundary.

- Relates to #396 — keying, replay, and savings reporting route through the provider codec factory.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in exact-match LLM response caching (buffered + streaming) with TTL, namespace/priority, bypass controls, header allowlisting, and in-memory/Redis backends.
  * Extended configuration and editor/schema support across Rust, Go, JavaScript/TypeScript, and Python, including cache telemetry (hit/miss/bypass) and savings metadata.
  * Added CLI response-cache health/reachability checks with safe fail-open when initialization can’t complete.
* **Bug Fixes**
  * Improved gateway replay by capturing upstream body bytes, enforcing strict JSON parsing rules, and preserving correct headers and SSE content type.
* **Tests**
  * Expanded unit/integration/coverage/benchmark coverage for keying, replay safety, marks, gateway behavior, and storage backends.


crates/adaptive/src/runtime/validation.rs
```diff
@@ -81,9 +83,157 @@ pub fn validate_config(config: &AdaptiveConfig) -> ConfigReport {
...
+            // A shared redis cache with no namespace mixes every caller's
+            // responses together — caution against cross-tenant reuse.
+            if config.namespace.is_empty() {
+                report.diagnostics.push(response_cache_warning(
+                    "response_cache.shared_empty_namespace",
+                    Some("namespace"),
+                    "redis backend with an empty namespace shares cached responses across all \
+                     callers; set a namespace per environment/tenant"
+                        .to_string(),
+                ));
```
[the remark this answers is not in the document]
zhongxuanwang-nv: The empty-namespace default is deliberate. `namespace` defaults to `""`, and a shared Redis cache within a single trust domain is this backend's primary use case, so a hard error would reject the minimal valid `kind = "redis"` config. Tenant isolation is layered elsewhere: auth headers are already a hard validation error if allowlisted into keys, and this warning plus the configuration docs direct multi-tenant deployments to set a per-tenant namespace. Keeping the warning as the middle ground.

crates/adaptive/tests/integration/response_cache_tests.rs
```diff
@@ -0,0 +1,1134 @@
...
+
+    call(&provider, stateful()).await;
+    call(&provider, stateful()).await;
+
+    assert_eq!(
+        calls.load(Ordering::SeqCst),
+        2,
+        "stateful Responses calls (store=true) must never be cached"
+    );
+}
```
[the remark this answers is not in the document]
zhongxuanwang-nv: This rule is already pinned at two layers. The key decision-table test (`nondeterministic_calls_bypass_only_when_disabled`) pins that nonzero `temperature` classifies as `KeyOutcome::Bypass("nondeterministic_temperature")` unless `cache_nondeterministic` opts in, and `stateful_responses_calls_bypass_the_cache` pins the shared `KeyOutcome::Bypass` → bypass-mark → live-call plumbing end to end (all bypass reasons flow through that single path). A second integration test through the same plumbing with a different reason string would duplicate existing coverage, which this suite deliberately avoids.

crates/adaptive/tests/integration/response_cache_tests.rs
```diff
@@ -0,0 +1,1134 @@
...
+    let provider = counting_provider(Arc::clone(&calls), sample_body());
+    call(&provider, chat_request("co-configured")).await;
+    call(&provider, chat_request("co-configured")).await;
+
+    assert_eq!(
+        calls.load(Ordering::SeqCst),
+        1,
+        "the cache must still hit when co-configured with the ACG execution intercept"
+    );
+}
```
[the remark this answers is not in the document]
zhongxuanwang-nv: There is no clean observable to assert at this layer: the default ACG provider is `passthrough` (no request transformation to detect on the miss), and ACG's execution events (`acg_debug::emit`) are env-gated stderr diagnostics, not subscriber-visible lifecycle events. Verifying priority ordering end to end would need ACG configured with a rewriting provider plugin, which belongs to the ACG suite rather than the cache suite. This test pins what it can here: co-registering both execution intercepts does not break cache hits — a registration conflict, double-keying, or intercept error would fail it.

crates/adaptive/src/plugin_component.rs
```diff
@@ -284,10 +285,75 @@ fn validate_adaptive_plugin_config(plugin_config: &Map<String, Json>) -> Vec<Con
...
+        _ => return,
+    };
+    validate_unknown_fields(
+        diagnostics,
+        policy,
+        Some(format!("response_cache.backend.{backend_kind}")),
+        backend_config,
+        known_fields,
+    );
+}
```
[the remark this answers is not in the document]
zhongxuanwang-nv: Fixed in 813e7119. A mistyped `url` was already rejected by the `missing_redis_url` check (`as_str()` returns `None` for non-strings), so the gap was `key_prefix` only: validation now errors on a non-string `key_prefix` instead of letting `build_redis_store` silently substitute the shared default prefix, mirroring the existing `max_bytes` mistype check.

crates/adaptive/src/response_cache/intercept.rs
```diff
@@ -0,0 +1,456 @@
...
+                }
+            }
+            // Forward to the consumer; a send error means it was dropped.
+            if tx.send(item).await.is_err() {
+                return;
+            }
+        }
+        // Store only protocol-complete streams: every collector finalizes a
+        // clean truncation as a well-formed partial.
+        if !collector_failed && saw_terminal {
```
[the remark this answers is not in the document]
zhongxuanwang-nv: Fixed in 813e7119. The tee now tracks per-choice completion (`StreamCompletion`): a chat stream is terminal only when every choice index that appeared carries a `finish_reason` (Anthropic `message_stop` / Responses `response.completed` unchanged). Added a failing-first test where an n=2 stream closes cleanly with choice 1 unfinished and verifies the aggregate is not stored.

crates/adaptive/src/response_cache/key.rs
```diff
@@ -0,0 +1,989 @@
...
+            "max_tokens": 64,
+            "messages": [{"role": "user", "content": "Answer"}]
+        }));
+        assert_ne!(
+            key_of("anthropic", &malformed, &config),
+            key_of("anthropic", &clean, &config),
+            "a null-text system block must not key like an absent system"
+        );
+    }
+}
```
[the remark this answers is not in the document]
zhongxuanwang-nv: Fixed in 813e7119. Moved to `tests/unit/response_cache/{key,mark,replay,store}_tests.rs` via the crate's `#[cfg(test)] #[path = ...]` include convention (which keeps private-item access) — including store.rs, which had the same issue.

crates/adaptive/src/response_cache/replay.rs
```diff
@@ -0,0 +1,368 @@
...
+    for chunk in synthesize_replay_chunks(aggregate).unwrap_or_default() {
+        if collect(chunk).is_err() {
+            return true;
+        }
+    }
+    let mut reassembled = codec.finalizer()();
+    let mut expected = aggregate.clone();
+    strip_stream_metadata(&mut reassembled);
+    strip_stream_metadata(&mut expected);
+    reassembled != expected
```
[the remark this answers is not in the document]
zhongxuanwang-nv: Fixed in 813e7119. `replay_is_lossy` now returns `true` for an unrecognized shape, so the streaming tier misses and runs live; buffered hits still serve the entry. The single-frame fallback in `replay_aggregate` remains as a defensive last resort behind the gate.

crates/adaptive/src/runtime/validation.rs
```diff
@@ -81,9 +83,157 @@ pub fn validate_config(config: &AdaptiveConfig) -> ConfigReport {
...
+                .get("url")
+                .and_then(Json::as_str)
+                .is_none()
+            {
+                report.diagnostics.push(response_cache_error(
+                    "response_cache.missing_redis_url",
+                    Some("backend.config.url"),
+                    "redis backend requires backend.config.url".to_string(),
+                ));
+            }
```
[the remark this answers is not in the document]
zhongxuanwang-nv: Fixed in 813e7119. A trimmed-empty `url` now reports the same `missing_redis_url` diagnostic instead of failing later at `redis::Client::open`.

crates/adaptive/tests/integration/response_cache_benchmark_tests.rs
```diff
@@ -0,0 +1,367 @@
...
+        calls.load(Ordering::SeqCst),
+        total_requests - repeats,
+        "the provider must run exactly total_requests - hits times"
+    );
+    // Supporting invariants make a failure point at the cause, not just a total.
+    assert_eq!(
+        stats.hits + stats.misses,
+        total_requests,
+        "every request is either a hit or a miss (no bypasses here)"
+    );
```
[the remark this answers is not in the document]
zhongxuanwang-nv: Fixed in 813e7119. The repeat test now asserts `bypasses == 0` and the bypass test asserts `misses == 0`, closing the partition (each suite already asserted the other two statuses sum to `total_requests`).

crates/adaptive/tests/integration/response_cache_tests.rs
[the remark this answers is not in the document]
zhongxuanwang-nv: Addressed in d7205ced. `nondeterministic_calls_emit_bypass_marks_and_run_live` now repeats an identical `temperature = 0.7` request with nondeterministic caching disabled, asserts two provider invocations, and asserts exactly two `response_cache` bypass marks with reason `nondeterministic_temperature`.

crates/adaptive/tests/integration/response_cache_tests.rs
```diff
@@ -0,0 +1,1344 @@
...
+    stream_call(&provider, chat_request("natural completion")).await;
+    stream_call(&provider, chat_request("natural completion")).await;
+
+    assert_eq!(calls.load(Ordering::SeqCst), 1);
+    assert_eq!(
+        close_calls.load(Ordering::SeqCst),
+        1,
+        "the live upstream must be closed exactly once before the entry is reusable"
+    );
+}
```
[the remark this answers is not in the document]
zhongxuanwang-nv: Addressed in d7205ced. The close-aware regressions were restored, and `naturally_completed_stream_closes_upstream_before_cache_commit` now asserts the upstream close count immediately after the first fully drained stream, before issuing the second request. The second call then proves the entry is reusable without another upstream invocation.

crates/adaptive/src/response_cache/store.rs
```diff
@@ -0,0 +1,501 @@
...
+    Err(AdaptiveError::InvalidConfig(
+        "response_cache: backend.kind = \"redis\" requires building with the 'redis-backend' \
+         feature"
+            .to_string(),
+    ))
+}
+
+#[cfg(test)]
+#[path = "../../tests/unit/response_cache/store_tests.rs"]
+mod tests;
```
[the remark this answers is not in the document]
zhongxuanwang-nv: Validation is now complete on final head d7205ced: `RUST_TEST_THREADS=1 just test-rust`, `cargo clippy --workspace --all-targets -- -D warnings`, `just test-python` (542 passed), `just test-node` (286 passed), `just test-go`, and `uv run pre-commit run --all-files` all pass. The matching docs branch also passes `just docs` with zero errors (only the unauthenticated redirect-check warning).

crates/adaptive/src/response_cache/key.rs
```diff
@@ -0,0 +1,426 @@
...
+        }
+        for key in &config.skip_keys {
+            object.remove(key);
+        }
+        normalize_tool_call_ids(object);
+    }
+
+    let headers = allowlisted_headers(&request.headers, &config.header_allowlist);
+
+    let key_doc = json!({
```
willkill07, line 119: This key is shared by every request with the same provider and body.  The default namespace is empty, and normal authorization headers cannot be included.

1. Is this intentional?
2. Is this expected to be shared across tenants?

In a shared gateway, two tenants could therefore receive the same cached answer. Could we add a safe tenant or upstream partition to the key, or reject an unscoped shared cache?

crates/adaptive/src/response_cache/replay.rs
```diff
@@ -0,0 +1,268 @@
...
+/// (the collector keeps the last snapshot wholesale, so reassembly is exact).
+fn synthesize_responses_chunks(aggregate: &Json) -> Vec<Json> {
+    let mut created = Map::new();
+    for key in ["id", "object", "model"] {
+        if let Some(value) = aggregate.get(key) {
+            created.insert(key.to_string(), value.clone());
+        }
+    }
+    created.insert("status".to_string(), json!("in_progress"));
+    let mut chunks = vec![json!({"type": "response.created", "response": created})];
```
willkill07: The cached OpenAI Responses events do not include `sequence_number`.

Live Responses streams include it, and strict clients can use it to validate event order, so a cached stream may be rejected even though the live stream works.

Could we add increasing sequence numbers to every replayed event and cover the wire shape in a test?

crates/adaptive/src/redis.rs
```diff
@@ -31,6 +31,57 @@ use crate::trie::serialization::TrieEnvelope;
...
+            event = "plugin_resource_access_failed",
+            plugin_kind = "adaptive",
+            resource_kind = "redis",
+            permission = "connect",
+            reason = "client_configuration";
+            "Plugin resource access validation failed"
+        );
+        AdaptiveError::Storage(format!("redis client: {e}"))
+    })?;
+    let conn = client.get_connection_manager().await.map_err(|e| {
```
willkill07, line 62: This first Redis connection has no timeout. If Redis is unreachable but the network silently drops packets, plugin startup and doctor can wait for the full TCP timeout instead of continuing without the optional cache.

Could we wrap `get_connection_manager()` in the same kind of short timeout used for cache operations, then return a storage error so the cache can fail open?

go/nemo_relay/adaptive.go
```diff
@@ -66,6 +67,45 @@ type AcgConfig struct {
...
 }
 
+// ResponseCacheConfig configures the opt-in LLM response cache: a section
+// of the adaptive config (a sibling to acg/adaptive_hints/tool_parallelism), not a
+// standalone plugin kind. The Rust core validates and installs it from the adaptive
+// runtime; this struct only has to carry the section through to the FFI validator.
+type ResponseCacheConfig struct {
+	// TTLSeconds is how long a stored answer stays reusable, in seconds (> 0).
+	// Always serialized: with omitempty an explicit 0 would vanish and Rust's
+	// default (3600) would silently replace it instead of failing validation.
```
willkill07: A normal Go partial config such as `ResponseCacheConfig{Namespace: "dev"}` still sends `ttl_seconds: 0` because this field is always serialized.

Rust then rejects the config instead of using the one-hour default.

Could we omit fields that were not supplied, or add custom JSON marshaling, so partial Go configs behave like the Rust, Python, and Node APIs?

crates/adaptive/src/response_cache/key.rs
```diff
@@ -0,0 +1,426 @@
...
+    let (mut body, effective_codec) = resolved_body(provider, request);
+
+    if let Some(object) = body.as_object_mut() {
+        // `AnnotatedLlmRequest.extra` is `#[serde(flatten)]`, so provider fields
+        // also land at top level — one skip pass covers both.
+        for key in DEFAULT_SKIP_KEYS {
+            object.remove(*key);
+        }
+        for key in &config.skip_keys {
+            object.remove(key);
```
willkill07: `skip_keys` can remove any top-level field before the request is hashed. The validation list protects only a few fields, so an operator can accidentally skip `response_format`, `seed`, `n`, or generation controls. Then different requests can reuse the same cached answer, such as returning plain text when JSON was requested.

Could we allow only known-safe fields to be skipped, or reject every field that can change the response?

willkill07 (changes requested):
In addition to addressing the Coderabbit feedback, I found a few issues.

crates/adaptive/src/response_cache/key.rs
zhongxuanwang-nv: It's true! Thanks Will for catching this! For now I'm just removing or rejecting arbitrary skip_keys explicitly! :)

go/nemo_relay/adaptive.go
zhongxuanwang-nv: Thanks Will for catching this! I would make only these two Go fields presence-aware:
```
TTLSeconds *uint64 `json:"ttl_seconds,omitempty"`
Priority   *int32  `json:"priority,omitempty"`
```

And preserve constructor defaults

```
func NewResponseCacheConfig() ResponseCacheConfig {
    ttlSeconds := uint64(3600)
    priority := int32(50)

    return ResponseCacheConfig{
        TTLSeconds:  &ttlSeconds,
        Priority:    &priority,
        KeyStrategy: "exact_request",
    }
}
```

crates/adaptive/src/redis.rs
zhongxuanwang-nv: Oh yeah hahaha I remember I also gave comments like these in other PR reviews — this is totally valid! i think currently Redis1.2.0 bounds each attempt and has finite retries, but the total retry can still take up to 20s. I'll do the thing you mentioned here! :)

crates/adaptive/src/response_cache/replay.rs
zhongxuanwang-nv: Thanks Will! Fixed this! :)

crates/adaptive/src/response_cache/key.rs
zhongxuanwang-nv: It's totally valid! Now blank or whitespace-only namespaces now produce response_cache.missing_namespace as a hard error for every backend :)

mnajafian-nv (approved):
LGTM! great work!

crates/adaptive/src/response_cache/key.rs
```diff
@@ -0,0 +1,465 @@
...
+};
+use serde_json::{Map, Value as Json, json};
+use sha2::{Digest, Sha256};
+
+use crate::config::ResponseCacheConfig;
+use crate::response_cache::store::CACHE_SCHEMA_VERSION;
+
+/// Top-level request-body keys that never affect the answer and are always
+/// dropped before fingerprinting (IDs, routing, bookkeeping, streaming flag).
+pub const DEFAULT_SKIP_KEYS: &[&str] = &["stream", "user", "metadata", "service_tier", "store"];
```
willkill07: `service_tier` is removed from the key but returned in the response. Example: default then priority can return the cached default-tier response. Keep it in the key.

crates/cli/src/gateway/mod.rs
```diff
@@ -331,52 +287,22 @@ fn build_buffered_func(
...
-                    *upstream_error.lock().expect("upstream error lock poisoned") = Some(error);
-                    return Err(FlowError::Internal(message));
-                }
-            };
-            if retry_aware && !status.is_success() {
+            let bytes = response
+                .bytes()
+                .await
+                .map_err(|error| FlowError::Upstream(transport_failure(&error)))?;
+            if !status.is_success() {
```
willkill07, line 371: This turns every provider error into a Relay error envelope. It drops the upstream body and headers such as `Retry-After`. Preserve ordinary upstream errors verbatim; keep structured failures only for retry routing.

willkill07 (changes requested):

crates/cli/src/gateway/mod.rs
```diff
@@ -80,18 +82,51 @@ pub(crate) async fn passthrough(
...
+
+impl CapturedUpstreamFailures {
+    fn capture(&self, failure: CapturedUpstreamFailure, description: &str) -> FlowError {
+        let token = uuid::Uuid::now_v7();
+        self.entries
+            .lock()
+            .expect("captured upstream failures lock poisoned")
+            .insert(token, failure);
+        FlowError::Internal(format!(
+            "{CAPTURED_UPSTREAM_FAILURE_PREFIX}{token} {description}"
```
willkill07: [P1] Keep the captured error opaque. `llm_call_execute` copies every execution error into the end-event status message. A `reqwest::Error` includes the request URL, so a configured proxy/base URL with query or userinfo credentials can leak into subscriber/exporter telemetry. Store only the token in `FlowError`, then recover the transport error locally. Add a test with a secret URL and assert it is absent from emitted events.

willkill07 (commented):
One issue found that should be addressed.

willkill07 (approved):

zhongxuanwang-nv:
/ok to test [5975aa6](https://github.com/NVIDIA/NeMo-Relay/pull/404/commits/5975aa6b4e048e57360bdfc200a7b65e4dbd1450)

willkill07 (approved):

PR #410, opened 2026-07-13
feat(cli)!: edit recursive built-in plugin config

willkill07:
> [!WARNING]
> **BREAKING CHANGE — Rust editor-metadata API:** `EditorFieldKind` now includes `List` and `TaggedUnion`, so downstream exhaustive matches must handle those variants. `EditorFieldSpec` now has required `list_item` and `tagged_union` fields, so downstream struct literals must initialize them. Consumers that inspect built-in editor schemas must also handle `learners`, `config_ids`, and `target_paths` reporting `List` rather than `Json`. This changes Rust source compatibility and editor-metadata behavior only; serialized plugin configurations remain compatible.

#### Overview

Add reusable recursive editing controls to `plugins edit`.

#### Details

- Add `List` and `TaggedUnion` editor metadata plus `EditorListItemSpec`, `EditorTaggedUnionSpec`, and `EditorVariantSpec` for recursive configuration editing.
- Extend `EditorFieldSpec` with list-item and tagged-union metadata; the `editor_config!` macro accepts matching `list` and `tagged_union` options.
- Provide list add/edit/remove and string-map key/value editing controls, including duplicate-key rejection.
- Report the existing typed collections in adaptive, Guardrails, and PII configuration as `List` instead of generic `Json` editor fields.

Start with `crates/core/src/config_editor.rs`, then follow the recursive rendering in `crates/cli/src/plugins.rs`.

- Relates to: RELAY-460

## Summary

- **New Features**
  - Added list support to the config editor schema, including list-item metadata and `editor_config!` `kind: List` / `list: ...` syntax.
  - Enabled structured interactive editing for list, string-map, and tagged-union fields (add/edit/remove, plus reset/clear shortcuts).

- **Improvements**
  - Updated editor schemas so `telemetry.learners`, NeMo Guardrails `config_ids`, and `builtin.target_paths` are treated as string lists.
  - Collection display now shows item counts and appends “(default)” when applicable; scalar editing is blocked for structured values.

- **Tests**
  - Expanded coverage for list-item metadata, tagged-union lifecycle/reset, default/clear behavior, and string-map key validation.


crates/core/src/config_editor.rs
```diff
@@ -26,6 +26,10 @@ pub enum EditorFieldKind {
     StringMap,
     /// Arbitrary JSON value.
     Json,
+    /// A collection whose entries are edited recursively.
+    List,
+    /// A tagged object whose variant is selected from a discriminator field.
```
mnajafian-nv, line 31: Nice direction overall. I think `TaggedUnion` is ahead of the implementation here, and because `config_editor` is public this also widens the downstream Rust contract. `EditorFieldSpec` still has no discriminator or variant metadata and `prompt_value()` rejects `TaggedUnion`, so callers can now match on a field kind the editor cannot actually handle. I would either scope tagged-union support to `EditorListItemSpec` for now or wire the full field path before merge.

crates/cli/src/plugins.rs
```diff
@@ -866,6 +892,280 @@ where
...
+                    .expect("list value is an array")
+                    .push(entry);
+            }
+            MenuResponse::Selected(index) if index <= entries.len() => {
+                edit_existing_list_item(theme, prompt, value, index - 1, item)?;
+            }
+            MenuResponse::Cancel | MenuResponse::Selected(_) => return Ok(*value != original),
+            MenuResponse::Shortcut(MenuShortcut::Help, _) => print_editor_help(),
+            MenuResponse::Shortcut(MenuShortcut::Reset | MenuShortcut::Clear, _) => {
+                *value = json!([])
```
mnajafian-nv: I think the new collection editor needs one more pass on reset semantics before merge. This path now hardcodes an empty collection on reset, which differs from the existing `reset to default/none` behavior for fields with real defaults. That becomes visible for collection fields like `mark_exclude_names`, where reset should restore the configured default rather than `[]`. Either threading collection defaults through here or narrowing the scope to empty-default collections would make the contract clearer.

crates/core/src/config_editor.rs
willkill07: Can you clarify this comment?

> `config_editor` is public this also widens the downstream Rust contract

It is public so other crates we ship can leverage/use `config_editor!` rather than reimplement the logic. It had to be shared somewhere.

afourniernv:
Reviewed the latest diff and live-tested the nested Guardrails collection editor flows, including list add/edit and map add/remove with saved persistence and preservation of existing map entries. `just test-rust` passes on the current head. I found no blocking issues for the shipped string-collection behavior.

afourniernv (approved):
Approved. Reviewed the duplicate map-key follow-up. Live-tested the nested Guardrails collection editor flows for list add/edit and map add/remove, confirming saved persistence and preservation of existing map entries. Verified the reset/default behavior and public editor-metadata changes. No blocking issues found.

crates/core/src/config_editor.rs
mnajafian-nv: Thanks, the concern is not that `config_editor` is shared. The concern is that this PR makes `TaggedUnion` part of the public Rust editor contract for downstream consumers before the top-level field flow is complete. Right now the real built-in config changes only need recursive list and string-map editing, and the only top-level `TaggedUnion` use in this PR is the harness test. The CLI path can edit the currently selected variant payload, but it still does not provide the normal field-level lifecycle for choosing or re-choosing a variant, or the same reset and clear semantics the other field kinds already have. That makes the public surface broader than the behavior we are actually ready to support. I would either scope `TaggedUnion` to the narrower list-item metadata we need now, or add an end-to-end top-level tagged-union editing flow and behavior test before we freeze it into the public editor API.

crates/cli/src/plugins/editor_model.rs
```diff
@@ -933,6 +933,14 @@ pub(super) fn display_field_value(
     field: EditorFieldSpec,
     value: &Value,
 ) -> String {
+    if field.kind == EditorFieldKind::List {
```
mnajafian-nv: Nice summary formatting. One thing to tighten before merge: these early returns bypass the existing `(default)` marker and also collapse malformed stored values to `0 items` or `0 entries`. That means a default collection now looks user-set, and a bad manual config can look like an empty collection instead of surfacing the wrong shape. I would keep the count summary for valid arrays and objects, but preserve the default marker and fall back to the raw rendered value when the stored shape does not match the field kind.

mnajafian-nv (approved):
LGTM, conditional approval after reviewing inline suggestions

mnajafian-nv (approved):

mnajafian-nv (approved):

PR #413, opened 2026-07-13
feat(logging): add configurable operational logging

ericevans-nv:
#### Overview

Add configurable, process-wide operational logging backed by the `log` facade and `spdlog-rs`. Operational commands initialize logging during top-level dispatch, before command execution. `nemo-relay config` and `nemo-relay plugins edit` intentionally skip initialization so invalid logging configuration cannot prevent users from repairing it.

This PR establishes the logging configuration and runtime. Broader production operational event call sites will be added separately.

#### Details

- Add core-owned `LoggingConfig` and `LoggingRuntime` APIs shared by CLI and library integrations.
- Support built-in defaults, normal `config.toml` discovery, environment configuration, explicit absolute TOML paths, and direct CLI settings. CLI sources take precedence over environment configuration, which takes precedence over discovered file configuration and defaults.
- Add an always-on stderr sink for initialized operational commands and optional asynchronous JSONL or human-readable file sinks.
- Bound asynchronous file queues, reject duplicate resolved sink paths, and drain pending file records during shutdown.
- Attach one root Relay ID to operational records for process-level correlation.
- Keep logging output on stderr or configured files; operational logging does not write to stdout.
- Keep logging configuration under `[logging]` in `config.toml`. A separate interactive logging editor is outside this PR.
- Documentation for the configuration surface is tracked in #419.
- Add focused core and CLI coverage for parsing, validation, formatting, sink behavior, configuration precedence, initialization policy, and shutdown flushing.
- Regenerate `ATTRIBUTIONS-Rust.md` for the new Rust dependencies.

- `crates/core/src/logging/` for configuration types, runtime ownership, formatting, sinks, and shutdown behavior
- `crates/cli/src/commands/logging.rs` and `crates/cli/src/commands/mod.rs` for configuration-source selection and early command initialization
- `crates/cli/src/configuration/logging.rs` for normal Relay TOML integration
- `crates/core/tests/coverage/logging_tests.rs` and `crates/cli/tests/coverage/shared/config_tests.rs` for runtime and configuration guarantees
- `crates/cli/tests/coverage/commands/main_tests.rs` for CLI precedence and command policy

- Closes: RELAY-451

## Summary by CodeRabbit

* **New Features**
  * Added configurable operational logging with human-readable or JSONL output.
  * Added stderr and asynchronous file sinks with per-sink level/format, queue limits, and periodic flushing.
  * Introduced logging configuration via CLI options, environment variables, and TOML (including scoped application).
  * Logging now initializes automatically for applicable CLI commands, while configuration/editor commands can bypass it.
* **Bug Fixes**
  * Strengthened validation for configuration values and file sink destination handling (including duplicates and invalid paths).
* **Tests**
  * Added extensive logging coverage for config loading, formatter output, sink runtime behavior, shutdown behavior, and CLI precedence/policy.
* **Chores**
  * Regenerated dependency attribution and license metadata.


willkill07 (changes requested):
Logging should be part of core, not just the CLI.

I also don't understand why/how `nemo_relay.h` has been changed.

Cargo.lock
```diff
@@ -92,15 +92,15 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.102"
```
afourniernv, line 95: Something looks off in the generated dependency changes. ATTRIBUTIONS-Rust.md reproduces cleanly, so the generator itself is working, but commit d675d2f4 appears to have globally re-resolved Cargo.lock. Along with the expected additions for spdlog-rs, it bumps unrelated existing dependencies including tokio 1.51.1 → 1.52.3, hyper 1.9.0 → 1.10.1, rustls 0.23.40 → 0.23.42, clap 4.6.0 → 4.6.1, serde_json 1.0.149 → 1.0.150, log 0.4.29 → 0.4.33, and anyhow 1.0.102 → 1.0.103. Could we restore Cargo.lock from the target branch, resolve only the new logging dependencies, and regenerate ATTRIBUTIONS-Rust.md? That would avoid bundling a broad dependency refresh into this PR.

afourniernv:
Is this PR only setting up the logging infrastructure, with the actual Relay events coming later? I ran the daemon with an info-level JSONL sink, hit /healthz a couple of times, and shut it down normally, but the log file stayed empty. I also couldn’t find any production log calls. At trace level, the only records I saw came from Axum. Since the PR says it adds operational diagnostics, could we include a few basic Relay events such as startup, shutdown or agent exit, and one failure path? I don’t think we need to replace every eprintln! here. If events are planned for a follow-up, we should make that clear in the title and description.

crates/cli/src/config.rs
```diff
@@ -577,10 +577,122 @@ impl GatewayConfig {
 pub(crate) struct ResolvedConfig {
     pub(crate) gateway: GatewayConfig,
```
afourniernv: I think we should consider modularizing this better - 

split these concerns: 

  - Public/resolved logging types
  - Raw TOML types
  - String parsing
  - Validation and default inheritance

  logging/
    mod.rs       # logger and sink runtime
    config.rs    # logging config types, TOML input, validation
    format.rs    # human/JSONL formatting

crates/cli/src/logging.rs
```diff
@@ -0,0 +1,537 @@
...
+            | "refresh_token"
+            | "authorization"
+            | "auth"
+            | "openai_api_key"
+            | "anthropic_api_key"
+    )
+}
+
+#[cfg(test)]
+pub(crate) fn format_event_for_test(
```
afourniernv: Could these tests run through the real formatter instead of duplicating it here? I tested this by breaking timestamp formatting and redaction in the production path, and all 24 logging tests still passed because this helper has its own implementation of both. Keeping the fixed test inputs makes sense, but they should go through render_record so  the tests catch production regressions.

crates/cli/src/logging.rs
```diff
@@ -0,0 +1,537 @@
...
+    match value {
+        Value::String(text) => text.clone(),
+        Value::Bool(flag) => flag.to_string(),
+        Value::Number(number) => number.to_string(),
+        Value::Null => "null".into(),
+        other => other.to_string(),
+    }
+}
+
+fn is_sensitive_field_name(name: &str) -> bool {
```
afourniernv: Could we either broaden this check or document it as best-effort only? Exact matching misses common names like client_secret, provider_api_key, session_token, private_key, and credentials nested under fields such as headers. Since these logs may contain operational configuration, the current list could give a stronger redaction guarantee than it actually provides.

crates/cli/src/config.rs
```diff
@@ -966,9 +1100,75 @@ fn apply_file_config(resolved: &mut ResolvedConfig, value: toml::Value) -> Resul
...
+    }
+    let level = match config.level.as_deref() {
+        Some(raw) => LogLevel::parse(raw)?,
+        None => default_level,
+    };
+    let format = match config.format.as_deref() {
+        Some(raw) => LogFormat::parse(raw)?,
+        None => LogFormat::Jsonl,
+    };
+    let queue_capacity = match config.queue_capacity {
```
afourniernv: Could we put a reasonable upper bound on queue_capacity? Crossbeam eagerly allocates storage for every queue slot. I tried 9223372036854775807, which is valid TOML and currently passes validation, and Relay panicked during startup inside crossbeam. Smaller accidental values could still allocate a large amount of memory before anything is logged. This should probably return a config error above a documented maximum.

afourniernv (changes requested):

afourniernv:
added dep seem reasonable to be given what can be accomplished with it vs rust stdlib

crates/cli/src/logging.rs
```diff
@@ -0,0 +1,537 @@
...
+                    "failed to create async logging sink for {}: {error}",
+                    resolved_path.display()
+                ))
+            })?;
+
+        thread_pools.push(pool);
+        sinks.push(async_sink);
+
+        if file_sink.flush_interval_millis > 0 {
+            flush_interval_millis = Some(
```
afourniernv: I don’t think flush_interval_millis is actually per-sink here. We take the smallest configured interval and install it on the logger, which flushes every sink. That means a sink configured with 0 for shutdown-only flushing will still be flushed periodically when another sink has an interval, and a 60-second sink could be flushed every second. Could this either become a global setting or be implemented independently per sink?

crates/cli/src/logging.rs
```diff
@@ -0,0 +1,537 @@
...
+                CliError::Config(format!(
+                    "failed to create logging thread pool for {}: {error}",
+                    resolved_path.display()
+                ))
+            })?;
+
+        let async_sink = AsyncPoolSink::builder()
+            .sink(file)
+            .thread_pool(Arc::clone(&pool))
+            .overflow_policy(OverflowPolicy::DropIncoming)
```
afourniernv: DropIncoming still returns an error when the queue is full, and that error reaches the logger’s default handler, which writes synchronously to stderr for every dropped record. During a slow-disk or logging burst, this could flood stderr or block on it, which undermines the non-blocking queue. Could we add a logger-level handler that rate-limits or summarizes queue-full errors?

afourniernv (changes requested):

ericevans-nv:
> Logging should be part of core, not just the CLI.
> 
> I also don't understand why/how `nemo_relay.h` has been changed.

Agreed. The logging runtime, resolved configuration types, sink construction, and formatting now live under crates/core/src/logging. The CLI only parses TOML input and initializes the core runtime. The nemo_relay.h change was unrelated generated-file drift and was restored in ba2cdec.

Cargo.lock
ericevans-nv: Fixed. Cargo.lock is now based on the target branch and only adds the direct logging dependencies and their transitive dependencies. The unrelated dependency version updates are no longer present. ATTRIBUTIONS-Rust.md was regenerated, and the lockfile and attribution pre-commit checks pass.

crates/cli/src/config.rs
ericevans-nv: Addressed. Public and resolved logging types now live in `crates/core/src/logging/config.rs`, formatting is in `crates/core/src/logging/format.rs`, and logger and sink runtime behavior is in `crates/core/src/logging/sink.rs`. CLI-specific TOML parsing, validation, and default inheritance are isolated in `crates/cli/src/config/logging.rs`.

crates/cli/src/logging.rs
ericevans-nv: Fixed. The test helper now supplies deterministic inputs to the real `render_record` function instead of maintaining a separate rendering implementation, so formatter regressions are covered by these tests.

crates/cli/src/logging.rs
ericevans-nv: Good point. The exact-name redaction heuristic was removed because it could not reliably protect nested values or every possible secret field name. The backend no longer presents automatic field-name redaction as a security boundary; operational logging call sites must avoid emitting sensitive values.

crates/cli/src/config.rs
ericevans-nv: Fixed. File-sink queues now have a non-configurable maximum of 8,192 pending entries. Both TOML parsing and the core runtime reject zero or values above that limit with a configuration error, preventing oversized eager allocations.

crates/cli/src/logging.rs
ericevans-nv: Fixed. The flush interval is now a global `LoggingConfig` setting rather than a per-sink setting. One cadence applies consistently to all configured file sinks, `0` means shutdown-only flushing, and no periodic flush task is started when there are no file sinks.

crates/cli/src/logging.rs
ericevans-nv: Fixed. Each asynchronous file sink now has an error handler that reports the first queue-full condition and rate-limits repeated notices to once per second. Other uncommon sink errors are still reported immediately.

ericevans-nv:
> Is this PR only setting up the logging infrastructure, with the actual Relay events coming later?

Yes. This PR is now intentionally scoped to the operational logging backend and initialization path. Relay event call sites will be added separately, so a configured file sink can remain empty until a component emits a log record.

crates/cli/src/commands/mod.rs
```diff
@@ -105,6 +105,7 @@ async fn run_default(
     //   `nemo-relay config` remains the reconfiguration path.
     if runtime_args.requested_daemon_mode() {
         let resolved = runtime_configuration::resolve_server_config(&runtime_args)?;
+        let _logging = nemo_relay::logging::init_logging(&resolved.logging)?;
```
willkill07: Shouldn't we also be able to log/record other parts, too?

- hook forward processes
- mcp
- wrapped execution
- basically everything that isn't plugins edit or config

crates/cli/src/configuration/mod.rs
```diff
@@ -1,8 +1,13 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+mod logging;
 mod types;
 
+pub(crate) use nemo_relay::logging::{
+    DEFAULT_FILE_FLUSH_INTERVAL_MILLIS, DEFAULT_FILE_SINK_QUEUE_ENTRIES, FileLogSinkConfig,
+    LogFormat, LogLevel, LogSinkConfig, LoggingConfig, MAX_FILE_SINK_QUEUE_ENTRIES,
+};
```
willkill07: Most of these imports are not used in the file

crates/cli/src/process/launcher.rs
```diff
@@ -49,6 +50,8 @@ struct TransparentRun {
     gateway_url: String,
     dry_run: bool,
     print: bool,
+    /// Kept alive for the live-run lifetime so non-blocking file writers can flush on drop.
+    _logging: Option<LoggingRuntime>,
```
willkill07: Can the runtime always exist as part of a process? Like a Box<Mutex<LoggingRuntime>> ?

crates/core/src/logging/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+
+/// Installs process-wide operational logging from resolved config.
+///
+/// Stderr is always enabled. Explicit file sinks fail startup if they cannot be opened.
+///
+/// Verbosity comes from [`LoggingConfig::level`]: records below that minimum severity are discarded.
+/// Dropping the returned [`LoggingRuntime`] flushes sinks and detaches this logger from the
+/// process-global `log` proxy when it is still installed.
+pub fn init_logging(config: &LoggingConfig) -> Result<LoggingRuntime> {
+    let root_relay_id = Uuid::now_v7().to_string();
```
willkill07: Isn't there already a root ID managed by the scope stack?

crates/core/src/logging/sink.rs
```diff
@@ -0,0 +1,318 @@
...
+        LogLevel::Error => log::LevelFilter::Error,
+        LogLevel::Warn => log::LevelFilter::Warn,
+        LogLevel::Info => log::LevelFilter::Info,
+        LogLevel::Debug => log::LevelFilter::Debug,
+        LogLevel::Trace => log::LevelFilter::Trace,
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
willkill07: Tests do not go under src/

afourniernv:
Rechecked the dependency and attribution changes against the current base. Cargo.lock now has 303 added lines and no removals, so no existing dependency versions are being upgraded or downgraded. The 29 new lockfile packages are from the spdlog-rs dependency closure. I also compared ATTRIBUTIONS-Rust.md by package section: all 395 existing entries are unchanged, with 29 additions and no removals or modified entries. The large GitHub diff is caused by its line matcher getting confused by repeated license text; git diff --patience shows the actual change as +5,429/-0. Regenerating the file with scripts/generate_attributions.sh rust leaves the worktree clean, and a merge-tree simulation against the current base completed without conflicts.

crates/cli/src/configuration/mod.rs
```diff
@@ -1119,6 +1125,7 @@ fn apply_file_config(resolved: &mut ResolvedConfig, value: toml::Value) -> Resul
     apply_file_gateway_config(&mut resolved.gateway, config.gateway)?;
     apply_file_upstream_config(&mut resolved.gateway, config.upstream);
     apply_file_agents_config(&mut resolved.agents, config.agents);
+    logging::apply_file_logging_config(&mut resolved.logging, config.logging)?;
```
willkill07, line 1161: This enables configuration in a config.toml. Does resolving occur in library mode too? We have to have universal consistency like with plugins (plugins.toml)

willkill07 (changes requested):
Overall this is shaping up nicely.

Open things to consider/resolve:
- env var config support. I know we talked about this in DMs, but this is the only way to log short-lived processes/entities
- logging config edit command for the CLI
- parity between library and CLI process
- early/eager init of logging subsystem to catch failures

afourniernv (approved):

afourniernv (changes requested):
Yanking approval per will requests
.

crates/cli/src/commands/mod.rs
ericevans-nv: Addressed by moving logging initialization to top-level command dispatch. Logging is now initialized for hook forwarding, MCP, wrapped execution, and the other operational commands. Only config and plugins edit skip logging so invalid logging configuration does not prevent users from repairing configuration.

crates/cli/src/configuration/mod.rs
ericevans-nv: Updated the configuration modules to import logging types directly where they are used instead of broadly re-exporting them through configuration/mod.rs. The external configuration tests now also import their required logging types explicitly.

crates/cli/src/process/launcher.rs
ericevans-nv: The runtime is now owned at top-level process dispatch and remains alive for the complete operational command. It does not require Box<Mutex<LoggingRuntime>> because callers emit through the global log facade and do not mutate the runtime after initialization; the underlying logger is already thread-safe. It remains optional only for config and plugins edit, which intentionally skip logging initialization so invalid logging configuration does not prevent repair.

crates/core/src/logging/mod.rs
ericevans-nv: Yes, the scope stack has a root UUID, but it identifies one execution scope stack rather than the Relay process. Relay may create separate scope stacks for different sessions, tasks, or threads, and operational logging also begins before a scope stack or session exists. root_relay_id is generated once when the logging runtime initializes and remains stable across process startup, configuration loading, all sessions, and shutdown. Scope or session IDs can still be included as event fields when available for finer-grained correlation.

crates/core/src/logging/sink.rs
ericevans-nv: Moved the sink rate-limiter test out of src into crates/core/tests/coverage/logging_sink_tests.rs. The external test module remains attached to sink.rs so it can test private sink internals without widening their production visibility.

crates/cli/src/configuration/mod.rs
ericevans-nv: Addressed by moving logging configuration and initialization into core. CLI and library integrations now share the same LoggingConfig, validation, defaults, and LoggingRuntime. The CLI handles its file-discovery policy, while library users provide the same configuration directly, through an absolute TOML path, or through environment variables.

ericevans-nv:
I addressed the open configuration and initialization items from the latest review:

- Logging now initializes during top-level dispatch for operational commands, including hook forwarding, MCP, and wrapped execution. Only `config` and `plugins edit` skip initialization so invalid settings cannot block configuration repair.
- Environment configuration supports `NEMO_RELAY_LOG`, `NEMO_RELAY_LOG_STDERR_FORMAT`, and an absolute TOML path through `NEMO_RELAY_LOG_CONFIG_PATH`.
- CLI and library integrations share the core `LoggingConfig` and `LoggingRuntime` types, validation, defaults, and initialization behavior.
- CLI runtime overrides are available through `--log-level`, `--log-stderr-format`, and `--log-config-path`.

Logging remains under `[logging]` in the normal Relay configuration. I kept a separate interactive logging editor outside this PR because the existing file, CLI, environment, and library configuration surfaces cover the runtime use cases without introducing another configuration workflow.

crates/cli/src/commands/mod.rs
```diff
@@ -54,6 +55,33 @@ pub(crate) async fn run(bootstrap_shutdown_token: Option<String>) -> ExitCode {
...
 async fn dispatch(bootstrap_shutdown_token: Option<String>) -> Result<ExitCode, error::CliError> {
     let cli = Cli::parse();
+
+    let initialize_logging = match cli.command.as_ref() {
+        Some(command) => !command.skips_logging(),
+        None => {
+            cli.server.to_runtime().requested_daemon_mode()
+                || runtime_configuration::any_config_file_exists()
+        }
+    };
```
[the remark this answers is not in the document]
ericevans-nv: Leaving this unchanged because the intended contract is eager operational logging for every command except `config` and `plugins edit`. Dry-run affects the requested operation, not process diagnostics, so invalid logging configuration should still fail before a dry-run begins.

crates/cli/src/commands/root.rs
```diff
@@ -121,3 +124,12 @@ pub(crate) enum Command {
...
     HookForward(HookForwardCommand),
 }
+
+impl Command {
+    /// Configuration-editing commands remain available even when operational logging settings are
+    /// invalid, so users can repair their configuration.
+    pub(crate) fn skips_logging(&self) -> bool {
+        matches!(self, Self::Config(_))
+            || matches!(self, Self::Plugins(command) if command.is_edit())
+    }
```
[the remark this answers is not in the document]
ericevans-nv: Leaving this unchanged for the same lifecycle reason: top-level CLI dispatch intentionally initializes logging for `run --dry-run`. The launcher test verifies that the launcher itself does not own logging initialization; dispatch owns it for the full command lifetime.

crates/cli/tests/coverage/shared/config_tests.rs
```diff
@@ -2850,3 +2860,274 @@ fn set_test_windows_dacl(path: &std::path::Path, sddl: &str) {
...
+    assert_eq!(resolved.logging.level, LogLevel::Debug);
+    assert_eq!(resolved.logging.stderr_format, LogFormat::Human);
+    assert_eq!(resolved.logging.flush_interval_millis, 250);
+    assert_eq!(resolved.logging.sinks.len(), 2);
+    match &resolved.logging.sinks[0] {
+        LogSinkConfig::File(sink) => {
+            assert_eq!(sink.path, log_a);
+            assert_eq!(sink.level, LogLevel::Trace);
+            assert_eq!(sink.format, LogFormat::Jsonl);
+            assert_eq!(sink.queue_capacity, 32);
```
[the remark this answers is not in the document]
ericevans-nv: Leaving this unchanged because `LoggingConfig::level` is intentionally the process-wide minimum for both the `log` facade and the spdlog logger. Per-sink levels may only apply a stricter filter; they do not bypass the global minimum. The core runtime test `sink_level_filter_drops_events_below_sink_minimum` covers the supported composition.

crates/cli/tests/coverage/shared/config_tests.rs
```diff
@@ -706,6 +716,52 @@ fn persistent_user_scope_excludes_project_gateway_and_plugin_layers() {
...
+    .unwrap();
+    std::fs::write(
+        user_config_dir.join("config.toml"),
+        r#"
+[logging]
+level = "warn"
+"#,
+    )
+    .unwrap();
+    let scope = PluginConfigDiscoveryScope::enter(&nested, &xdg);
```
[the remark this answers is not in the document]
ericevans-nv: This test calls `resolve_logging_config` directly, which resolves only discovered configuration files and does not read `NEMO_RELAY_LOG*`; those variables are consumed by `LoggingArgs::resolve` before this function is called. `PluginConfigDiscoveryScope` already isolates the environment variables that can affect this direct discovery path, including `NEMO_RELAY_CONFIG_SCOPE` and `XDG_CONFIG_HOME`, so inherited logging override variables cannot change this test.

willkill07:
/ok to test [697a2cb](https://github.com/NVIDIA/NeMo-Relay/pull/413/commits/697a2cb8f6b9563eeda7f92795b2972af775aa34)

willkill07 (approved):

willkill07:
/ok to test [6c5e76c](https://github.com/NVIDIA/NeMo-Relay/pull/413/commits/6c5e76cc05876f0832ef19f1e0d9d27ab5e6b747)

Salonijain27 (approved):

afourniernv (approved):

afourniernv (approved):

PR #416, opened 2026-07-14
feat(observability)!: add multi-sink ATOF export

willkill07:
> [!WARNING]
> **BREAKING CHANGE — ATOF configuration and manual exporter APIs:** observability config now requires `components.config.version = 2` for ATOF and represents destinations as a tagged `atof.sinks` array. Legacy `output_directory`, `filename`, `mode`, and `endpoints` fields are rejected rather than translated. Manual ATOF exporters now accept one typed file or stream sink; `path` is absent for stream-backed exporters.

#### Overview

Add fan-out ATOF export through independently configured file and stream sinks.

#### Details

- Introduce observability config version 2 and validate tagged file or stream ATOF sinks, including migration diagnostics for the legacy shape.
- Create one exporter per sink so every event reaches every configured destination and teardown shuts down all exporters.
- Update Rust, FFI, Python, Node, and Go manual APIs for one typed sink, including optional paths for stream exporters and `header_env` stream headers.
- Update CLI destination reporting and doctor probes, and isolate Python and Node test recipes from ambient user plugin configuration.

Start with `crates/core/src/observability/plugin_component.rs` for the v2 validation and fan-out lifecycle, then `crates/core/src/observability/atof.rs` for the single-sink exporter contract.

Closes RELAY-459

## Summary by CodeRabbit

* **New Features**
  * Updated ATOF observability configuration to schema v2 with typed `sinks` (file and stream), including multi-sink destination support.
  * Added `header_env` support for stream sinks.
  * Stream-only ATOF exporters now expose an absent output path (instead of a file path).

* **Bug Fixes**
  * Improved observability health checks and diagnostics to probe file/HTTP/NDJSON/WebSocket transports using `sinks[...]` details, with clearer messaging when no file sinks are configured or streaming is unavailable.

* **Documentation**
  * Updated examples, bindings, and tests to use the `sinks` model (with compatibility aliases for prior endpoint naming).


crates/core/src/observability/plugin_component.rs
```diff
@@ -144,49 +147,53 @@ impl Default for ObservabilityConfig {
...
 
-/// Filesystem-backed ATOF JSONL exporter config.
+/// Multi-sink ATOF JSONL exporter config.
 ///
 /// When enabled, this section wraps
 /// [`crate::observability::atof::AtofExporter`] and writes the raw ATOF event
-/// stream as JSONL. The exporter uses the current working directory and a
-/// timestamped filename when no explicit path settings are supplied.
-#[derive(Debug, Clone, Serialize, Deserialize)]
+/// stream to one or more explicitly configured file or stream sinks.
```
mnajafian-nv, line 154: Since this is the public replacement for the old ATOF shape, I’d update the user-facing docs in the same branch before merge. README, CLI setup docs, and the observability config docs still teach `output_directory` / `filename` / `mode` / `endpoints`, while this code now rejects that shape in v2. Leaving them split would make the breaking change hard to adopt safely.

crates/core/src/observability/plugin_component.rs
```diff
@@ -2171,7 +2442,7 @@ fn observability_registration_error(error: impl std::fmt::Display) -> PluginErro
 }
 
 fn default_observability_config_version() -> u32 {
-    1
+    2
```
mnajafian-nv, line 2445: One contract mismatch to tighten here: this default means an omitted observability config version is accepted as v2, while the PR warning currently says ATOF now requires `components.config.version = 2`. I’d either make the field explicit-only or soften the PR/docs wording so the migration story matches the runtime.

willkill07: docs are a separate PR. I can't change the API behavior or shape without actually introducing an API breaking change, which I want to avoid here.

crates/core/src/observability/plugin_component.rs
willkill07: Disagree for reasons I've stated privately.

mnajafian-nv (approved):
LGTM!

Issue #401, opened 2026-07-10
[Epic]: Upgrade Switchyard to production-ready native and library integration


bbednarski9:
### Current implementation status — August 2026

The active direction is now a **Switchyard-owned external native plugin** built
against NeMo Relay's published native API v1 Rust SDK. The plugin drives
`libsy::Algorithm::run_stream` in-process while Switchyard owns provider HTTP
dispatch through `switchyard-llm-client`; it does not require
`switchyard-server`, the Decision API, a sidecar, or a Relay core/ABI change.

| Work | Status | Role in the integration |
| --- | --- | --- |
| [Switchyard #270](https://github.com/NVIDIA-NeMo/Switchyard/pull/270) | Open, ready for review, CI green | Establishes the external `nvidia.switchyard` plugin on Relay `0.7.0-rc.5`. It supports seeded weighted-random and LLM-classifier routing for buffered and streaming OpenAI Chat, OpenAI Responses, and Anthropic Messages traffic, with routing marks, retries, exactly-once fallback, and preserved same-protocol stream events. |
| [Switchyard #271](https://github.com/NVIDIA-NeMo/Switchyard/pull/271) | Open, ready for review, CI green | Independently hardens the shared provider HTTP/SSE client with redirect rejection, timeouts, bounded bodies and frames, and diagnostic redaction. It is beneficial for production use but is not a functional or merge dependency of [Switchyard #270](https://github.com/NVIDIA-NeMo/Switchyard/pull/270). |
| [Switchyard #274](https://github.com/NVIDIA-NeMo/Switchyard/pull/274) | Draft, CI green | Independently enforces buffered response loss policy. It complements [Switchyard #270](https://github.com/NVIDIA-NeMo/Switchyard/pull/270) by turning unsupported strict cross-protocol response conversion into a translation error that the plugin can send through its trusted-fallback path. |
| [Switchyard #250](https://github.com/NVIDIA-NeMo/Switchyard/pull/250) | Draft reference implementation | Prototypes `StageRouter`, including request mutation, tier prompts, handoff notes, and classifier fallback. It is based on the superseded [Switchyard #220](https://github.com/NVIDIA-NeMo/Switchyard/pull/220)/native-API-v2 integration and will be refactored onto the plugin architecture introduced by [Switchyard #270](https://github.com/NVIDIA-NeMo/Switchyard/pull/270) rather than landed as-is. |
[...]

PR #438, opened 2026-07-15
fix(switchyard): require healthy sidecar at activation

bbednarski9:
#### Overview

Require the experimental remote Switchyard plugin to confirm that its separately managed `switchyard-server` dependency is healthy before Relay activates the plugin.

Previously, Relay could register the Switchyard execution intercepts while the Decision API was unavailable. The first routed request would then encounter the connection failure and use the trusted fallback, which could hide a deployment or startup-order error. This PR makes service availability an activation prerequisite while retaining the existing runtime fallback if the service fails after activation.

#### Details

- Derive the root `/health` endpoint from `decision_api_url` without adding another configuration flag.
- Perform up to three bounded health requests during asynchronous plugin registration, retaining the existing two-second per-attempt timeout and using 100 ms/200 ms asynchronous backoff between attempts.
- Require a successful JSON response containing `{"status":"ok"}` before registering buffered or streaming execution intercepts.
- Apply the activation requirement to both `enforce` and `observe_only`; the routing rollout semantics themselves are unchanged.
- Use a separate health client so Decision API credentials are not sent to the unauthenticated health endpoint.
- Keep static configuration validation network-free and preserve trusted request-time fallback behavior after successful activation.
- Document the sidecar startup requirement in the crate and example READMEs.

The health gate is router-neutral. Tests cover Random, LLM, and StageRouter profile IDs in both rollout modes, transient startup recovery, retry exhaustion, and non-success, malformed, non-ready, and unreachable health responses.

Validation:

- `cargo test -p nemo-relay-switchyard`
- `cargo test -p nemo-relay-cli --features switchyard --test switchyard_process_e2e`
- `cargo clippy -p nemo-relay-switchyard --all-targets -- -D warnings`
- `cargo clippy -p nemo-relay-cli --features switchyard --test switchyard_process_e2e -- -D warnings`
- `examples/switchyard/run-real-e2e.sh` against Switchyard `8f9db9a6a47f848cdff1d262276ba25a8ae9cbc8`

Start with `SwitchyardRuntime::require_healthy_sidecar` and its call from `SwitchyardPlugin::register` in `crates/switchyard/src/component.rs`, then review the activation cases in `crates/switchyard/tests/unit/component_tests.rs`.

- Relates to #401

## Summary by CodeRabbit

* **New Features**
  * Added Switchyard sidecar health validation during plugin activation via `GET /health`, requiring `{"status":"ok"}` with a short timeout and limited exponential-backoff retries.
* **Documentation**
  * Updated Switchyard runtime documentation (including the example) to require a healthy `/health` endpoint before activation.
* **Tests**
  * Expanded end-to-end and unit tests with local `/health` servers to cover retry success, retry exhaustion, and invalid/unreachable responses.
* **Automation**
  * Updated the real E2E run script to wait for Switchyard health immediately after startup.


zhongxuanwang-nv (approved):
Used powerful AIs to help review, added a small one-line fix that may prevent a race condition (alr committed), and everything else looks great!

willkill07 (approved):

PR #452, opened 2026-07-16
fix(gateway): decode compressed request bodies

AjayThorve:
#### Overview

Decode supported `Content-Encoding` values before building the gateway's managed LLM request. This restores semantic request attributes for clients such as the Codex app server, which sends Responses API payloads with `Content-Encoding: zstd`.

#### Details

- Decode `zstd` and `identity` encodings in reverse application order for Relay's managed request representation without allocating for unencoded or identity-only bodies.
- Bound decoded output and the zstd decoder window to `max_passthrough_body_bytes` to avoid decompression amplification and oversized decoder allocations.
- Preserve the original bytes and headers when decoding fails or an encoding is unsupported, so raw passthrough behavior is unchanged.
- Remove `Content-Encoding` only when the managed pipeline reserializes the decoded JSON body as identity bytes.
- Add gateway regressions for semantic zstd decoding, chained encodings, output and window bounds, unsupported and malformed encodings, header cleanup, and passthrough invariants.

Without this, the request reached the provider successfully but Relay recorded `content: null`; OpenInference then emitted opaque `body_bytes` and omitted model and token semantics. The fix was also validated through the published `openai-codex` SDK app-server path and a NeMo Fabric Codex SDK + Relay end-to-end run, where Phoenix received populated chain, LLM, and tool spans.

Validation:

- `cargo test -p nemo-relay-cli gateway::tests -- --nocapture` (54 passed)
- `cargo deny check bans licenses sources`
- `uvx pre-commit run attributions-rust --all-files`
- NeMo Fabric opt-in Codex SDK + patched Relay E2E (one-shot semantic artifacts and two-turn session continuity)

No public API or configuration changes.

Start with `crates/cli/src/gateway/request.rs`. The key invariant is that decoding creates a bounded managed representation while `PreparedGatewayRequest` retains the original transport body. Then review `effective_dispatch_request` in `crates/cli/src/gateway/mod.rs`, which removes the stale encoding header only after JSON reserialization succeeds.

- Relates to NVIDIA/NeMo-Fabric#66

## Summary by CodeRabbit

* **New Features**
  * Added observability support for `zstd`-compressed request bodies containing valid JSON, with decoding that’s encoding-aware, bounded, and preserves `Content-Encoding` when applicable.
* **Bug Fixes**
  * Prevented stale `Content-Encoding` headers from being sent after gateway request bodies are rewritten/re-encoded.
  * Improved handling of unsupported, malformed, or oversized compressed observability payloads by falling back to passthrough behavior and preserving original headers.
* **Tests**
  * Expanded gateway test coverage for `zstd` decoding and `Content-Encoding`/payload edge cases.
* **Chores**
  * Updated third-party attribution metadata, including added `zstd` license details.

### DIFF
--- a/ATTRIBUTIONS-Rust.md
+++ b/ATTRIBUTIONS-Rust.md
@@ -22934,6 +22934,215 @@ limitations under the License.
 
 ```
 
+## jobserver - 0.1.35
+**Repository URL**: https://github.com/rust-lang/jobserver-rs
+**License Type(s)**: Apache-2.0
+### License: https://spdx.org/licenses/Apache-2.0.html
+```
+                              Apache License
+                        Version 2.0, January 2004
+                     http://www.apache.org/licenses/
+
+TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+1. Definitions.
+
+   "License" shall mean the terms and conditions for use, reproduction,
+   and distribution as defined by Sections 1 through 9 of this document.
+
+   "Licensor" shall mean the copyright owner or entity authorized by
+   the copyright owner that is granting the License.
+
+   "Legal Entity" shall mean the union of the acting entity and all
+   other entities that control, are controlled by, or are under common
+   control with that entity. For the purposes of this definition,
+   "control" means (i) the power, direct or indirect, to cause the
+   direction or management of such entity, whether by contract or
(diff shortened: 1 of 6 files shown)


willkill07 (approved):

Salonijain27 (approved):

PR #453, opened 2026-07-16
feat(observability): derive OTLP IDs from Relay UUIDs

willkill07:
#### Overview

- Derive OpenTelemetry and OpenInference trace and span identifiers from Relay event UUIDs.

#### Details

- Map the full root Relay UUID to the OTLP trace ID and the lower 64 bits of each Relay UUID to its OTLP span ID.
- Apply the mapping to lifecycle spans and all generated orphan or tool-projected mark spans.
- Cover root, child, orphan, and zero-byte conversion behavior for both exporters.

Review `crates/core/src/observability/mod.rs` for the shared conversions, then the OTel and OpenInference event processors and their deterministic mapping tests.

- Closes RELAY-491

## Summary by CodeRabbit

- **Enhancements**
  - OpenTelemetry and OpenInference tracing now deterministically derive trace/span identifiers from Relay event UUIDs, producing consistent IDs across agent, model/tool-projection, and orphan checkpoint spans (including scenarios without parent linkage).

- **Tests**
  - Updated observability unit tests to use fixed Relay UUIDs and validate byte-level trace/span derivation and expected span linkage.
  - Adjusted exporter parity test expectations to ensure OTEL and OpenInference session traces correlate with matching trace IDs.

### DIFF
--- a/crates/core/src/observability/mod.rs
+++ b/crates/core/src/observability/mod.rs
@@ -73,6 +73,18 @@ pub(crate) fn default_mark_exclude_names() -> Vec<String> {
 }
 
 #[cfg(any(feature = "otel", feature = "openinference"))]
+pub(crate) fn relay_trace_id(uuid: uuid::Uuid) -> opentelemetry::trace::TraceId {
+    opentelemetry::trace::TraceId::from_bytes(*uuid.as_bytes())
+}
+
+#[cfg(any(feature = "otel", feature = "openinference"))]
+pub(crate) fn relay_span_id(uuid: uuid::Uuid) -> opentelemetry::trace::SpanId {
+    let mut bytes = [0; 8];
+    bytes.copy_from_slice(&uuid.as_bytes()[8..]);
+    opentelemetry::trace::SpanId::from_bytes(bytes)
+}
+
+#[cfg(any(feature = "otel", feature = "openinference"))]
 pub(crate) fn push_common_optimization_attributes(
     attributes: &mut Vec<opentelemetry::KeyValue>,
     summary: &crate::codec::optimization::LlmOptimizationSummary,
@@ -561,3 +573,17 @@ where
 #[cfg(all(test, any(feature = "otel", feature = "openinference")))]
 #[path = "../../tests/unit/observability/attribute_projection_tests.rs"]
 mod attribute_projection_tests;
+
+#[cfg(all(test, any(feature = "otel", feature = "openinference")))]
+mod tests {
+    use super::{relay_span_id, relay_trace_id};
+    use uuid::Uuid;
+
+    #[test]
+    fn relay_id_conversions_preserve_zero_bytes() {
+        let uuid = Uuid::nil();
+
+        assert_eq!(relay_trace_id(uuid).to_bytes(), [0; 16]);
+        assert_eq!(relay_span_id(uuid).to_bytes(), [0; 8]);
+    }
+}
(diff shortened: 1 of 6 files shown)


crates/core/src/observability/mod.rs
```diff
@@ -72,6 +72,18 @@ pub(crate) fn default_mark_exclude_names() -> Vec<String> {
     vec!["llm.chunk".to_string()]
 }
 
+#[cfg(any(feature = "otel", feature = "openinference"))]
+pub(crate) fn relay_trace_id(uuid: uuid::Uuid) -> opentelemetry::trace::TraceId {
+    opentelemetry::trace::TraceId::from_bytes(*uuid.as_bytes())
+}
+
+#[cfg(any(feature = "otel", feature = "openinference"))]
+pub(crate) fn relay_span_id(uuid: uuid::Uuid) -> opentelemetry::trace::SpanId {
```
mnajafian-nv, line 81: Deterministic derivation makes sense here. One edge case still needs tightening before merge: this helper can return `SpanId::INVALID` whenever the lower 8 bytes of a caller-supplied UUID are all zero. That is broader than `Uuid::nil()`, and it turns a previously exportable event into an invalid OTLP span context. Please guard or remap that case here so exported span IDs are always non-zero, and add coverage for a non-nil UUID whose low 64 bits are zero.

crates/core/src/observability/mod.rs
```diff
@@ -561,3 +573,17 @@ where
...
 #[path = "../../tests/unit/observability/attribute_projection_tests.rs"]
 mod attribute_projection_tests;
+
+#[cfg(all(test, any(feature = "otel", feature = "openinference")))]
+mod tests {
+    use super::{relay_span_id, relay_trace_id};
+    use uuid::Uuid;
+
+    #[test]
+    fn relay_id_conversions_preserve_zero_bytes() {
```
mnajafian-nv, line 583: This test currently blesses the invalid-ID case by asserting that `Uuid::nil()` maps to all-zero trace and span IDs. I would flip the contract here: nil or low-zero UUIDs should either be rejected or mapped to a deterministic non-zero fallback, because OTLP validity requires both IDs to be non-zero. That keeps the test aligned with the exporter behavior we want to guarantee.

mnajafian-nv (commented):
Great direction overall. I left two focused comments on the OTLP ID derivation edge case so we can make sure the new exporter contract stays valid for all caller-supplied Relay UUIDs.

crates/core/src/observability/mod.rs
willkill07: I want to guarantee correlation.

A 1 in 18 quintillion chance of every bit being zero is a calculated and reasonable risk.

mnajafian-nv: Your correlation goal is right. The probability argument in the thread is not the right one for this code path, because SpanId depends only on the lower 64 bits, not the full UUID. As written, the PR is relying on an implicit UUID-shape contract that is neither validated nor documented. So the design is sound, but the contract is incomplete.

willkill07: Can you clarify what you mean here? We use UUIDv7 internally which mandates that the lower 64-bits are all random.

mnajafian-nv (approved):
Thanks, conditional approval. The helper thread still has one suggestion ready for your review.

crates/core/src/observability/mod.rs
mnajafian-nv: Thanks.

mnajafian-nv (approved):

PR #465, opened 2026-07-17
fix!: unify LLM stream close behavior

willkill07:
> [!WARNING]
> **BREAKING CHANGE — Rust and Go stream APIs:** `LlmJsonStream` is no longer a `Pin<Box<dyn Stream<...>>>` alias; Rust callers that construct it directly or rely on the aliased boxed-stream type must migrate to `LlmJsonStream::new` or `LlmJsonStream::from_closeable`. Go `LlmStream.Close` now returns an `error`, so code that depends on its previous `Close()` method type must update. Early explicit close now finalizes the partial response and emits an interrupted lifecycle end event.

#### Overview

Make explicit close behavior for managed LLM streams consistent across Rust, Python, Node, Go, and the C FFI.

#### Details

- Replace the opaque core stream alias with a close-aware managed stream wrapper that finalizes once, emits an interrupted end event, and rejects reads after close.
- Preserve Python `await stream.aclose()` while making it wait for async-generator cancellation and `aclose()` cleanup.
- Add Node `await stream.close()`, Go `stream.Close() error`, and C `nemo_relay_stream_close`.
- Make early close drain queued chunks and surface producer-cleanup errors consistently in each binding.
- Document explicit-close behavior and early-stop examples for the primary Rust, Python, and Node bindings.
- Add core, Python, Node, Go, and FFI regression coverage for idempotence, partial finalization, cleanup, and post-close exhaustion.

Start with `crates/core/src/api/runtime/callbacks.rs` and `crates/core/src/stream.rs` for the shared close contract, then review the Python, Node, FFI, and Go bridge implementations. See `docs/instrument-applications/code-examples.mdx` for consumer usage.

- Relates to: RELAY-496

## Summary by CodeRabbit

* **New Features**
  * Added explicit async stream cleanup/close APIs across Python (`aclose`), Node.js (`close`), Go (`Close() error`), and C (`nemo_relay_stream_close`), all with idempotent behavior and observable completion.
  * Streaming now supports cooperative cancellation and explicit completion signaling.
* **Bug Fixes**
  * Stopping early now reliably halts producers, finalizes partial responses, and emits the expected “interrupted” end behavior.
* **Documentation**
  * Updated migration notes, examples, and glossary to require explicit close/cleanup when exiting streams early and to clarify finalizer behavior.
* **Tests**
  * Updated/expanded coverage to validate the new close/cancellation lifecycle.

### DIFF
--- a/crates/python/src/py_api/mod.rs
+++ b/crates/python/src/py_api/mod.rs
@@ -796,8 +796,9 @@ fn llm_call_execute<'py>(
 ///     func: An async callable ``(LlmRequest) -> AsyncIterator[Any]`` that returns JSON chunks.
 ///     collector: A callable ``(Any) -> None`` invoked with each intercepted chunk
 ///         (after stream execution intercepts have been applied).
-///     finalizer: A callable ``() -> Any`` invoked once when the stream is exhausted.
-///         Its return value is the aggregated response (converted to JSON).
+///     finalizer: A callable ``() -> Any`` invoked once when the stream is exhausted
+///         or explicitly closed. Its return value is the aggregated response
+///         (converted to JSON).
 ///     handle: Optional parent scope handle.
 ///     attributes: Optional ``LlmAttributes`` bitflags.
 ///     data: Optional JSON-serializable application data.
@@ -806,6 +807,9 @@ fn llm_call_execute<'py>(
 ///     codec: Optional request codec used for annotated-aware request intercepts.
 ///     response_codec: Optional response codec used to attach annotated response data
 ///         to emitted end events.
+///
+/// Consumers that stop reading early must call ``await stream.aclose()`` to
+/// wait for producer cleanup and surface cleanup errors.
 ///
 /// Returns:
 ///     An awaitable that resolves to an ``LlmStream`` async iterator of JSON chunks.
--- a/crates/python/src/py_api/mod.rs
+++ b/crates/python/src/py_api/mod.rs
@@ -100,12 +100,30 @@ fn py_annotated_llm_response(
 pub(crate) async fn forward_stream_to_channel(
     mut stream: RustJsonStream,
     tx: tokio::sync::mpsc::Sender<FlowResult<serde_json::Value>>,
+    mut cancel: tokio::sync::watch::Receiver<bool>,
+    closed: tokio::sync::watch::Sender<bool>,
 ) {
-    while let Some(item) = stream.next().await {
-        if tx.send(item).await.is_err() {
+    loop {
+        if *cancel.borrow() {
             break;
         }
+        let item = tokio::select! {
+            _ = cancel.changed() => break,
+            item = stream.next() => item,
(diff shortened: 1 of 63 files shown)


crates/python/src/py_types/core.rs
```diff
@@ -57,6 +54,30 @@ impl PyLlmStream {
...
+                    PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(
+                        "stream close task ended before releasing the native stream",
+                    )
+                })?;
+            }
+            Ok(())
+        })
+    }
+}
+
```
mnajafian-nv, line 85: Nice cleanup. This closes and drains buffered chunks before `aclose()` returns, so the next read stops consistently.

```suggestion
        let cancel = self.cancel.clone();
        let mut closed = self.closed.clone();
        let receiver = Arc::clone(&self.receiver);

        pyo3_async_runtimes::tokio::future_into_py(py, async move {
            cancel.send_replace(true);
            while !*closed.borrow() {
                closed.changed().await.map_err(|_| {
                    PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(
                        "stream close task ended before releasing the native stream",
                    )
                })?;
            }

            let mut receiver = receiver.lock().await;
            receiver.close();
            while receiver.try_recv().is_ok() {}

            Ok(())
        })
```

crates/python/src/py_api/mod.rs
```diff
@@ -100,12 +100,30 @@ fn py_annotated_llm_response(
...
+        tokio::select! {
+            _ = cancel.changed() => break,
+            result = tx.send(item) => {
+                if result.is_err() {
+                    break;
+                }
+            }
+        }
     }
+    closed.send_replace(true);
```
ericevans-nv: `aclose()` waits for this `closed` flag, but it is published before the owned stream is dropped and before downstream Python-generator cleanup is acknowledged. Repeating the new regression test caused `aclose()` to return while the generator’s `finally` block never completed within the timeout. Could the completion signal represent the full underlying iterator cleanup before `aclose()` returns?

ericevans-nv (changes requested):
The new stream-close API is a valuable improvement, but the closure acknowledgment is currently nondeterministic. Repeating the new regression test caused `stream.aclose()` to return while the underlying generator’s `finally` block had not completed; run 18 of a 20-run loop timed out waiting for cleanup.

Please make `aclose()` wait for the full downstream Python-generator cleanup—not only the outer forwarding task’s cancellation—and ensure cleanup failures can be represented appropriately. Once that lifecycle is deterministic, the existing regression should pass consistently under repetition.

ericevans-nv (approved):
Approved. The intermittent close-acknowledgment behavior identified in review is a non-critical reliability concern and can be handled as follow-up work.

mnajafian-nv (approved):

zhongxuanwang-nv (approved):

Issue #323, opened 2026-06-27
[Enhancement]: Track topology-aware ACG cache/token benefit


teerthsharma:
### Affected area

Adaptive runtime

### Problem or opportunity

This is a feature request.

ACG can observe repeated agent workflows, but the current benefit is hard to evaluate if the metric is only "converged" or "not converged". For repeated agent tasks, the useful question is whether ACG can keep the reusable prompt prefix stable enough for cache reuse while allowing task-specific suffix content to change.

A repeated agent pipe can include the system policy, tool schema, workflow scaffold, and output contract. The task-specific suffix changes each run, but the reusable prefix should remain measurable as stable structure.

### Proposed enhancement

Track topology-aware ACG benefit with explicit cache/token-oriented metrics for repeated agent workflows.

The enhancement should make it possible to report whether ACG learned a reusable prefix, whether that prefix stayed stable across changed tasks, and whether the run avoided repeated uncached prefix processing.

Useful metrics include:

- stable prefix length and fingerprint
- profile convergence state
- observations per profile
- profile reuse count
- modeled uncached tokens
- modeled cached/read tokens
- repeated uncached prefix tokens avoided
- provider/local cache hit count when available
- full end-to-end timing when measured by a harness

### Runtime contract and binding impact

The source of truth should remain in the Rust adaptive runtime. Rust should expose the telemetry/state needed to report topology-aware ACG benefit. Node.js and Python should preserve access to the same adaptive fields where they expose adaptive runtime behavior.

Go, WebAssembly, and C FFI impact is N/A unless those experimental bindings expose the same adaptive telemetry surface later.

### Alternatives considered

Leave convergence as the only visible metric. That is weaker because a converged/not-converged result does not show whether repeated prefix work was avoided or whether the stable workflow structure stayed reusable while task content changed.

Benchmark only end-to-end latency. That is useful but not sufficient because local harness latency can be affected by machine load, ordering, time drift, or model scheduling. Token/cache counters are needed alongside latency to explain the mechanism.

### Acceptance criteria

- ACG can report stable reusable prefix behavior separately from task-specific suffix variation.
- Topology-aware profiles expose convergence state, observations, and reuse count.
- Cache/token benefit can be reported with uncached tokens, cached/read tokens, and repeated uncached prefix tokens avoided.
- Latency claims are supported by full end-to-end timing, success/timeout counts, and exact experiment setup.
- Baseline reporting uses N/A for topology convergence fields that do not apply.

[...]

PR #322, opened 2026-06-27
feat: reuse topology-aware ACG cache prefixes

teerthsharma:
#### Overview

Continuation of #282 with a narrowed focus: topology-aware ACG should learn reusable agent structure for cache reuse, not merely report faster convergence.

The practical target is repeated agent workflows where the stable prefix is reusable while the task suffix changes. That prefix can include system policy, tool schema, workflow scaffold, and output contract. The task-specific suffix remains variable.

- [x] I confirm this contribution is my own work, or I have the right to submit it under this project's license.
- [x] I searched existing issues and open pull requests, and this does not duplicate existing work.

#### Details

This branch keeps the topology-aware behavior inside `nemo-relay-adaptive` and updates ACG around that narrower cache/reuse model:

- derives ACG learning keys from stable scaffold when a cacheable scaffold exists
- tracks stable prompt-prefix fingerprints in stability analysis
- reuses converged profile stability only when the stable prefix still matches
- reopens learning when the stable prefix or output contract changes
- preserves suffix-only variation as task-specific work-item content
- threads topology-aware config through Rust, Python, Node, and Go adaptive config helpers
- adds unit and integration tests for stable-prefix convergence, cached reuse, and topology-change reopening

Local validation on the cleaned commit:

```text
cargo fmt --all --check
cargo clippy -p nemo-relay-adaptive --all-targets -- -D warnings
cargo test -p nemo-relay -p nemo-relay-adaptive
cd crates/node && node --test tests/adaptive_tests.mjs
```

A local interleaved Hermes Agent A/B run was also used to validate the new focus on cache/token movement:

```text
baseline task N, topology task N, 100 measured pairs, 5 warmup pairs
baseline completed: 100/100
topology completed: 100/100
topology cache hits: 100/100
repeated uncached prefix tokens avoided: 1,415,772
modeled provider time saved: 63.714s
observed full local E2E saved: 12.637s
```

The benchmark uses a local Anthropic-compatible cache-control emulator, so the cost claim is limited to token/cache movement and modeled provider time, not a real provider invoice.

#### Where should the reviewer start?

Start with the ACG learning and cache-reuse path:

```text
crates/adaptive/src/acg_profile.rs
crates/adaptive/src/acg_learner.rs
crates/adaptive/src/acg/stability.rs
crates/adaptive/src/acg/ir_builder.rs
crates/adaptive/src/topology.rs
crates/adaptive/tests/unit/acg_learner_tests.rs
[...]

PR #481, opened 2026-07-20
fix(adaptive): reuse stable ACG scaffolds

teerthsharma:
#### Overview

ACG includes the first non-system message in its learning key. A workflow whose
system prompt, tool schemas, and output contract never change is therefore split
into a separate profile for every distinct first user task, so observations never
accumulate and the stable scaffold is never recognized as a reusable prefix.

This change buckets learning by the stable system/tool/structured-output
scaffold, preserves the prior task seed for requests without one, and gates reuse
on a fingerprint bound to the exact scaffold key.

Scope is the three items assessed as meaningful in the [#323 review](https://github.com/NVIDIA/NeMo-Relay/issues/323#issuecomment-4976690335):
scaffold-aware keying, a stable-prefix fingerprint guard, and the
structured-output contract in `PromptIR`. No governor, drift detector,
convergence machinery, topology state, configuration surface, or binding change.
Nothing is added to Python, Node, Go, WebAssembly, or C FFI.

#### Details

5 source files (+432/-51), 18 test files (+938/-35).

- The learning key derives from the stable scaffold. Requests carrying a system
  prompt, tool schemas, or a structured-output contract bucket under
  `seed=stable-scaffold`; requests without one keep the prior task-seed behavior.
- `structured_output_schema_id` is populated and the canonicalized
  `response_format` becomes a scaffold block ahead of the first non-system
  message, so an unchanging output contract sits inside the stable prefix.
- Fingerprints bind to the exact scaffold key. A prefix stopping inside the
  scaffold binds to that scaffold; one extending past it binds to the complete
  source request, because the normalized IR retains no lossless provider-prefix
  representation.
- The gate fails closed on legacy, missing, shorter, mismatched, and
  beyond-scaffold-without-a-source-request fingerprints.
- `find_stable_prefix_length` is removed. This change replaces its only caller
  and nothing else in the workspace used it.

#### Learning-key fragmentation on `main`

Control: `main` at `2ab7a070`, no source change. Two requests sharing a system
prompt and tool schema, differing only in the first user task, key as:

```text
agent-a::model=gpt-4o::seed=user:sha256:d006::system=sha256:1a842e58b::tools=sha256:69848c057
agent-a::model=gpt-4o::seed=user:sha256:5607::system=sha256:1a842e58b::tools=sha256:69848c057
```

`system` and `tools` are identical; only `seed` differs, so the two runs land in
separate profiles and neither accumulates enough observations to produce a
reusable prefix. On this head the pair produces one key with
`seed=stable-scaffold`.

#### Reuse determinism

A profile persisted by one process is read back by another, so the analysis must
not depend on the per-process seed Rust draws for `HashMap` iteration. Over `N`
processes producing outcomes with multiplicities `c_1..c_k`, the cross-process
agreement rate is

```text
A = sum_i c_i (c_i - 1) / (N (N - 1))
```

the probability two independent processes agree, and equivalently the probability
that a persisted profile satisfies the reuse gate elsewhere. `A = 1` exactly when
the analysis is seed-independent.

Measured over `N = 20` processes on one 20-observation window whose first turn
varies in 1 of 20 runs, with four tool spans sharing a sequence index:

| Rule | Span order `A` | Prefix length `A` | Fingerprint `A` | Prefix |
|---|---|---|---|---|
| Per-span run on `sequence_index` (`main`) | 0.000 | 0.479 | 0.479 | `1` ×9, `2` ×11 |
| Per-span run on `(index, rank, span id)` | 1.000 | 1.000 | 1.000 | `1` |
| Exact prefix mass | 1.000 | 1.000 | 1.000 | `6` |

`main` derives the prefix as the leading run of per-span scores, then
re-sequences those scores by first-seen index. That index does not order the span
set: span ids carry the role and tool suffix, so `assistant-1-search` and
`assistant-1-fetch` are distinct spans at index 1, and the index is a minimum
across observations. At `A = 0.479`, better than half of cross-process profile
reads miss.

Two changes follow. Score ordering extends to `(index, stability rank, span id)`,
injective because span ids are unique within one analysis. Prefix economics walks
that vector and stops pricing at the first non-stable span, so ranking the least
stable first ends the priced prefix at a contested position instead of pricing
across it.

The prefix length no longer reads that vector. It descends the prefix tree of
block sequences while one child holds the configured share, which is the rule
named in the #323 review: the same exact prefix observed for N samples. Descent
also requires a strict majority, which two disjoint children cannot both hold, so
the dominant child is unique where it exists and no tie-break is reachable; where
none dominates, the prefix stops. Equivalently, under `d(x, y) = 2^-lcp(x, y)`
the window is ultrametric, and every point of a closed ball is a center, so the
ball is fixed by its members rather than by traversal order.

That also recovers reuse. On this window 19 of 20 observations share an exact
6-block prefix; the per-span rule reports 1 because the minority span at index 1
stops the run. The outlier still fails the fingerprint gate, so the wider prefix
stays exactly validated.

#### Test coverage

Variable first-user tasks under one scaffold; changed system prompts, tool
schemas, and output contracts; canonical structured output; whitespace changes
hidden by normalized IR; prefixes beyond the scaffold; legacy state with no
fingerprint; same-run aggregation; fresh-cache rehydration; Redis restart;
serialization compatibility.

Adversarial cases: an interleaving storage backend that yields between the
learner's load and store halves shows concurrent runs never pair a fingerprint
with a foreign observation window; two workflows sharing a system prompt and
anchor turn collapse to one key and then fail closed once the prefix extends past
the scaffold; every rotation of one span set produces a single canonical ordering
without depending on a hash seed.

The Redis restart test previously read its learning key out of the hot cache, so
it passed for whichever key `process_run` wrote — and `process_run` also persists
under the plain agent id for rehydration, so a mis-keyed scaffold record could
have satisfied it. It now asserts the key is not the bare agent id and carries the
stable-scaffold marker before reloading.

#### Validation

PR head `25c771e7` on `main` at `2ab7a070`, Windows 11, rustc 1.97.1, go 1.26.1:

| Suite | Result |
|---|---|
| `cargo test -p nemo-relay-adaptive` | 538/538 lib, 8/8 surface, 36/36 response cache, 4/4 response-cache bench, 12/12 runtime integration, 1 doctest passed and 3 ignored |
| `NEMO_RELAY_RUN_REDIS_TESTS=1 … --features redis-backend --test redis_integration` | 11/11 against Redis 8.0.5 on `127.0.0.1:6379` |
| `just test-python` | 609 passed, 5 skipped |
| `just test-node` | 342 passed, 2 skipped, 0 failed of 344 |
| `cargo clippy -p nemo-relay-adaptive -p nemo-relay-python --all-targets -- -D warnings` | clean |
| `cargo fmt --all -- --check` | clean |

The 36 response-cache and 4 response-cache benchmark tests cover the opt-in
exact-match LLM response cache from #404 and pass unchanged on this head.

#### Limits

No provider-cache, cost, or latency improvement is claimed. The analysis-only
trace report, the three-arm interleaved experiment, and the provider
cache-read/cache-write, billed-cost, parity, and p50/p95/p99 evidence requested in
points 1 through 3 of the #323 review are not supplied here; this diff is the
internal keying and validation change those points would measure. An earlier
interleaved run showed the fragmentation directly (144 keys and 0 hints versus 1
key and 120 hints) but used an older `main` and a benchmark overlay, so it is not
offered as evidence for this diff.

The exact-prefix-mass rule widens prefixes relative to `main`, which is a
behavior change and not only a determinism fix; reuse remains gated on the exact
fingerprint.

Two suites did not run on this host: `just test-go` requires `clang`/`lld`, and
full-workspace `just test-rust` reaches unrelated CLI filesystem failures on
Windows. CI provides the authoritative Linux matrix.

Persisted JSON stays backward-compatible through an optional fingerprint field,
which is source-visible to downstream Rust constructing `StabilityAnalysisResult`
with a struct literal.

`crates/adaptive/src/acg_profile.rs` for the learning-key boundary, then
`crates/adaptive/src/acg/stability.rs` for the prefix and ordering rules and
`crates/adaptive/src/acg_component.rs` for the gate. Focused tests are in
`crates/adaptive/tests/unit/acg_profile_tests.rs`,
`crates/adaptive/tests/unit/acg/stability_internal_tests.rs`, and
`crates/adaptive/tests/unit/acg_component_tests.rs`.

- Relates to #323.
- Replaces the scaffold-keying and exact-validation core of #322. The
  convergence, governor, drift, and binding scope rejected there is absent.


willkill07:
/ok to test [2eb003b](https://github.com/NVIDIA/NeMo-Relay/pull/481/commits/2eb003b24d6760cf77e9ef8a857f573a920f79ba)

willkill07 (approved):
Approving. Approach seems sensible and is limited in scope to ACG.

teerthsharma:
> Approving. Approach seems sensible and is limited in scope to ACG.

thanks for your time

Issue #501, opened 2026-07-20
[Bug]: Anthropic request codec drops provider-native tool history


afourniernv:
## Affected area

- Rust core runtime
- Python binding
- Middleware or guardrails

## Current behavior

`AnthropicMessagesCodec` can corrupt a valid provider request during a no-op managed LLM call.

The request decoder attempts to deserialize the complete Anthropic `messages` array into Relay's normalized `Vec<Message>`. Anthropic-native blocks such as `tool_use` and `tool_result` are not representable by the current normalized content types. The failed deserialization is silently converted to an empty vector:

https://github.com/NVIDIA/NeMo-Relay/blob/f8a9e96d2d644a38e741fdd4809ac2d7bad5c2db/crates/core/src/codec/anthropic.rs#L438-L442

The encoder then overlays the normalized values onto the original request. This replaces the original tool history with `messages: []` and rewrites a block-form `system` value as a string, dropping block metadata such as `cache_control`:

https://github.com/NVIDIA/NeMo-Relay/blob/f8a9e96d2d644a38e741fdd4809ac2d7bad5c2db/crates/core/src/codec/anthropic.rs#L538-L551

This occurs even when no request interceptor modifies the request because the managed execution path decodes and re-encodes whenever a request codec is supplied.

Given this input:

```json
{
  "model": "claude-sonnet-4-20250514",
  "max_tokens": 128,
  "system": [
    {
      "type": "text",
      "text": "Be terse",
      "cache_control": {"type": "ephemeral"}
    }
  ],
  "messages": [
    {"role": "user", "content": "Check the weather"},
    {
      "role": "assistant",
      "content": [
        {
          "type": "tool_use",
          "id": "toolu_1",
          "name": "weather",
          "input": {"city": "Phoenix"}
        }
      ]
    },
    {
      "role": "user",
      "content": [
        {
          "type": "tool_result",
          "tool_use_id": "toolu_1",
          "content": "sunny"
        }
      ]
    }
  ]
}
```

the provider callback receives:

```json
{
  "model": "claude-sonnet-4-20250514",
  "max_tokens": 128,
  "system": "Be terse",
  "messages": []
}
```

## Expected behavior

A no-op `decode()`/`encode()` round trip must preserve a valid provider request, including Anthropic-native message blocks and block metadata.

[...]

PR #509, opened 2026-07-20
fix(annotated)!: integrate lossless request codecs

willkill07:
#### Overview

> [!WARNING]
> **BREAKING CHANGE:** **[annotated request schema and plugin compatibility]** `AnnotatedLlmRequest` adds public fields and message, content, tool, and tool-choice variants. Rust exhaustive matches and direct struct literals must be updated. Request annotations and request-intercept outcomes now use the `@2` schema tags. Rust native plugins and Rust `grpc-v1` workers that consume annotations must rebuild against NeMo Relay 0.6; Python workers must update to the 0.6 worker SDK. Dynamic plugins that register LLM request intercepts must declare `compat.relay = ">=0.6,<1.0"`; the host rejects ranges that still admit Relay 0.5. Node.js, Go, and raw C FFI callback boundaries remain JSON-based and do not require new binding DTOs.

Fix Issue 501 by making request-codec annotations lossless across Anthropic Messages, OpenAI Responses, and OpenAI Chat Completions, and by enabling those request codecs in the `nemo-relay` gateway.

#### Details

* Extend `AnnotatedLlmRequest` with shared provider instructions, a tagged mutable `api_specific` surface, portable message/content/tool unions, component extras, and explicit provider-native fallbacks.
* Complete request modeling against the pinned Anthropic and OpenAI SDK schemas, including provider-only input items, built-in tools, tool choices, legacy Chat fields, and provider-specific top-level controls.
* Guarantee JSON-value identity for no-op `encode(decode(original), original)` calls by comparing edits with a decoded baseline and patching only changed fields. Unknown fields, nulls, alternate representations, item order, and untouched native values remain unchanged.
* Patch edited nested messages, content blocks, tools, and tool choices against their original provider values so sibling unknown fields and explicit nulls survive portable edits.
* Reject malformed recognized structures, provider-mismatched native edits, unencodable normalized edits, and raw-body mutation before the provider callback. Remove the Responses private sentinel and Chat's implicit `stream_options.include_usage` injection.
* Enable request codecs for buffered and streaming `/v1/messages`, `/v1/chat/completions`, and `/v1/responses` gateway calls. Count-token, model, probe, and non-LLM passthrough routes remain unannotated.
* Add Python constructor properties and serialization coverage. Version request annotations and request-intercept outcomes as `@2`, require the 0.6 SDK for Python workers, and enforce a Relay 0.6 compatibility floor for dynamic plugins that register request intercepts. Node, FFI, and Go callback boundaries continue carrying annotations as JSON.
* Add provider schema fixtures, exact round-trip and surgical-edit tests, an Issue 501 managed-runtime reproduction, a six-way gateway route/streaming matrix, migration documentation, and updated typed-wrapper guidance.
* Migration note: gateway interceptors for the three generation routes must mutate `annotated_request`; changing raw `request.content` while a request codec is active remains an explicit runtime error.

Validation:

* `just test-rust`
* `just test-python` (536 passed)
* `just test-node` (277 passed)
* `just test-go`
* `just test-python-plugin`
* `just test-claude-plugin-e2e`
* `just test-codex-plugin-e2e`
* `just docs`
* `uv run pre-commit run --all-files`

Start with `crates/types/src/codec/request.rs` for the public annotation contract, then review the baseline-diff helper and provider codecs under `crates/core/src/codec/`. The gateway wiring and buffered/streaming matrix are in `crates/cli/src/gateway/mod.rs` and `crates/cli/tests/coverage/shared/server_tests.rs`.

* Closes NVIDIA/NeMo-Relay#501

## Summary by CodeRabbit

* **New Features**
  * Added lossless request annotations for Anthropic Messages, OpenAI Chat Completions, and OpenAI Responses routes, including streaming.
  * Added support for shared instructions, developer messages, tool calls/results, multimodal content, and provider-native fields.
  * Exposed `instructions` and `api_specific` through Python request objects.
* **Bug Fixes**
  * Preserves unchanged fields, ordering, explicit nulls, metadata, and unknown provider data during edits.
  * Rejects unsupported raw-body mutations when request codecs are active.
* **Documentation**
  * Expanded codec behavior, routing, migration, and interception guidance.

## Summary by CodeRabbit

- **New Features**
  - Added lossless, baseline-aware request codecs for Anthropic Messages, OpenAI Chat Completions, and OpenAI Responses to enable safe annotation edits while preserving unchanged JSON.
  - Expanded annotated requests to support `instructions`, `api_specific`, developer/function/tool/provider-native variants, and richer tool schemas.
  - Added Python accessors for `instructions` and `api_specific`; enhanced OpenInference message-role projection.

- **Bug Fixes**
  - Stricter validation/encoding behavior, improved refusal/text extraction, and enforced rejection of unsupported raw-body and stream-mode edits when codecs are active.
  - Enforced annotated-request consumer compatibility.

- **Documentation**
  - Updated codec and migration guidance for the new annotation surfaces.

- **Tests**
  - Updated and expanded codec, gateway, plugin, and observability test coverage.


willkill07:
Verified end-to-end successfully.

|Client | Version | Route | Result |
|--|--|--|--|
|Claude Code | 2.1.160 | /v1/messages | 10 cold + 2 concurrent runs passed|
|Codex | 0.144.4 | /v1/responses | 10 cold + 2 concurrent runs passed |

</div></div><p>Commands:</p><ul><li><code>just test-claude-plugin-e2e</code></li><li><code>just test-codex-plugin-e2e</code></li></ul><p>Both used the real installed clients, temporary Relay plugin installations, codec-enabled gateways, and an isolated deterministic provider. Requests returned <code>pong</code>, lifecycle/LLM scopes paired correctly, provider responses were preserved in ATOF, and cleanup succeeded.</p><p>This verifies client/gateway/codec compatibility, not live Anthropic/OpenAI service availability. The PR branch remains clean and unchanged.</p></body></html>

crates/core/src/codec/anthropic.rs
```diff
@@ -530,66 +798,222 @@ impl LlmCodec for AnthropicMessagesCodec {
...
+                );
             }
         }
-
-        // Overlay tools in Anthropic format: { name, description, input_schema }.
-        // Denormalize from ToolDefinition (drop type/function wrapper, rename parameters -> input_schema).
-        if let Some(ref tools) = annotated.tools {
-            let anthropic_tools = encode_anthropic_tools(tools);
-            insert_serialized(obj, "tools", &anthropic_tools, "tools")?;
+        if annotated.tools != baseline.tools {
```
afourniernv, line 887: Could we preserve nested provider fields when editing `tools` or `tool_choice`? I reproduced a request where changing only `parallel_tool_calls` and one function name drops an unknown `tool_choice` field and explicit-null `description` / `strict` fields. This path rebuilds the complete provider values rather than patching the original, so it does not yet satisfy the documented surgical-edit guarantee. A regression case with nested unknown fields and explicit nulls while editing a sibling field should catch it; the Chat and Responses tool paths appear to use the same full-array reconstruction pattern.

afourniernv:
One compatibility question before this lands: the PR adds new `Message`/`ContentPart`/`ToolDefinition` variants and `AnnotatedLlmRequest` fields, but worker invocations still label the payload `nemo.relay.AnnotatedLlmRequest@1`. Since plugin guidance recommends `compat.relay = ">=0.5,<1.0"`, a Rust worker built against 0.5 can be admitted by a 0.6 host and then fail to deserialize a new variant. Is `@1` intended to remain backward-readable here? If so, we need a catch-all/compatible representation; otherwise I think this needs a schema-version/compatibility change and an explicit 0.6 migration note for the Rust source breaks.

crates/core/src/codec/anthropic.rs
willkill07: This has been done

bbednarski9:
This seems worth a look:

encode_changed_items treats equal-length arrays as structurally aligned, while patch_changed_json always associates original values by index. After reordering, unmodeled provider fields stay at their old positions instead of following their logical items.

I reproduced this by swapping two Chat image_url parts containing nested provider markers. The resulting first part was {url: "b", vendor_marker: "A"} rather than marker "B".

crates/core/src/codec/openai_responses.rs
```diff
@@ -277,187 +288,643 @@ fn optional_vec<T>(items: Vec<T>) -> Option<Vec<T>> {
...
+    })?;
+    if obj.get("type").and_then(Json::as_str) != Some("function") {
+        let kind = obj.get("type").and_then(Json::as_str).unwrap_or("unknown");
+        return Ok(ToolDefinition::ProviderNative {
+            provider: "openai_responses".into(),
+            kind: kind.into(),
+            value: value.clone(),
+        });
     }
+    let (function, wrapper_extra) =
```
bbednarski9, line 683: Double check behavior here:

The decoder accepts both flat and wrapped function tools, but the encoder always produces the flat form. Editing a wrapped tool’s name leaves function.name unchanged and adds a conflicting top-level name.

My probe expected function.name == "new_name" but received "old_name". The provider can therefore execute the stale tool definition or reject the malformed mixed representation.

crates/core/src/codec/openai_chat.rs
```diff
@@ -315,84 +893,356 @@ impl LlmCodec for OpenAIChatCodec {
             max_tool_calls: None,
             top_logprobs: obj.get("top_logprobs").and_then(|v| v.as_u64()),
             stream: obj.get("stream").and_then(|v| v.as_bool()),
+            api_specific: Some(ApiSpecificRequest::OpenAIChat {
+                audio: obj.get("audio").cloned(),
+                frequency_penalty: obj.get("frequency_penalty").and_then(Json::as_f64),
```
bbednarski9: [P2] Reject wrong JSON types for modeled scalar controls

I reproduced this at the current head with focused tests for Chat frequency_penalty: "high", Anthropic top_k: "many", and Responses background: "yes". In all three cases decode succeeds, the recognized field is absent from annotated.extra, and encode(decode(request), request) returns the original malformed value unchanged. Middleware therefore sees the control as absent while the provider still receives invalid JSON, which conflicts with the PR guarantee that malformed recognized structures are rejected before the provider callback.

Could the decoders distinguish missing/null from a present wrong type and return InvalidArgument? Please add regression cases for modeled scalar controls across these codec paths.

bbednarski9:
## P2 verification: malformed modeled scalars are silently accepted

I reproduced this issue against the current PR head, commit `61cd11ceaf94a04781a0d23e7bb84c8208840f94`.

### Cases tested

| Codec | Recognized modeled field | Malformed value |
|---|---|---|
| OpenAI Chat Completions | `frequency_penalty` | `"high"` instead of a number |
| Anthropic Messages | `top_k` | `"many"` instead of an integer |
| OpenAI Responses | `background` | `"yes"` instead of a boolean |

For each case, the focused diagnostic test performed the same checks:

1. Constructed an otherwise-valid provider request containing the malformed modeled scalar.
2. Called the codec's `decode` method and confirmed that it returned `Ok` rather than `FlowError::InvalidArgument`.
3. Confirmed that the recognized field was absent from `annotated.extra`, so middleware cannot inspect or correct it as an unmodeled provider value.
4. Called `encode(&annotated, &original)` and confirmed that the result exactly equaled the original request, including the malformed scalar value.

The focused command was:

```text
env CARGO_TARGET_DIR=/Users/bbednarski/Projects/code-review/NeMo-Relay/target cargo test -p nemo-relay review_probe_malformed_modeled -- --nocapture
```

Result:

```text
running 3 tests
test codec::anthropic::tests::review_probe_malformed_modeled_anthropic_scalar_is_forwarded ... ok
test codec::openai_responses::tests::review_probe_malformed_modeled_responses_scalar_is_forwarded ... ok
test codec::openai_chat::tests::review_probe_malformed_modeled_chat_scalar_is_forwarded ... ok

test result: ok. 3 passed; 0 failed
```

These were diagnostic assertions of the current behavior, so the passing result confirms the bug rather than the desired contract.

### Impact

A present value of the wrong JSON type is treated the same as an absent value by the typed annotation. Because the baseline-aware encoder preserves untouched raw fields, the malformed value is still forwarded to the provider. Middleware therefore sees the control as absent while the provider receives invalid JSON. This conflicts with the PR's stated behavior of rejecting malformed recognized structures before invoking the provider callback and may defer errors to provider-specific validation.

### Expected behavior

For recognized modeled fields, decoding should distinguish:

- missing or accepted `null`, where appropriate; from
- a present value with the wrong JSON type.

The latter should return `FlowError::InvalidArgument`. Regression coverage should include modeled scalar controls across all three codec paths, not only malformed nested components.

Inline discussion: https://github.com/NVIDIA/NeMo-Relay/pull/509#discussion_r3619270748

crates/cli/src/gateway/mod.rs
afourniernv, line 119: **[P1] Use the intercepted request to determine stream handling**
The gateway selects `run_managed_streaming` or `run_managed_buffered` from `prepared.streaming` before request interceptors run.

I reproduced an interceptor changing `stream` from `true` to `false`. The encoded request sent upstream correctly used `stream: false`, but the gateway remained on the streaming path and returned an empty response body.

We should either determine the response path from the final encoded request or reject interceptor changes to `stream`. Maybe covering both `true -> false` and `false -> true`.

crates/core/src/codec/anthropic.rs
```diff
@@ -530,66 +795,237 @@ impl LlmCodec for AnthropicMessagesCodec {
...
+                    Some(match obj.get("tool_choice") {
+                        Some(original) => super::patch_changed_json(original, &before, &edited),
+                        None => edited,
+                    })
+                }
+                (Some(edited), None) => Some(encode_tool_choice_with_parallel_hint(
+                    edited,
+                    annotated.parallel_tool_calls,
+                )?),
+                (None, _) => None,
```
afourniernv: **[P2] Do not silently drop standalone `parallel_tool_calls` edits**

When the original request has no `tool_choice`, setting `annotated.parallel_tool_calls = Some(false)` reaches this branch and produces neither a provider-visible change nor an error.

Since Anthropic carries this setting inside `tool_choice`, we should either synthesize an `auto` choice with `disable_parallel_tool_use: true` or reject the edit as unencodable.

bbednarski9 (changes requested):

crates/core/src/codec/openai_responses.rs
willkill07: Resolving

crates/core/src/codec/openai_chat.rs
willkill07: Resolving

crates/cli/src/gateway/mod.rs
willkill07: Resolving

crates/core/src/codec/anthropic.rs
willkill07: Resolving

crates/core/src/codec/mod.rs
```diff
@@ -25,6 +25,246 @@ pub mod response;
...
+        .enumerate()
+        .filter_map(|(index, item)| item.is_none().then_some(index))
+        .collect::<Vec<_>>();
+    let unmatched_baseline = used
+        .iter()
+        .enumerate()
+        .filter_map(|(index, item)| (!item).then_some(index))
+        .collect::<Vec<_>>();
+
+    if !structurally_changed && edited.len() == baseline.len() {
```
bbednarski9, line 239: [P2] Do not positionally pair multiple unmatched edited items

The new pure-reorder regression passes, but a reorder combined with edits still attaches unmodeled provider fields to the wrong logical items. I reproduced this with two Chat `image_url` parts containing nested provider markers: decode `a/A` and `b/B`, swap the parts, then change both URLs to `b-edited` and `a-edited`. The encoded result is `b-edited/A`, `a-edited/B`; the expected result is `b-edited/B`, `a-edited/A`.

Focused probe:

```text
cargo test -p nemo-relay review_probe_reordered_and_edited_parts -- --nocapture

left:  [b-edited + vendor_marker A, a-edited + vendor_marker B]
right: [b-edited + vendor_marker B, a-edited + vendor_marker A]
```

When every moved item is also edited, the equality passes above find no match and `structurally_changed` remains false. This fallback then assigns every unmatched item to the same index, after which `patch_changed_array` grafts raw fields from the wrong original element.

This is fundamentally ambiguous with `PartialEq` alone. Could this path avoid positional patching unless identity is proven by equality or a stable key? A safe minimal behavior would reject an ambiguous compound edit rather than silently cross-wire provider data; a broader solution would carry stable provenance or nested extras with each annotated item. Please add both a shared parity test and a provider-level reorder-plus-edit regression.

bbednarski9 (commented):
Follow-up testing found one residual P2 in the new array-alignment fix.

crates/core/src/codec/mod.rs
willkill07: Resolving

willkill07: > A safe minimal behavior would reject an ambiguous compound edit

This was done

bbednarski9:
## P2: malformed modeled structured controls are still forwarded

I reproduced this against the current PR head, commit `8756844d52f34a9ee7b8787054bd4b3300351830`.

The scalar validation added in the recent commits works, including Chat `seed`, but recognized fields stored as raw `Json` still do not validate their outer JSON shape.

### Cases tested

| Codec | Recognized field | Malformed value | Expected shape |
|---|---|---|---|
| OpenAI Chat Completions | `response_format` | `"json_object"` | object |
| OpenAI Responses | `include` | `"reasoning.encrypted_content"` | array |
| Anthropic Messages | `metadata` | `"user-123"` | object |

For all three cases, the diagnostic test confirmed that:

1. `decode()` returned `Ok`.
2. The recognized field was excluded from `annotated.extra`.
3. `encode(&annotated, &original)` reproduced the malformed value unchanged.

I also exercised the Chat case through the real gateway request-interceptor path. A request containing `response_format: "json_object"` received HTTP 200 from the relay test gateway, and the mock upstream captured that malformed value unchanged. This shows the value reaches the provider callback rather than being rejected at the codec boundary.

The focused results were:

```text
running 3 tests
test codec::anthropic::tests::review_probe_malformed_modeled_structure_is_accepted_anthropic ... ok
test codec::openai_responses::tests::review_probe_malformed_modeled_structure_is_accepted_responses ... ok
test codec::openai_chat::tests::review_probe_chat_seed_mutation_and_malformed_structure ... ok

test result: ok. 3 passed; 0 failed

running 1 test
test server::tests::review_probe_gateway_seed_and_malformed_structure ... ok

test result: ok. 1 passed; 0 failed
```

These passing tests assert the current behavior, so they confirm the gap rather than the desired contract. This conflicts with the PR description's guarantee that malformed recognized structures are rejected before the provider callback.

### Suggested fix and cost

The decode paths currently use direct `.cloned()` reads for these fields (and several adjacent object/array-valued controls). I suggest adding shallow `optional_object` and `optional_array` helpers alongside the existing scalar helpers, then using them for fields whose provider schema has an unambiguous outer type. Unknown members inside valid objects should remain untouched.

This should have negligible compute overhead. `serde_json::Value` has already been parsed, so validation only checks its enum variant—constant-time, with no schema traversal or reserialization. The existing `.cloned()` call already performs the potentially meaningful copy; inspecting the value's outer type first does not add another allocation.

Relevant decode sites:

- `crates/core/src/codec/openai_chat.rs` around `metadata`, `response_format`, and the other raw-JSON API controls.
- `crates/core/src/codec/openai_responses.rs` around `truncation`, `reasoning`, `include`, `metadata`, and raw-JSON API controls.
- `crates/core/src/codec/anthropic.rs` around `metadata`, `cache_control`, `output_config`, and `thinking`.

bbednarski9 (approved):

bbednarski9 (approved):

PR #518, opened 2026-07-21
docs: clarify subscriber delivery and shutdown

willkill07:
#### Overview

Clarify when NeMo Relay subscribers and observability exporters finish processing events, including the behavior during graceful and abrupt shutdown.

#### Details

- Define emitted, delivered, and exported events in the subscriber concept documentation.
- Document the shared subscriber flush barrier and the downstream barriers for ATOF, ATIF, OpenTelemetry, and OpenInference.
- Explain the file and stream differences for ATOF, including the limits of process-writer flushing and abrupt termination.
- Clarify ATIF export timing and remove the redundant subscriber flush from the Rust example.
- Update the Node.js custom subscriber example to allow an event-loop turn after native dispatch.
- Update public ATOF method comments in the Rust, Python, and Node.js bindings without changing behavior or signatures.
- Preserve the existing fail-open telemetry behavior. This change has no runtime, configuration, compatibility, or breaking change.

Validation:

- `just docs-linkcheck`
- `just docs`
- `git diff --check`
- Runtime test suites were not run because the change is limited to documentation, examples, and public doc comments.

Start with `docs/about-nemo-relay/concepts/subscribers.mdx`, which defines the canonical delivery contract and the boundary between subscriber flush and exporter-specific completion.

- Relates to: RELAY-541

## Summary by CodeRabbit

- **Documentation**
  - Clarified telemetry lifecycle milestones (emitted vs delivered vs exported) and the correct barrier semantics for queued delivery, file sink flushing, and stream sink draining/closing (including stream timeouts being logged without failing).
  - Updated `force_flush()`/`shutdown()` and `flush_subscribers()` guidance across Rust, Python, native typings, and FFI, including re-entrant behavior and ordering.
  - Expanded observability plugin teardown and configuration guidance (including plugin teardown steps and warnings about telemetry loss/truncated outputs on abrupt termination).
  - Refreshed ATIF/ATOF, OpenTelemetry/OpenInference, and Node “Getting Started” (extra event-loop turn after scope creation).

### DIFF
--- a/docs/about-nemo-relay/concepts/subscribers.mdx
+++ b/docs/about-nemo-relay/concepts/subscribers.mdx
@@ -24,6 +24,11 @@ waiting for subscriber callbacks or exporter work.
 Events describe what happened. Subscribers are the components that watch those
 events.
 
+In this documentation, an event is **emitted** when the runtime submits it for
+subscriber dispatch. An event is **delivered** to a subscriber when its callback
+runs. An event is **exported** when an exporter completes the downstream work
+defined by its API. On native targets, these are separate milestones.
+
 That separation matters:
 
 - The runtime can emit one canonical event stream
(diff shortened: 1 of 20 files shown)


docs/about-nemo-relay/concepts/subscribers.mdx
```diff
@@ -127,15 +132,38 @@ order and subscriber snapshot order.
...
 
-Use the flush API when application shutdown, tests, or examples must observe
-side effects from subscriber callbacks that have already been queued:
+An event-producing call returning is not a delivery barrier. Use the following
+milestones to choose the barrier that matches the output you need to observe:
+
+| Milestone | What It Establishes | What It Does Not Establish |
+|---|---|---|
+| Event-producing call returns | The runtime call completed. On native targets, the runtime normally queues subscriber work for background dispatch. | Subscriber callbacks ran, exporter work completed, or output is visible or durable. |
+| Subscriber flush returns | All native subscriber callbacks queued before the flush call have completed. | Exporter-owned workers drained or a remote backend accepted the data. |
```
bbednarski9, line 143: **[P2] Document the re-entrant flush exception**

This guarantee does not hold when `flush_subscribers()` is called from inside a native subscriber callback. The dispatcher returns `Ok(())` immediately when `IN_DISPATCHER` is set, so callbacks later in the same subscriber snapshot can still run after the call returns.

I reproduced this against the current PR head with ordered outer and inner scope-local subscribers: the outer callback called `flush_subscribers()` and observed the inner callback count as `0`; after an external flush, the count was `1`.

Please qualify this guarantee as applying when called outside subscriber callbacks and warn that flush/export/shutdown/plugin-clear barriers should not be invoked re-entrantly, or change re-entrant flushes to return an explicit error.

bbednarski9 (commented):
One remaining documentation-accuracy issue was confirmed with a focused regression test against the current PR head.

bbednarski9 (approved):

PR #542, opened 2026-07-23
fix: emit OpenInference system prompts as input messages

mnajafian-nv:
#### Overview

Align OpenInference export behavior with the expected attribute contract by stopping Relay from writing prompt text into `llm.system`. System content is now exported as ordered `llm.input_messages` entries instead.

#### Details

- Remove `llm.system` emission from both replay-payload projection and annotated request projection in the OpenInference exporter.
- Preserve replay `systemPrompt` content by exporting it as the leading `llm.input_messages.<index>` entry with role `system`.
- Preserve `AnnotatedLlmRequest.instructions` as the leading system input message when present.
- Preserve annotated `Message::System` content in `llm.input_messages` instead of the legacy `llm.system` field.
- Harden replay input projection so `prompt` still falls back when replay `messages` is empty or every replay message is skipped, and skip incomplete replay message objects unless they provide both a string `role` and displayable `content`.
- Update unit coverage to assert that `llm.system` is absent, system and user messages are exported in the expected order, replay fallback remains contiguous, and migrated system-message content is preserved.

Validation:
- `cargo fmt --all`
- `just test-rust`
- `just test-python`
- `just test-go`
- `just test-node`
- `uv run pre-commit run --files crates/core/src/observability/openinference.rs crates/core/tests/unit/observability/openinference_tests.rs`

Start in `crates/core/src/observability/openinference.rs`, especially the request projection path around `push_llm_request_attributes()` and `push_replay_input_messages()`. The most relevant regression coverage is in `crates/core/tests/unit/observability/openinference_tests.rs`.

- Relates to: none

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OpenInference tracing by representing the system prompt as the first `llm.input_messages` entry with `role: system`, instead of emitting a separate `llm.system` scalar.
  * Ensured consistent `llm.input_messages` indexing/ordering for both annotated requests (from instructions) and replay inputs, with correct shifting of subsequent messages.
  * Added more robust replay handling: only well-formed message entries (with required role/content) are projected, and fallback behavior works for empty or incomplete replay inputs.

* **Tests**
  * Updated and expanded OpenInference flattened-attribute tests to match the new system-message semantics and indexing.

### DIFF
--- a/crates/core/src/observability/openinference.rs
+++ b/crates/core/src/observability/openinference.rs
@@ -1038,7 +1038,13 @@ fn push_annotated_request_attributes(
     if let Some(params) = request.params.as_ref().and_then(to_json_string) {
         attributes.push(KeyValue::new(oi::llm::INVOCATION_PARAMETERS, params));
     }
-    push_annotated_input_messages(attributes, &request.messages);
+    let mut next_index = 0usize;
+    if let Some(instructions) = request.instructions.as_ref().and_then(message_content_text) {
+        push_message_role(attributes, "llm.input_messages", next_index, "system");
+        push_message_text_value(attributes, "llm.input_messages", next_index, instructions);
+        next_index += 1;
+    }
+    push_annotated_input_messages(attributes, &request.messages, next_index);
     if let Some(tools) = request.tools.as_deref() {
         push_annotated_tools(attributes, tools);
     }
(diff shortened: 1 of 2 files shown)


crates/core/src/observability/openinference.rs
```diff
@@ -1259,6 +1275,7 @@ fn push_replay_input_message(attributes: &mut Vec<KeyValue>, index: usize, messa
             text,
         ));
     }
+    true
```
willkill07: I think this true should be hoisted inside of the `if let`. Otherwise, empty/null input message would falsely categorized as being replayed/inserted.

willkill07 (approved):
One nit, otherwise LGTM

crates/core/src/observability/openinference.rs
mnajafian-nv: Agreed!

willkill07 (approved):

willkill07 (approved):

PR #545, opened 2026-07-23
fix: avoid duplicate ATIF user steps on continuations

mnajafian-nv:
#### Overview

Fixes an ATIF exporter bug where a continued LLM request after tool work could emit a second duplicate `user` step for the same logical turn. The exporter now keeps one `user` step per fresh user turn, preserves same-turn continuation requests on the matching `agent` step, and avoids carrying an unpaired continuation request onto a later unrelated step.

#### Details

- Update the ATIF LLM-start mapping so only fresh user turns emit a `user` step.
- Treat same-turn continuations as part of the matching `agent` step instead of repeating the user message.
- Preserve the full continuation request in `Step.extra.llm_request` so request fidelity is not lost.
- Add turn-state detection for raw chat requests, OpenAI Responses-style inputs, and annotated requests.
- Make raw message turn-state detection content-aware so user messages that carry only tool-use or tool-result continuation content stay in the same logical turn.
- Keep OpenAI Responses typed non-user continuation items in the same logical turn instead of re-emitting a duplicate `user` step.
- Only treat LLM ends with event data as pairable, which prevents a stashed continuation request from leaking onto a later unrelated `agent` step.
- Add regression coverage for OpenAI Responses continuations with `function_call_output` and `shell_call`, Anthropic tool-result continuation, and an empty-end continuation followed by a fresh user turn.
- Update the existing full agent-loop test to reflect the corrected ATIF step sequence.

**Validation:**

- `cargo fmt --all`
- `cargo test -p nemo-relay observability::atif::tests::`
- `cargo test -p nemo-relay-cli --lib -- --nocapture`
- `uv run pre-commit run --files crates/core/src/observability/atif.rs crates/core/tests/unit/atif_tests.rs`

Start in `crates/core/src/observability/atif.rs`, especially the turn-state detection around `llm_start_user_step_message()`, `chat_messages_turn_state()`, and the pairability lookup in `EventLookupMaps::from_events_with_correlation_events()`. The most important regression coverage is in `crates/core/tests/unit/atif_tests.rs`.

- Relates to: none

### DIFF
--- a/crates/core/src/observability/atif.rs
+++ b/crates/core/src/observability/atif.rs
@@ -1044,17 +1044,7 @@ fn openai_responses_item_turn_state(
             Some("user") | None => Some(RequestTurnState::FreshUser),
             Some(_) => Some(RequestTurnState::Continuation),
         },
-        Some(
-            "function_call"
-            | "custom_tool_call"
-            | "mcp_tool_call"
-            | "tool_call"
-            | "function_call_output"
-            | "custom_tool_call_output"
-            | "mcp_tool_result"
-            | "tool_result"
-            | "reasoning",
-        ) => Some(RequestTurnState::Continuation),
+        Some(_) => Some(RequestTurnState::Continuation),
         _ => None,
     }
 }
--- a/crates/core/src/observability/atif.rs
+++ b/crates/core/src/observability/atif.rs
@@ -18,7 +18,7 @@
 //!
 //! | NeMo Relay Event     | ATIF Step               | Notes                                |
 //! |-----------------|-------------------------|--------------------------------------|
-//! | LLM Start       | `user` step             | Messages extracted from LlmRequest   |
+//! | LLM Start       | `user` step             | Fresh user turns only; same-turn continuations stay on the agent step |
 //! | LLM End         | `agent` step            | Response content, tool_calls promoted|
 //! | Tool Start      | *(skipped)*             | tool_calls come from LLM End instead |
 //! | Tool End        | agent observation         | Correlated by `source_call_id`       |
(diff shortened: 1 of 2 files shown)


crates/core/src/observability/atif.rs
```diff
@@ -1418,11 +1563,17 @@ impl EventLookupMaps {
                     llm_start_model_names.insert(event.uuid(), model_name);
                 }
             }
+            if event.scope_category() == Some(crate::api::event::ScopeCategory::End)
+                && event.category().map(|category| category.as_str()) == Some("llm")
+            {
+                llm_end_uuids.insert(event.uuid());
```
ericevans-nv, line 1621: Could this only treat an LLM end as pairable when `event.data()` is present? `handle_llm_start()` stashes a continuation request whenever this set contains the UUID, but `handle_llm_end()` returns immediately when that end has no data. The stashed request can then survive and be serialized onto the next unrelated agent step. Please handle this no-output path and add a regression covering an empty end followed by another LLM call.

ericevans-nv (changes requested):
The continuation classification and regression coverage look well targeted. I found one lifecycle edge case that can misattribute a continuation request: an LLM end event without output is currently considered paired, causing the request to be stashed, but `handle_llm_end()` returns without consuming it. That pending request can then attach to a later unrelated agent step. Please handle the no-output end path and add a focused regression before merging.

crates/core/src/observability/atif.rs
willkill07, line 1: The changes seem overly specific for OpenAI Responses.

I don't see handling or coverage for Anthropic Messages

crates/core/src/observability/atif.rs
mnajafian-nv: Thanks Eric! Fixed :)

crates/core/src/observability/atif.rs
mnajafian-nv: Thanks, Will! I generalized the continuation handling so this now covers the Anthropic raw-messages continuation path too, not just OpenAI Responses. I also added a regression for the Anthropic tool-result case and updated the PR description to match :)

ericevans-nv (approved):

willkill07 (approved):

PR #546, opened 2026-07-23
feat(sanitizers)!: expose per-call codec capabilities

### LINKED ISSUE #526: [Enhancement]: Make PII redaction use the active per-call provider codec
## Affected area

* Rust core runtime
* Middleware or guardrails
* Plugins
* Observability or exporters
* Documentation or examples

## Problem or opportunity

The built-in `pii_redaction` component currently accepts one component-wide `codec` value. Every profile in `components.config.profiles` inherits that same codec, and the component is a singleton, so a configuration cannot install one codec-aware PII policy per provider surface.

Relay can handle managed LLM calls for multiple provider APIs in the same runtime. The gateway now resolves and installs the matching request, response, and streaming codecs per route, including Anthropic Messages, OpenAI Chat Completions, and OpenAI Responses. The PII sanitizer does not consume that active per-call codec. Instead, it independently constructs one fixed request and response codec from `plugins.toml`.

As a result, mixed-provider deployments must select one nominal PII codec:

* Calls using the selected provider get codec-aware normalization.
* Calls using another provider may fall back to raw JSON traversal when decoding fails.
* Provider schemas can overlap, so a mismatched codec can partially decode a request instead of taking the fallback path.
* Exact normalized `target_paths` cannot be guaranteed across every supported provider API.
* Adding profiles per codec does not solve the problem because profiles cannot select a codec and every enabled profile runs on every call.

Provider-agnostic actions such as all-leaf detector scans and the `trajectory_context` preset can operate on raw/provider-shaped data, but configuration validation still requires one codec whenever LLM surfaces are enabled.

## Proposed enhancement

Make PII redaction use the active codec or provider surface associated with each managed LLM call instead of requiring one fixed codec for the entire component.

The exact API can follow existing runtime ownership boundaries. Possible approaches include:

* Pass a stable provider-surface identifier and the existing normalized annotation into context-aware sanitize callbacks.
* Add an internal context-aware sanitizer registration path for first-party components.
* Resolve the correct stateless codec from the active `ProviderSurface` at invocation time.

Suggested behavior:

1. When a managed call has an active request or response codec, PII redaction uses that codec for the corresponding call.
2. A configured legacy `codec` remains available as a compatibility fallback for callers that do not provide per-call codec context.
3. Calls without codec context use an explicit, documented raw-JSON fallback only where the configured policy is provider-agnostic.
4. Policies that require normalized target paths report a clear diagnostic when neither per-call codec context nor a compatible legacy codec is available.
5. Codec resolution or decoding failures use privacy-safe behavior and do not silently publish unsanitized content when the component claims coverage.
6. Profile priority, array-order tie-breaking, scope ownership, subscriber ordering, and rollback semantics remain unchanged.

Avoid passing Rust codec trait objects across native or worker plugin boundaries. If the shared callback contract changes, expose stable JSON/string-compatible provider context across bindings and plugin protocols.

## Runtime contract and binding impact

The core managed LLM runtime already owns the active request, response, and streaming codecs. This enhancement should preserve that ownership and make the relevant stable context available to the PII sanitization boundary.

The serialized PII configuration remains `snake_case` across Rust, Python, Node.js, and experimental bindings. Existing configurations with an explicit component-wide `codec` must continue to work.

For new configurations, the fixed codec should become optional when the runtime supplies per-call codec context or when every enabled policy is explicitly provider-agnostic. CLI editor, JSON Schema, language helpers, and documentation should expose the same semantics.

Buffered and streaming calls must apply the same provider selection and sanitization rules.

## Alternatives considered

* Configure one Relay process per provider API: works, but prevents one runtime from safely serving multiple supported provider routes under a shared privacy policy.
* Add one PII profile per codec: profiles share the component-wide codec and all profiles run on every call.
* Add multiple top-level `pii_redaction` components: unsupported because the component is a singleton and would still require per-call routing predicates.
* Rely on mismatched-codec decode failure and raw fallback: provider schemas overlap, so failure is not deterministic and normalized target paths are not portable.
* Omit `codec` and downgrade validation errors to warnings: useful for experimentation with provider-agnostic actions, but not an appropriate production privacy contract.
* Use raw middleware: possible, but loses the first-party typed configuration, validation, editor support, detector presets, and cross-runtime behavior.

## Acceptance criteria

- [ ] One enabled `pii_redaction` component can sanitize mixed managed calls using every codec returned by `supported_codec_names()`.
- [ ] Each request, buffered response, and streaming finalization uses the codec associated with that specific call.
- [ ] A codec selected for one provider is never applied to a call whose active provider surface is different.
- [ ] Exact normalized `target_paths` behave consistently across Anthropic Messages, OpenAI Chat Completions, and OpenAI Responses.
- [ ] Provider-agnostic policies can omit the fixed component codec without weakening unrelated configuration validation.
- [ ] Calls without active codec context have documented, privacy-safe fallback or failure behavior.
- [ ] Existing configurations with an explicit component-wide codec remain compatible.
- [ ] Tests cover sequential and concurrent mixed-provider calls, overlapping request shapes, buffered responses, streaming responses, decode failures, and absence of raw PII in emitted events.
- [ ] CLI editor/schema support and relevant binding helpers expose the updated configuration contract.
- [ ] PII redaction documentation includes one mixed-provider configuration example and migration guidance.

Relates to NVIDIA/NeMo-Relay#245, NVIDIA/NeMo-Relay#301, NVIDIA/NeMo-Relay#396, NVIDIA/NeMo-Relay#509, NVIDIA/NeMo-Relay#512, NVIDIA/NeMo-Relay#513, and NVIDIA/NeMo-Relay#525.

willkill07:
#### Overview

Exposes the active per-call LLM codec to observability sanitizers across in-process and plugin boundaries, then uses that resolved capability to make PII redaction provider-correct and fail-closed for mixed managed LLM traffic.

> \[!WARNING\]<br>**BREAKING CHANGE:** LLM request and response sanitizers now use the required two-argument contract `(payload, directional_context) -> optional payload`. No registration-time compatibility shim remains. Rust, Python, Node.js, native plugins, raw C FFI consumers, worker plugins, and downstream bindings must update their callbacks. All Rust worker sanitizer callbacks—mark, scope, tool, and LLM—are now async. Native dynamic plugins remain ABI v2 and workers remain `grpc-v1`, but the v2 context/host tables and grpc-v1 invocation/service definitions change in place.

#### Details

* Replaces split legacy/contextual LLM sanitizer registration with one canonical request and response callback representation while preserving priority, array-order tie breaking, scope ownership, registration rollback, snapshots, and short-circuit omission.
* Introduces distinct `LlmSanitizeRequestContext` and `LlmSanitizeResponseContext` values with structured codec identities: none, built-in, runtime, and opaque. Every active codec, including opaque codecs, can be resolved for the lifetime of the callback.
* Gives request sanitizers `decode(request)` and `encode(annotated, original)`; gives response sanitizers `decode(response)`.
* Extends native ABI v2 and the raw C FFI with borrowed directional codec handles and host-owned transformation operations. Safe wrappers reject retained handles after the callback returns, contain codec panics, preserve ownership for aliased results, and fail closed for malformed output or unrepresentable runtime codec IDs.
* Extends `grpc-v1` with directional contexts and authenticated, invocation-scoped codec capabilities. Host RPCs reject forged, expired, unauthorized, and wrong-direction capability IDs, and SDK authors see async codec proxies rather than raw IDs.
* Makes Rust worker sanitizer registration async-only while Python workers accept immediate values or awaitables under the same required two-argument callback contract.
* Keeps sanitizer failures fail-closed for observability: failures or `None`/`null` results omit the payload and annotation without changing the client-visible request or response.
* Invalidates caller-supplied normalized annotations whenever sanitization changes a buffered payload, rebuilding only through the active directional codec. Decode failure or no codec omits the stale annotation.
* Updates PII redaction to use the resolved active codec per call, use configured `codec` only when no active codec exists, support provider-agnostic policies without a configured codec, and omit codec-dependent observability data when normalization is unavailable.
* Hardens normalized target overlays: object-field removals are applied incrementally, unsafe array removals and inconsistent projections fail closed, and OpenAI Responses keeps the top-level `output_text` alias synchronized with sanitized structured output.
* Makes Node custom-codec reference cleanup event-loop-affine so early-dropped streams cannot delete N-API references from a Tokio thread.
* Documents parameter order, directional contexts, codec identity and resolution, capability lifetime, omission behavior, native/worker contracts, and mixed-provider PII usage with Fern tabs. Adds the 0.7 breaking-change release note and a complete 0.6-to-0.7 migration procedure.

Validation:

* Repository-wide pre-commit suite passed before the final review fixes; the final changed-file pre-commit suite also passed, including FFI header sync, Cargo fmt/Clippy/check, Node formatting, and documentation links.
* Node: 290 tests passed.
* Python: 556 tests passed.
* Go: all package tests passed.
* Raw FFI: 82 unit tests and 77 integration tests passed.
* Targeted buffered and streaming OpenAI Responses PII regressions passed and verified serialized events contain no original secret.
* The public Rust native-plugin example passes its standalone `cargo check`.
* The affected Rust core, PII, native-plugin, worker protocol/SDK, and binding capability suites passed. One unrelated CLI test in the workspace was contaminated by an ancestor user plugin configuration; the exact test passed from an isolated checkout of this commit.
* Documentation build completed with 0 errors.

Start with `crates/core/src/api/runtime/callbacks.rs` and `crates/core/src/api/llm.rs` for the canonical callback contract, codec propagation, and annotation invalidation. Then review `crates/plugin/src/lib.rs`, `crates/core/src/plugin/dynamic/native.rs`, `crates/core/src/plugin/dynamic/worker.rs`, and `crates/worker-proto/proto/nemo/relay/worker/v1/plugin_worker.proto` for cross-boundary capability lifetime and authorization. Finish with `crates/pii-redaction/src/builtin.rs` and `crates/pii-redaction/src/overlay.rs` for per-call codec selection and fail-closed redaction, then review `docs/reference/migration-guides.mdx` for the public upgrade contract.

* Closes NVIDIA/NeMo-Relay#526
* Closes NVIDIA/NeMo-Relay#526

## Summary by CodeRabbit

* **New Features**
  * Codec-aware LLM sanitizers: callbacks now receive per-call request/response codec context for directional normalization/redaction.
  * Sanitizers can omit observability payloads/annotations by returning no value (`null`/`None`).
  * Codec-aware support expanded across Node.js, Python, Go, Rust, native plugins, and worker integrations.
  * PII redaction now selects the active codec per call for mixed-provider traffic.
* **Breaking Changes**
  * Callback signatures changed to `(payload, context)` and must return an optional payload (`null`/`None` to omit).
  * Native plugins now require Native ABI v2 (rebuilt with the matching SDK/protocol).
* **Documentation**
  * Added migration guidance and updated middleware/plugin docs for codec-aware sanitizer behavior.


relda88 (commented):
Good improvement overall. One edge case: empty string vs null aren't treated the same downstream — might need a normalization step.

afourniernv:
I understand both large changes now:

1. Sanitizers use the codec attached to each individual call instead of the plugin being tied to one configured codec.
2. Worker sanitizers can use that host-owned codec asynchronously over gRPC.

```text
Relay host process                 PII worker process
------------------                 ------------------
Owns the active provider codec     Owns the redaction policy
```

Since these are separate processes, the worker cannot call the codec directly or access its memory. The flow is:

1. Relay sends the payload to the PII worker.
2. The worker asks Relay to decode it using the active codec.
3. Relay returns the normalized content.
4. The worker redacts the content.
5. The worker asks Relay to encode it back into the provider shape.
6. Relay returns the sanitized provider payload.
7. The worker returns that payload.
8. Relay emits the sanitized event.

That lets one isolated worker handle mixed providers and custom codecs without copying provider-specific codec logic into the worker.

The async part matters because decode and encode are gRPC round trips to another process, not local function calls. `await` still waits for each result; it is not fire-and-forget. It lets the worker release its thread while the Relay host responds, so the worker can service other invocations instead of blocking on codec operations. Relay still waits for the final sanitized payload before emitting the event.

crates/core/src/api/llm.rs
```diff
@@ -428,36 +430,42 @@ fn emit_llm_start_with_subscribers(
...
+    let mut annotated_request = match (sanitized_request.as_ref(), request_codec.as_deref()) {
+        (Some(sanitized_request), Some(codec))
             if sanitized_request.headers != request.headers
                 || sanitized_request.content != request.content =>
         {
-            codec.decode(&sanitized_request).ok().map(Arc::new)
+            codec.decode(sanitized_request).ok().map(Arc::new)
         }
-        _ => annotated_request,
+        (Some(_), _) => annotated_request,
```
afourniernv: I reproduced a manual-lifecycle case where the sanitizer redacts `event.input`/`event.output`, but a caller-provided `annotated_request` or `annotated_response` still contains the original secret. On the request path, this branch preserves the annotation when no codec is available. The response path has the same issue because `resolve_llm_end_annotation` returns the supplied annotation before decoding the sanitized data. Since normalization/exporters prefer the annotation, the serialized event still leaks the original value through `category_profile`.

Can we invalidate or regenerate the annotation whenever sanitization changes the payload? Without an active codec, clearing the annotation is safer than retaining the pre-sanitization value.

crates/pii-redaction/src/builtin.rs
```diff
@@ -286,16 +285,83 @@ impl CompiledBuiltinBackend {
...
+        sanitized_annotated: AnnotatedLlmRequest,
+    ) -> Option<LlmRequest> {
+        let sanitized = serde_json::to_value(sanitized_annotated).ok()?;
+        let mut sanitized_request = request.clone();
+
+        for target_path in self.target_paths.iter() {
+            let target_segments = json_pointer_segments(target_path)?;
+            let Some(target_value) = sanitized_json_pointer_value(&sanitized, &target_segments)
+            else {
+                continue;
```
afourniernv: I reproduced a concrete `action = "remove"` leak here. With `/messages/0/name` and `/messages/1/name`, the full codec encode correctly fails on multi-item alignment. The sanitized annotation omits both fields, so this branch skips both paths and the fallback returns `Some(original_request)` with both names still present.

For `remove`, absence from the sanitized value means deletion, not necessarily that the configured target was missing. Could the fallback distinguish absent-in-both from removed, apply object deletion where possible, and return `None` if it cannot safely re-encode?

afourniernv (commented):
few comments -

willkill07:
@coderabbitai review

willkill07:
@coderabbitai review

crates/core/src/api/llm.rs
willkill07: Done

crates/pii-redaction/src/builtin.rs
willkill07: Done

afourniernv (approved):

PR #555, opened 2026-07-24
feat: add size-based operational log rotation

### LINKED ISSUE #553: [Enhancement]: Add size-based rotation and retention for operational log file sinks
## Affected area

- Rust core runtime
- Documentation or examples

## Problem or opportunity

NeMo Relay operational file sinks currently append to the configured file without rotation or retention. The asynchronous queue bounds pending records in memory, but it does not bound the size of the persisted log file.

Long-running Relay processes can therefore create continuously growing files that require external cleanup and may eventually consume available disk space.

## Proposed enhancement

Add optional size-based rotation and retention settings to operational file sinks.

Users should be able to configure:

- The maximum size of the active log file.
- The maximum number of rotated files retained.

When the active file reaches the configured size, Relay should rotate it and remove files beyond the retention limit. Existing asynchronous delivery, periodic flushing, stderr logging, formatting, filtering, and root Relay ID behavior should remain unchanged.

When rotation is not configured, existing append-only behavior should remain the default for backward compatibility.

This enhancement applies only to Relay operational file sinks. Remote logging destinations and agent telemetry exporters are separate concerns.

## Runtime contract and binding impact

Rotation should be implemented by the core logging runtime so every frontend using the same `LoggingConfig` and `LoggingRuntime` receives consistent behavior.

No Python, Node.js, Go, or C FFI API changes are required. Existing processes continue using append-only files unless rotation is explicitly configured.

A rotation failure should use the existing sink-error reporting path to stderr and should not crash an otherwise healthy Relay process.

## Alternatives considered

External tools such as `logrotate`, container logging drivers, or service managers can manage log retention. However, they are not always available for embedded applications, local development, or directly configured Relay file sinks. They may also require coordination with the process's open file handle.

Truncating the file at startup would limit history between runs but would not prevent unbounded growth during a long-running process.

## Acceptance criteria

- A file sink can optionally configure a maximum active-file size and retained-file count.
- Reaching the configured size rotates the active file without blocking Relay's operational path.
- Files exceeding the configured retention count are removed.
- Omitting rotation settings preserves the existing append-only behavior.
- Invalid values, including zero or unsafe limits, return clear configuration errors.
- Rotation failures are reported through the existing stderr sink-error path without crashing Relay.
- Focused tests cover rotation, retention, append-only compatibility, asynchronous flushing, and shutdown.
- The operational logging documentation explains the new configuration and behavior.

This extends the append-only sink introduced by #413 without overlapping the operational event instrumentation completed in #455.

ericevans-nv:
#### Overview

Add optional size-based rotation and retention to Relay operational file-log sinks. File sinks remain append-only unless rotation is explicitly configured.

#### Details

* Add validated `max_file_size_bytes` and `retained_files` settings for file sinks.
* Require both rotation settings together and reject zero or unsafe retention values.
* Rotate complete log records before the next write would exceed the configured size.
* Retain the configured number of managed backups while preserving the active file.
* Preserve historical backups outside the current retention window when the configured limit is reduced, avoiding automatic deletion of existing data.
* Keep file writes behind the existing asynchronous sink and report rotation failures through its error path.
* Detect conflicts between active sink paths and generated backup paths.
* Preserve existing append-only behavior when rotation is omitted.
* Document the configuration and add focused core and CLI coverage.

Start with the rotating writer in `crates/core/src/logging/rotation.rs` and its integration with the asynchronous file sink in `crates/core/src/logging/sink.rs`. Focused behavior coverage is in `crates/core/tests/coverage/logging_tests.rs`.

* Closes NVIDIA/NeMo-Relay#553

## Summary by CodeRabbit

* **New Features**
  * Added size-based rotation for operational file log sinks using `max_file_size_bytes` and `retained_files` (with a `retained_files` max of 9).
  * When enabled, log writing rotates by size while keeping records intact and appending to the active file.
* **Bug Fixes**
  * Enforced that rotation settings must be configured together; invalid values now fail configuration.
  * Prevented collisions between active log paths and any potential rotated/retained file targets.
* **Documentation**
  * Updated the operational logging TOML examples and guidance for rotation configuration and limits.
* **Tests**
  * Expanded parsing and integration coverage for rotation, retention ordering, boundary rotation, large-record behavior, and collision scenarios.

### DIFF
--- a/crates/core/src/logging/config.rs
+++ b/crates/core/src/logging/config.rs
@@ -28,6 +28,13 @@ pub const DEFAULT_FILE_FLUSH_INTERVAL_MILLIS: u64 = 1000;
 /// configuration above this bound is rejected with a config error. It cannot be raised.
 pub const MAX_FILE_SINK_QUEUE_ENTRIES: usize = 8_192;
 
+/// Fixed hard maximum number of retained backup files per rotating file sink.
+///
+/// Size-based rotation renames existing backup files on each rotation, so an unbounded value can
+/// make one log write perform excessive filesystem work. This limit counts backup files and does
+/// not include the active log file.
+pub const MAX_FILE_SINK_RETAINED_FILES: usize = 100;
+
 /// Operational logging configuration for [`LoggingRuntime::configure`](super::LoggingRuntime::configure).
 ///
 /// `level` is the process-wide **minimum severity**: call sites may emit any level, but records
@@ -219,6 +226,8 @@ pub struct FileLogSinkConfig {
     /// Maximum pending asynchronous queue entries for this file sink. Must be greater than 0 and
     /// at most [`MAX_FILE_SINK_QUEUE_ENTRIES`].
     pub queue_capacity: usize,
+    /// Optional size-based rotation and retention settings.
+    pub rotation: Option<FileLogRotationConfig>,
 }
 
 impl Default for FileLogSinkConfig {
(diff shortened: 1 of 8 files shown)


crates/core/src/logging/config.rs
```diff
@@ -28,6 +28,13 @@ pub const DEFAULT_FILE_FLUSH_INTERVAL_MILLIS: u64 = 1000;
 /// configuration above this bound is rejected with a config error. It cannot be raised.
 pub const MAX_FILE_SINK_QUEUE_ENTRIES: usize = 8_192;
 
+/// Fixed hard maximum number of retained backup files per rotating file sink.
+///
+/// Size-based rotation renames existing backup files on each rotation, so an unbounded value can
+/// make one log write perform excessive filesystem work. This limit counts backup files and does
+/// not include the active log file.
+pub const MAX_FILE_SINK_RETAINED_FILES: usize = 100;
```
willkill07: 100 feels like a lot?

Don't you think 9 is enough?

willkill07 (approved):
Just one question on maximum files. with rotation naming, `.0` -> `.9` makes sense (for up to 10) but anything beyond that isn't default sorted.

crates/core/src/logging/config.rs
ericevans-nv: Agreed—nine retained backups is sufficient. I reduced the hard limit to 9 in c5d19bb6 and updated the documentation. This keeps the current `_1` through `_9` backup names naturally sortable and further bounds the filesystem work performed during rotation.

crates/core/src/logging/rotation.rs
```diff
@@ -0,0 +1,148 @@
...
+        let destination = rotated_log_path(base_path, index);
+        fs::rename(source, destination)?;
+    }
+    Ok(())
+}
+
+pub(crate) fn rotated_log_path(base_path: &Path, index: usize) -> PathBuf {
+    let stem = base_path.file_stem().unwrap_or(base_path.as_os_str());
+    let mut file_name = stem.to_os_string();
+    file_name.push(format!("_{index}"));
```
willkill07: One other nit: I've usually seen `.` used rather than `_` as the separator`. On linux systems, I've seen it be a suffix (e.g. `file.jsonl.1`) but I personally prefer it as you have it (immediately before the extension `file.1.jsonl`)

willkill07 (approved):

mnajafian-nv (approved):
LGTM, plus Will's nit

crates/core/src/logging/rotation.rs
ericevans-nv: Good point. I updated rotated backup names in ebcc9e27 to use `.` before the index, such as `relay.log.1.jsonl`, while preserving the original extension.

willkill07 (approved):

PR #565, opened 2026-07-27
feat: add structured config editor

willkill07:
#### Overview

Add a structured interactive editor for the supported non-agent Relay configuration.

#### Details

- Add `nemo-relay config edit` with user (default), project, and global target scopes.
- Edit gateway limits, provider upstreams, and operational logging file sinks while preserving unrelated TOML and redacting auth headers.
- Document command scope and credential guidance, and add parser, model, validation, and safe non-TTY behavior coverage.

Start with `crates/cli/src/commands/configure/editor.rs`; it contains the TOML-backed editor model and the interactive menus. Then review the command dispatch and focused tests.

- Relates to: RELAY-493

## Summary by CodeRabbit

* **New Features**
  * Added an interactive `nemo-relay config edit` editor for gateway limits, upstream/provider settings, operational logging, and file sinks, with TTY-only editing and scope selection (`--user`, `--project`, `--global`).
  * Updated global-scope saves to refuse upstream authorization-header credentials; added safer global plugin config handling with system-readable file permissions.
* **Documentation**
  * Expanded “Edit Gateway Configuration” guidance, including save/creation behavior, section preservation, and credential handling via environment variables.
* **Tests**
  * Added CLI parsing, editor validation/preview redaction, atomic write guarantees, and global `plugins.toml` permission/secret rejection coverage.

### DIFF
--- a/crates/cli/src/commands/configure/editor.rs
+++ b/crates/cli/src/commands/configure/editor.rs
@@ -39,7 +39,8 @@ impl From<&ConfigEditCommand> for TargetScope {
 
 pub(super) fn edit(command: ConfigEditCommand) -> Result<(), CliError> {
     ensure_tty()?;
-    let path = target_path(TargetScope::from(&command))?;
+    let scope = TargetScope::from(&command);
+    let path = target_path(scope)?;
     let mut document = ConfigDocument::read(path)?;
     let theme = ColorfulTheme::default();
 
@@ -63,7 +64,7 @@ pub(super) fn edit(command: ConfigEditCommand) -> Result<(), CliError> {
             2 => edit_logging(&theme, &mut document)?,
             3 => print_preview(&document),
             4 => {
-                document.write()?;
+                document.write(scope)?;
                 println!("  ✓ Saved {}", document.path().display());
                 return Ok(());
             }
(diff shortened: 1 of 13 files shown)


crates/cli/src/commands/configure/editor.rs
```diff
@@ -0,0 +1,971 @@
...
+        };
+        Ok(Self { path, document })
+    }
+
+    fn path(&self) -> &Path {
+        &self.path
+    }
+
+    fn write(&self) -> Result<(), CliError> {
+        crate::filesystem::atomic_write_private(&self.path, self.document.to_string().as_bytes())
```
ericevans-nv: `--global` appears intended to provide machine-wide configuration: an administrator writes it with `sudo`, and ordinary Relay processes read it through automatic `/etc/nemo-relay/config.toml` discovery. However, this method always uses the private `0600` policy, so a global save creates a root-owned file that ordinary users cannot read. Could global writes use a scope-appropriate system-readable policy, such as `0644` or controlled `0640`, while keeping authorization headers out of that file? If global configuration is intentionally root-only, could the documentation state that Relay must also run as root?

crates/cli/src/plugins/config_io.rs
```diff
@@ -170,11 +170,20 @@ impl PluginConfigDocument {
...
-        if let Some(parent) = self.path.parent() {
-            std::fs::create_dir_all(parent)?;
+        crate::filesystem::atomic_write(&self.path, rendered.as_bytes()).map_err(CliError::Config)
+    }
+
+    pub(crate) fn write_for_scope(&self, scope: TargetScope) -> Result<(), CliError> {
+        let rendered = self.render()?;
+        match scope {
+            TargetScope::Global => {
+                crate::filesystem::atomic_write_system_readable(&self.path, rendered.as_bytes())
```
ericevans-nv, line 180: Could we reject persisted secrets in global plugin configuration before writing it as `0644`? The editor restores schema-declared password values before saving, so `plugins edit --global` can replace an existing private `plugins.toml` with a world-readable file containing those values.

ericevans-nv (approved):
Approving with one security follow-up noted inline: global plugin saves should reject persisted secrets before applying the system-readable `0644` policy.

crates/cli/src/plugins/config_io.rs
willkill07: done!

ericevans-nv (approved):

PR #566, opened 2026-07-28
fix: make doctor report invalid explicit config

willkill07:
#### Overview

#### Details

- Route top-level `--config` overrides into `nemo-relay doctor` so it reports the selected runtime and sibling plugin configuration.
- Fall back to safe default logging while Doctor reports missing or malformed configuration.
- Add CLI and report regressions for missing/invalid explicit config and focused plugin Doctor behavior.
- Document inspecting an explicit configuration path with Doctor.

- `crates/cli/src/diagnostics/mod.rs`, which turns configuration resolution failures into a normal Doctor Resolution check while applying the selected configuration scope.

- Fixes RELAY-582

## Summary by CodeRabbit

- **New Features**
  - `doctor` now honors an explicitly targeted `--config`, including when that config is missing or invalid.
  - Diagnostics now scope plugin TOML discovery to the explicit config’s sibling `plugins.toml`, and plugin checks can run even if workspace runtime config is unusable.
  - JSON output includes richer configuration-resolution details, including affected/missing paths.
- **Bug Fixes**
  - Doctor-specific logging config resolution now falls back gracefully on failure and emits a warning.
  - Configuration-layer reporting adjusts when `--config` is provided, and the UI label switches to “Explicit” when applicable.
- **Documentation**
  - Updated “Getting Started” troubleshooting with an example `nemo-relay doctor --config /path/to/config.toml doctor`.

### DIFF
--- a/crates/cli/README.md
+++ b/crates/cli/README.md
@@ -104,6 +104,13 @@ After setup, inspect local readiness:
 nemo-relay doctor
 ```
 
+To troubleshoot a specific configuration file, pass it explicitly. Doctor reports a missing or
+invalid file in its configuration checks instead of stopping before diagnostics:
+
+```bash
+nemo-relay --config /path/to/config.toml doctor
+```
+
 Run a supported agent through the gateway:
 
 ```bash
--- a/crates/cli/src/commands/diagnostics.rs
+++ b/crates/cli/src/commands/diagnostics.rs
@@ -29,11 +29,20 @@ pub(crate) struct AgentsCommand {
     pub(crate) json: bool,
 }
 
-pub(super) async fn execute(command: DoctorCommand) -> Result<ExitCode, CliError> {
+pub(super) async fn execute(
+    command: DoctorCommand,
+    server: &super::serve::ServerArgs,
+) -> Result<ExitCode, CliError> {
     if let Some(plugin) = command.plugin {
         return execute_plugin_doctor(plugin, command.install_dir, command.json);
     }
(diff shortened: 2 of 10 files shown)


bbednarski9 (approved):

bbednarski9 (changes requested):
One codex comment looks worth fixing...

[P1] --config doesn’t scope the runtime-layer diagnostics
Doctor correctly resolves the explicit file, but [collect_configuration](https://github.com/NVIDIA/NeMo-Relay/blob/490af3b985295ced7235d0bcb034f6b7ca546a15/crates/cli/src/diagnostics/mod.rs#L136-L172) still checks the implicit workspace, global, and system files. [layer_status](https://github.com/NVIDIA/NeMo-Relay/blob/490af3b985295ced7235d0bcb034f6b7ca546a15/crates/cli/src/diagnostics/mod.rs#L263-L297) marks an existing file active or failed, and those failures directly affect the [exit code](https://github.com/NVIDIA/NeMo-Relay/blob/490af3b985295ced7235d0bcb034f6b7ca546a15/crates/cli/src/diagnostics/render.rs#L8-L23).

Consequently:
An invalid .nemo-relay/config.toml can make nemo-relay --config /valid/explicit.toml doctor fail, even though the runtime ignores the workspace file.
Valid implicit files are displayed as (loaded) even though explicit configuration disables their discovery.
The actual explicit runtime file is not displayed as a configuration layer.
The runtime configuration layers need the same explicit-path scoping added for plugins.toml, with a regression test containing an invalid ignored workspace config.

bbednarski9 (approved):

Issue #446, opened 2026-07-15
[Enhancement]: Add MCP and tool ID for tool-call provenance and correlation


bbednarski9:
## Affected area

* Rust core runtime
* Python binding
* Middleware or guardrails
* Plugins
* Observability or exporters
* Framework integrations
* Documentation or examples

**Proposed target release:** 0.7

Related foundational proposals:

* NVIDIA/NeMo-Relay#444: seed isolated scope stacks from external parent context
* NVIDIA/NeMo-Relay#445: add a provider-neutral annotated tool-result contract

## Problem or opportunity

Many applications execute tools through MCP clients or harness-native abstractions rather than Relay managed-tool APIs. The agent and tool server may run in different processes with independent Relay runtimes.

Today there is no reusable adapter pattern that preserves both:

* Tool-call semantics such as name, arguments, call ID, structured result, and error status.
* Causal context linking the remote tool execution to the initiating agent or LLM scope.

LLM request codecs can recognize provider tool-use structures after they have been encoded into a model request, but they cannot reconstruct execution metadata or distributed parentage that the harness has already discarded.

Relay should provide adapter APIs and reference mappings. It should not own application transport or operate as a mandatory MCP proxy.

## Proposed enhancement

Provide reusable MCP and harness adapter helpers that map external tool protocols into Relay lifecycle and annotation APIs.

For MCP, the adapter should preserve:

* The tools/call JSON-RPC request ID.
* Tool name and arguments.
* CallToolResult content, structuredContent, isError, and allowed metadata.
* Protocol errors separately from tool execution errors.
* An explicitly supplied external parent context using the API proposed in NVIDIA/NeMo-Relay#444.
* An annotated result using the contract proposed in NVIDIA/NeMo-Relay#445.

Provide at least one reference integration demonstrating an agent-side caller and a separately instrumented tool server. The caller or application chooses an authenticated transport for correlation context, such as application headers or protocol metadata.

Adapters must apply existing Relay sanitization and must not automatically trust externally supplied parent identifiers or arbitrary metadata.

## Runtime contract and binding impact

Core should expose transport-neutral building blocks. Rust and Python should receive reference adapter support first because they are common MCP integration surfaces. Other bindings can adopt the same serialized contracts incrementally.

An application that does not use MCP should be able to implement the same adapter pattern for its harness-specific tool abstraction without depending on MCP packages.

## Alternatives considered

[...]

PR #575, opened 2026-07-28
feat!: add canonical tool execution results

### LINKED ISSUE #445: [Enhancement]: Standardize tool execution results across Relay
## Goal

Establish a consistent tool-execution result contract across Relay.

Tool results should contain the application-visible result together with optional opaque annotation metadata. The contract should behave consistently across in-process, worker, and native plugin surfaces.

## Expected behavior

- Tool-result annotations remain separate from the application-visible result.
- Middleware can preserve, replace, or remove annotations.
- Security and observability components can process the result and annotation independently.
- Existing continuation, error, short-circuit, and pending-mark behavior remains consistent.
- Incompatible plugin and Relay versions are rejected with clear migration guidance.

## Compatibility

This introduces a breaking semantic contract for tool-execution results. The change must have a documented Relay compatibility boundary, migration guidance, and coverage across all supported plugin surfaces.

## Acceptance criteria

- All tool-execution paths use the standardized result contract.
- Result and annotation behavior is consistent across plugin boundaries.
- Sanitization and export behavior are tested independently.
- Compatibility validation prevents mixing legacy and standardized result semantics.
- The migration and compatibility expectations are documented.

bbednarski9:
#### Overview

This PR makes one canonical tool-result contract apply across the runtime, bindings, plugins, and observability surfaces.

> [!WARNING]
> **BREAKING CHANGE:** **[Tool execution APIs and plugin wire contracts]** Tool producers, execution-intercept continuations, managed execution returns, and manual completion no longer exchange raw JSON; they now use the canonical `ToolExecutionResult { result, annotation }` contract. Update forwarding intercepts to preserve or deliberately replace/remove `annotation`, and update Rust, Python, Node.js, C/FFI, and Go call sites to the new result shape. Relay 0.8 resets the tool-result semantics under `native_api = "1"` and `worker_protocol = "grpc-v1"`. Rebuild every native and worker plugin and declare `compat.relay` so it excludes versions before 0.8 (recommended: `>=0.8.0,<1.0`). Native ABI v4 layouts remain unchanged. Workers retain the `grpc-v1` identifier and `nemo.relay.worker.v1` package, service names, and RPC method names, but the protobuf contract changes: `RelayHostRuntime.ToolNext` now returns `ToolExecutionResultResponse`, and `ToolExecutionInterceptResult.outcome` is a typed `ToolExecutionInterceptOutcome` instead of a `JsonEnvelope`. Regenerate custom protobuf bindings before rebuilding workers.

Tool producers, execution continuations, managed execution, and manual completion now exchange:

```rust
ToolExecutionResult {
    result: Json,
    annotation: Option<Json>,
}
```

Tool execution intercepts return the flat companion contract:

```rust
ToolExecutionInterceptOutcome {
    result: Json,
    annotation: Option<Json>,
    pending_marks: Vec<PendingMarkSpec>,
}
```

#### Details

- Replaces raw JSON tool results with `ToolExecutionResult` in Rust, Python, Node.js, C/FFI, Go, native plugins, and the Rust/Python worker SDKs.
- Keeps `pending_marks` Relay-owned while allowing each execution intercept to explicitly preserve, replace, remove, or short-circuit the opaque annotation.
- Normalizes missing and JSON-null annotations to absence without interpreting either the application-owned `result` or `annotation` value.
- Preserves complete MCP-style results—including `content`, `structuredContent`, `_meta`, and `isError`—inside `result`; Relay annotations remain adjacent to, and independent from, MCP `_meta`.
- Tool response sanitizers receive and sanitize only `result`; they cannot inspect or rewrite `annotation`. Relay then attaches each non-null annotation at `category_profile.tool_result_annotation` before the general scope-end event-sanitizer chain, which can replace or remove `data`, `category_profile` (including the annotation), and event `metadata` before subscribers and exporters receive the event.
- The first-party PII redaction plugin's built-in action policies apply `tool_output` to `result` and the projected annotation as independent JSON documents. `target_paths` are relative to each document; root removal omits the annotation key, while other typed tool profile fields remain unchanged.
- Sanitizers affect observability only; Relay returns the original, unsanitized `ToolExecutionResult` to the application.
- Projects annotations through ATOF, ATIF, full OpenTelemetry, and OpenInference as one opaque value. The GenAI projection intentionally omits the Relay-specific attribute.
- Keeps native plugin ABI v4 layouts and callback signatures unchanged while making canonical tool-result JSON the Relay 0.8 native API 1 baseline. Native tool-result envelopes continue to use the versioned schema tags defined beside the shared DTOs.
- Keeps the `grpc-v1` protocol identifier and `nemo.relay.worker.v1` package, service names, and RPC method names, while structurally defining the `ToolExecutionResult` and `ToolExecutionInterceptOutcome` wrappers in protobuf. Their application-owned result, annotation, and pending-mark array values use lossless `JsonValue` bytes rather than `google.protobuf.Value` so arbitrary JSON and integer precision are preserved.
- Changes `RelayHostRuntime.ToolNext` from `JsonResult` to `ToolExecutionResultResponse` and `ToolExecutionInterceptResult.outcome` from `JsonEnvelope` to typed `ToolExecutionInterceptOutcome`. All workers must regenerate bindings and rebuild for the Relay 0.8 baseline.
- Requires dynamic-plugin manifests and persisted records to declare a `compat.relay` range that excludes pre-0.8 Relay releases. Plugin loading separately retains the existing check that the range matches the running host.
- Rejects legacy raw producer results on public and dynamic-plugin boundaries rather than maintaining parallel frame APIs or ambiguous adapters.
- Leaves managed `tool_call_id` behavior unchanged in this PR; #446 remains a focused follow-up.

Breaking surfaces include callback return values, execution-intercept `next`, managed execute return values, manual tool completion input, and the two worker protobuf type seams above. Native plugin ABI v4 itself does not change.

Focused validation completed on the current implementation:

- `cargo check -p nemo-relay-worker-proto -p nemo-relay-worker -p nemo-relay --features worker-grpc --locked`
- `cargo clippy -p nemo-relay-worker-proto -p nemo-relay-worker -p nemo-relay --all-targets --features worker-grpc -- -D warnings`
- `cargo test -p nemo-relay-worker-proto`; `cargo test -p nemo-relay-worker --test worker_sdk_tests`; and focused host-worker continuation/error tests passed, including structural wrappers, lossless JSON, null normalization, malformed payloads, and pending marks
- Python worker SDK focused protocol and descriptor tests: 21 passed; Ruff passed
- Python worker example: 4 passed; added a managed-environment regression for `ToolNext`, result mutation, annotation preservation, and pending marks. The test compiled locally and runs end to end when the lifecycle-managed Python environment is provisioned
- Pinned `grpcio-tools` generation verified the `ToolNext`, tool-result, and outcome descriptors
- Fern documentation checks and strict broken-link validation passed
- `git diff --check`

The local fixture rebuild and full matrix were intentionally deferred because the Mac data partition had only 2.7 GiB free. CI runs the complete generated-binding, fixture, and cross-language matrix for this head.

Start with `crates/types/src/api/tool.rs` for the canonical application contracts and `crates/worker-proto/proto/nemo/relay/worker/v1/plugin_worker.proto` for their structural worker representation. Then review `crates/core/src/plugin/dynamic/worker.rs`, `crates/worker/src/lib.rs`, and `python/plugin/src/nemo_relay_plugin/_api.py` for the local host/SDK conversions and enforcement. `crates/worker-proto/src/lib.rs` intentionally remains a generic generated-protocol and JSON-value helper crate.

For runtime semantics, review `crates/core/src/api/runtime/state.rs` and `crates/core/src/api/tool.rs`. `crates/core/tests/integration/middleware_tests.rs` captures annotation propagation, repeated/concurrent continuation behavior, pending-mark ownership, sanitization order, error behavior, and MCP-shaped result regressions.

For dynamic compatibility, review `crates/core/src/plugin/dynamic.rs`, `crates/core/src/plugin/dynamic/native.rs`, `crates/core/src/plugin/dynamic/worker.rs`, and `docs/reference/migration-guides.mdx` together. The key design choice is a Relay 0.8 cutover under the existing v1 identifiers: every plugin must rebuild and exclude pre-0.8 Relay versions. Native ABI v4 remains unchanged; workers retain the v1 identifier/package/API names but must regenerate bindings for the new structural tool-result protobuf types.

- Closes #445
- Relates to #446

## Summary by CodeRabbit

- **New Features**
  - Introduced a standardized tool execution result with a required `result` and optional opaque `annotation`.
  - Preserved annotations across callbacks, middleware, manual completion, SDKs, ATIF, and OpenTelemetry outputs.

- **Breaking Changes**
  - Tool callbacks and execution APIs now require the structured result format; legacy raw results are rejected.

- **Updates**
  - Dynamic plugins retain native API 1 and `grpc-v1`, with a Relay 0.8 rebuild requirement and updated migration guidance.


bbednarski9:
@coderabbit review

crates/types/src/api/tool.rs
```diff
@@ -9,6 +9,34 @@ use serde::{Deserialize, Serialize};
...
+}
+
+fn normalize_annotation(annotation: Option<Json>) -> Option<Json> {
+    annotation.filter(annotation_is_present)
+}
+
+fn annotation_is_absent(annotation: &Option<Json>) -> bool {
+    normalized_annotation(annotation).is_none()
+}
+
```
willkill07: This being here makes no sense

crates/worker/src/lib.rs
willkill07, line 1: structured changes need to happen in the proto definition instead of materializing a Json schema ad-hoc

crates/ffi/nemo_relay.h
willkill07, line 1: Note: no API changes in this file. Just notes on how to interpret the Json envelope

crates/worker-proto/src/lib.rs
willkill07, line 18: I think this is why I'm so confused.

You are bumping to v2 but still use the v1 worker definition. This is extremely misleading and makes conformity impossible.

docs/build-plugins/dynamic-plugins/native-dynamic/about.mdx
```diff
@@ -50,15 +50,18 @@ library = "target/release/libacme_native_policy.dylib"
...
-Typed native middleware requires `compat.relay = ">=0.8.0,<1.0"`. Relay rejects
-its registration when the manifest admits a host that predates the SDK-owned
-executor and native ABI v4. Plugins that use only synchronous registrations,
-such as this example, should retain `compat.relay = ">=0.5,<1.0"`.
+Native API 2 requires `compat.relay = ">=0.8.0,<1.0"`. It uses
+`ToolExecutionResult` for tool continuations and preserves opaque result
+annotations through synchronous and asynchronous tool intercepts. This JSON
+contract is independent of the negotiated native ABI table version. The current
+Rust SDK uses ABI v4, while custom ABI v2 and v3 plugins can also declare native
+API 2. Relay still accepts `native_api = "1"` for existing ABI v2, v3, and v4
```
willkill07: There is zero need to change `native_api` to `2` since there is no breaking change in wire format. Forcing folks to recompile and update is enough. We can include this in the migration guide.

willkill07 (changes requested):

docs/build-plugins/dynamic-plugins/native-dynamic/about.mdx
bbednarski9: You’re right that the native host-table ABI does not change: ABI v4 keeps the same layouts and callback signatures, and existing `native_api = "1"` binaries remain supported. The reason for `native_api = "2"` is the semantic contract carried through the existing JSON slots. Under API 1, tool `next` returns raw result JSON and Relay adapts legacy intercept outcomes with `annotation: None`; under API 2, `next` returns `ToolExecutionResult` and intercepts exchange the canonical annotated outcome. Changing that meaning in place for API 1 would silently break already-built plugins even though the C layout is unchanged. Separately, upgrading and recompiling against the current Rust plugin SDK is source-breaking because `ToolNext::call` changes from `Result<Json>` to `Result<ToolExecutionResult>`. I’ll clarify the PR wording to distinguish the unchanged ABI/API-1 compatibility from the API-2 semantic contract and SDK migration.

willkill07: I still think that a clear migration path should be preferred. Otherwise we will be arbitrarily increasing numbers until stabilization, compounding into user confusion. It's already bad enough with host ABI table v4, grpc-v1, and native_api 1.

I'd argue that if the FFI bindings do not have a strong versioning contract in place, then it's fine for the native_api version contract to "break" while keeping the same version.

mnajafian-nv (commented):
Great work, I don’t have anything new beyond Will’s open review comments here. I’m good to approve once those are addressed.

crates/types/src/api/tool.rs
bbednarski9: - Moved TOOL_RESULT_ANNOTATION_PROFILE_KEY beside CategoryProfile in [event.rs (line 219)](/private/tmp/nemo-relay-pr575-ci/crates/types/src/api/event.rs:219).
- Kept envelope schema constants in [tool.rs (line 12)](/private/tmp/nemo-relay-pr575-ci/crates/types/src/api/tool.rs:12).
- Simplified annotation normalization from four helpers to three.
- Updated all core, observability, PII, and test imports. No obsolete compatibility re-export remains.

willkill07 (approved):

willkill07 (approved):

mnajafian-nv (approved):

PR #578, opened 2026-07-29
fix: bound transparent gateway shutdown

bk-nvidia:
#### Overview

Prevent `nemo-relay codex` and other transparent agent runs from waiting indefinitely during gateway shutdown when an in-flight streaming request never completes.

#### Details

The transparent launcher currently sends the gateway shutdown signal and awaits the Axum server task without a deadline. Axum graceful shutdown waits for every in-flight request, so a provider stream stuck inside managed execution can keep the Relay wrapper—and therefore the terminal job—alive indefinitely after the coding agent exits.

This change gives transparent gateway shutdown a three-second grace period. If the server task remains active, Relay aborts and joins it so cleanup can finish and the wrapper can exit. Standalone and managed shared gateways are unchanged.

Validation:

- `cargo test -p nemo-relay-cli gateway_stop_aborts_a_task_that_exceeds_the_grace_period`
- `cargo test -p nemo-relay-cli --lib` (1,121 passed)

Start with `RunningGateway::stop_with_timeout` in `crates/cli/src/process/launcher.rs`. The regression test uses a permanently pending gateway task and verifies that the timeout aborts and drops it.

- Relates to #577.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transparent gateway shutdown reliability with a bounded 3-second grace period to prevent CLI hangs.
  * Ensures shutdown succeeds when the gateway is cancelled due to interrupt/timeout, while surfacing other unexpected failures consistently.
  * Logs when the gateway shutdown grace period is exceeded.
* **Tests**
  * Added coverage verifying gateway stop aborts tasks that remain unresponsive past the grace period, and updated existing shutdown tests accordingly.

### DIFF
--- a/crates/cli/src/process/launcher.rs
+++ b/crates/cli/src/process/launcher.rs
@@ -26,6 +26,8 @@ use crate::server::GatewayOverrides;
 
 use super::{PreparedAgentLaunch, RunOverrides};
 
+const TRANSPARENT_GATEWAY_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(3);
+
 /// Runs a child coding-agent command behind an ephemeral local gateway.
 ///
 /// The gateway binds to an OS-assigned loopback port, prepares agent-specific hook/gateway wiring,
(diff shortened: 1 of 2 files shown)


crates/cli/src/process/launcher.rs
```diff
@@ -378,13 +380,29 @@ impl RunningGateway {
...
-        let _ = self.shutdown_tx.send(());
-        self.task
+        self.stop_with_timeout(TRANSPARENT_GATEWAY_SHUTDOWN_TIMEOUT)
             .await
-            .map_err(|error| CliError::Launch(format!("gateway task failed: {error}")))?
+    }
+
+    async fn stop_with_timeout(mut self, timeout: Duration) -> Result<(), CliError> {
+        let _ = self.shutdown_tx.send(());
+        match tokio::time::timeout(timeout, &mut self.task).await {
```
willkill07: Do we want to add a debug or info log event here?

willkill07 (approved):
One suggestion for surfacing visibility of shutdown.

Otherwise LTGM

willkill07:
/ok to test [1908eee](https://github.com/NVIDIA/NeMo-Relay/pull/578/commits/1908eee7add5e3b9e7f46b5c98a210051d0cb1a4)

willkill07 (approved):

ericevans-nv (approved):

PR #597, opened 2026-07-31
feat(adaptive): add opt-in caching for tool results

zhongxuanwang-nv:
#### Overview

Adds opt-in tool-result caching to Adaptive response caching. A cache hit suppresses the real tool call, so caching remains disabled by default and only explicitly classified read-only, TTL-stable tools are eligible.

#### Details

- Adds default, class, and per-tool override policies with exact/wildcard resolution, TTL, bypass rate, argument skips, and tool-version keying.
- Shares the existing cache store while keeping tool and LLM keys disjoint; key and store failures fail open to a live tool call.
- Adds Rust, Python, Node.js, and Go configuration parity, validation, doctor diagnostics, saved-invocation marks, and focused integration/benchmark coverage.
- Keeps the feature diff focused: user-facing tool-cache documentation remains outside this PR, and only essential API/safety comments are retained.
- Is based directly on NVIDIA/NeMo-Relay `main` at `b467deae`; this branch reconciles with the merged canonical tool-execution-result API while preserving annotations on cache hits. Earlier logical-key and staged documentation changes are not part of this diff.
- Breaking changes: none.
- Validation: `just ci=true test-rust` (3,717 passed), response-cache unit tests (61 passed), response-cache integration tests (50 passed), Python (613 passed), Node.js (342 passed), Go, workspace clippy with warnings denied, and all repository pre-commit hooks pass. The canonical non-CI Rust run also completed all unit/integration tests but encounters the existing upstream `scope_stack.rs` doctest reference to unavailable `nemo_relay::Result`.

Start with `crates/adaptive/src/response_cache/tool.rs` for policy resolution and fail-open execution, then `crates/adaptive/tests/integration/response_cache_tests.rs` for the behavioral contract.

- Relates to NVIDIA/NeMo-Relay#404

## Summary by CodeRabbit

* **New Features**
  * Added opt-in caching for tool results alongside exact-match LLM response caching.
  * Added configurable tool policies for classes, overrides, TTLs, argument exclusions, bypass rates, priorities, and optional error caching.
  * Exposed tool-cache configuration across Python, Node.js, and Go APIs.
  * Added diagnostics for tool-cache status and configuration issues.

* **Bug Fixes**
  * Improved cache-key normalization and validation for tool arguments, headers, wildcard rules, and expired entries.
  * Improved handling of cache failures and malformed streaming responses.


zhongxuanwang-nv:
@coderabbitai review

bbednarski9:
Big PR. We are going to need to align this with https://github.com/NVIDIA/NeMo-Relay/pull/575 for the 0.8 release. Please include me in reviews before merge

zhongxuanwang-nv:
@bbednarski9 Sounds great! I think 575 can go in before this, and I would request your review after this PR is completely ready!

crates/adaptive/src/response_cache/mark.rs
```diff
@@ -103,25 +103,29 @@ fn probed_savings(entry: &CacheEntry) -> (Option<u64>, Option<f64>) {
 pub(crate) struct CacheMark<'a> {
     status: &'a str,
     reason: Option<&'a str>,
+    surface: &'a str,
```
willkill07: Surfaces should be a concrete enum rather than an arbitrary string

crates/adaptive/src/response_cache/mark.rs
```diff
@@ -130,6 +134,12 @@ impl<'a> CacheMark<'a> {
         self
     }
 
+    /// Overrides the surface label (defaults to `"llm"`; tool marks set `"tool"`).
+    pub(crate) fn surface(mut self, surface: &'a str) -> Self {
+        self.surface = surface;
+        self
+    }
```
willkill07: I don't like arbitrary strings

crates/adaptive/src/config.rs
```diff
@@ -405,6 +410,7 @@ nemo_relay::editor_config! {
             nested: BackendConfig,
             default: BackendConfig,
         },
+        tools => { label: "tools", kind: Json, optional: true },
```
willkill07: The kind should be more comprehensive than just "Json" -- it should be a real type.

willkill07 (changes requested):

crates/adaptive/src/response_cache/mark.rs
zhongxuanwang-nv: Oops, yes it should definitely be!!

crates/adaptive/src/response_cache/mark.rs
zhongxuanwang-nv: Yeah.. indeed it doesn't make much sense here

crates/adaptive/src/config.rs
zhongxuanwang-nv: Yes!! Just brought a schema in :)

willkill07:
@zhongxuanwang-nv i'm deferring review on this until the tool result PR of @bbednarski9 is merged

willkill07:
@zhongxuanwang-nv bryan's PR has been merged. I believe there are changes to be made based on updating this branch

crates/adaptive/src/response_cache/config.rs
```diff
@@ -63,3 +65,112 @@ nemo_relay::editor_config! {
...
+    pub classes: BTreeMap<String, ToolClass>,
+    /// Per-tool refinements keyed by exact name or wildcard.
+    pub overrides: BTreeMap<String, ToolOverride>,
+}
+
+impl Default for ToolCacheConfig {
+    fn default() -> Self {
+        Self {
+            enabled: false,
+            priority: 150,
```
willkill07: why is this priority selected?

crates/adaptive/src/response_cache/config.rs
```diff
@@ -63,3 +65,112 @@ nemo_relay::editor_config! {
...
+        priority => { label: "priority", kind: Integer },
+        cache_errors => { label: "cache_errors", kind: Boolean },
+        default => {
+            label: "default",
+            kind: Section,
+            nested: ToolClass,
+            default: ToolClass,
+        },
+        classes => { label: "classes", kind: Json },
+        overrides => { label: "overrides", kind: Json },
```
willkill07, line 114: Aribitrary Json is not kind for the editor. Is there a way to generalize this appropriately? Node and Python bindings have type safety for these. Ideally, the editor should, too. Though if there is a new feature required for editor, let's not worry about it.

crates/adaptive/src/response_cache/config.rs
```diff
@@ -63,3 +65,112 @@ nemo_relay::editor_config! {
...
+    pub arg_skip: Option<Vec<String>>,
+}
+
+nemo_relay::editor_config! {
+    impl ToolClass {
+        cacheable => { label: "cacheable", kind: Boolean },
+        ttl_seconds => { label: "ttl_seconds", kind: Integer, optional: true },
+        bypass_rate => { label: "bypass_rate", kind: Float, optional: true },
+        arg_skip => { label: "arg_skip", kind: Json },
+        members => { label: "members", kind: Json },
```
willkill07: Same here...

crates/adaptive/src/response_cache/config.rs
```diff
@@ -63,3 +65,112 @@ nemo_relay::editor_config! {
...
+    }
+}
+
+nemo_relay::editor_config! {
+    impl ToolOverride {
+        cacheable => { label: "cacheable", kind: Boolean, optional: true },
+        ttl_seconds => { label: "ttl_seconds", kind: Integer, optional: true },
+        bypass_rate => { label: "bypass_rate", kind: Float, optional: true },
+        tool_version => { label: "tool_version", kind: String, optional: true },
+        arg_skip => { label: "arg_skip", kind: Json, optional: true },
```
willkill07: And here

crates/adaptive/src/response_cache/tool.rs
```diff
@@ -0,0 +1,390 @@
...
+    }
+    let mut cursor = first.len();
+    let (last, middles) = rest
+        .split_last()
+        .expect("a starred pattern splits into at least two segments");
+    for segment in middles {
+        match name[cursor..].find(segment) {
+            Some(position) => cursor += position + segment.len(),
+            None => return false,
+        }
```
willkill07: This does a greedy first-match only.

Do we want to restrict patterns to `*suffix`, `prefix*`, or `*contains*` only?

That way it's:
- O(N) single-pass
- simple to reason about

Otherwise, it's faster (and likely better) to convert a wildcard pattern to a regular expression and check to see if the name matches.

crates/adaptive/tests/integration/response_cache_benchmark_tests.rs
```diff
@@ -326,7 +326,6 @@ async fn reinitialized_cache_starts_empty() {
          would let the second run hit on the first run's distinct prompts"
     );
 }
-
```
willkill07, line 329: Why was this deleted?

willkill07 (changes requested):
The logic for caching seems sound to me. I just have some concerns about:
1. the user experience in the CLI configuration editor
2. surprising default priority value
3. pattern matching for tool names

crates/adaptive/src/response_cache/config.rs
zhongxuanwang-nv: That's a good point! I think here is just wants a default value higher than the NeMo Guardrails + PII redaction's default value (which is 100), so guardrails/redaction can run before cache lookups :) Do you think we should keep this default priority?

crates/adaptive/tests/integration/response_cache_benchmark_tests.rs
zhongxuanwang-nv: Oops! I'm adding this back :)

crates/core/tests/unit/llm_api_tests.rs
```diff
@@ -245,6 +245,24 @@ fn response_sanitizer_context_preserves_all_codec_identity_states() {
     );
 }
 
+#[test]
```
bbednarski9: Is this test out of scope for this PR?

bbednarski9: maybe just missing a comment about relevance to tool caching?

crates/adaptive/src/response_cache/key.rs
```diff
@@ -85,20 +89,21 @@ pub fn build_cache_key(
         "openai_chat_token_cap": chat_token_cap_spelling,
         "body": body,
         "headers": headers,
+        "header_allowlist": header_allowlist,
```
bbednarski9: Adding header_allowlist to the key makes sense. But this would invalidate caches made with CACHE_SCHEMA_VERSION=1. We could either bump to CACHE_SCHEMA_VERSION=2 (which i dont think @willkill07 would like because we're trying to keep schema version updates contained), or we would call this a breaking change to existing caches  (like Redis) where previous key-value pairs would be orphaned. I think we should just mark this as a breaking change PR with a apt description warning

crates/adaptive/src/response_cache/config.rs
```diff
@@ -63,3 +65,112 @@ nemo_relay::editor_config! {
...
+        },
+        classes => { label: "classes", kind: Json },
+        overrides => { label: "overrides", kind: Json },
+    }
+}
+
+/// Policy shared by a class of tools.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[serde(default)]
+pub struct ToolClass {
```
bbednarski9, line 121: What do you think about adding tool_version here so that we dont break the entire cache on a tool version bump?

bbednarski9 (changes requested):

crates/adaptive/src/response_cache/config.rs
zhongxuanwang-nv: Thanks Bryan! I do agree; i just implemented this :)

crates/core/tests/unit/llm_api_tests.rs
zhongxuanwang-nv: Yeah I agree this is out-of-scope for this PR! Thanks Bryan for catching! I would just remove it

crates/adaptive/src/response_cache/config.rs
zhongxuanwang-nv: Thanks Will! I now am using the typed string-list editor controls.

crates/adaptive/src/response_cache/config.rs
zhongxuanwang-nv: Hey Will, I think these fields are currently `BTreeMap<String, ToolClass>` and `BTreeMap<String, ToolOverride>`, so I probably can't generalize these for now? But I did generalize for other types!

crates/adaptive/src/response_cache/tool.rs
zhongxuanwang-nv: Thanks Will!! This is a great catch!!

New changes have been pushed, and now it
- Supports exact names, prefix*, *suffix, *contains*, and *.
- Uses direct starts_with, ends_with, and contains checks—no greedy segmented matching or regex compilation.
- Rejects unsupported forms like foo*bar, a*b*c, **, and foo** during configuration validation.
- Simplifies overlap detection while retaining deterministic specificity ordering.
- Documents the restricted syntax across Rust, Python, Node, Go, and the user guide.

crates/adaptive/src/response_cache/key.rs
zhongxuanwang-nv: Thanks Bryan! Yes, I think this is definitely a fair question — I think for now a clean way might be to preserve the existing v1 key document when the `header_allowlist` is empty, and include the policy discriminator only when the allowlist is non-empty, so the existing default-config cache entries reachable while intentionally cold-starting caches whose header-keying policy is configured or changes :)

crates/adaptive/src/response_cache/config.rs
willkill07: They are different types of middleware, so I don't see how their relative priority matters

willkill07 (approved):

crates/adaptive/src/response_cache/config.rs
bbednarski9: @zhongxuanwang-nv did this get resolved?

zhongxuanwang-nv: @willkill07 Thanks Will!! That's true;  I should not have included PII redaction there! I think here it's mainly just NeMo Guardrails, because its `tool_input` / `tool_output` use tool execution intercepts, like how this PR is currently doing, so if I use a priority <= Guardrail's default 100, by alphabetical order the tool cache would intercept the call first, and since a cache hit would stop there, Guardrail's tool checks would be skipped. 

So I'm wondering do you think we should still keep the 150, or 101, or just 100 which is what default priority is?

bbednarski9: if i understand correctly, I like 150 here as the default as we intend for it to fire post guardrails but this gives room for custom middleware to be inserted between guardrails and tool cache hits if the dev wants that. Whereas 101 would need to be overridden. Looks good to me im approving

bbednarski9 (approved):
ty for addressing comments, excited to see this feature applied!

crates/adaptive/src/response_cache/config.rs
willkill07: You are both missing my point. Conditional execution middleware is a different type of middleware with its own priority chain execution.

I'm asking why is 150 chosen when everywhere else is 100. It's inconsistent.

zhongxuanwang-nv: @willkill07 Thanks Will again! Yes! I think now everything's cleared up!

zhongxuanwang-nv:
/ok to test [2c4d208](https://github.com/NVIDIA/NeMo-Relay/pull/597/commits/2c4d20864cc776245f986cce0197424cfc8ea8c4)

willkill07 (approved):

zhongxuanwang-nv:
/ok to test [36de420](https://github.com/NVIDIA/NeMo-Relay/pull/597/commits/36de4202aa755163f150c05ca572f0f6bc5ae890)

willkill07 (approved):

PR #605, opened 2026-07-31
fix: make Python plugin teardown asyncio-safe

willkill07:
#### Overview

Prevent `plugin.clear()` from deadlocking a running Python `asyncio` loop when queued subscriber delivery depends on an event sanitizer coroutine.

#### Details

- Add `plugin.clear_async()` and run native teardown on the Rust blocking worker pool so the Python event loop remains available to pending sanitizer callbacks.
- Make synchronous `plugin.clear()` raise an actionable `RuntimeError` when called from a running event loop.
- Route async plugin context-manager cleanup through `clear_async()` and update Python type stubs and documentation examples.
- Add timeout-isolated regression coverage for a plugin-owned mark sanitizer and subscriber with undrained delivery.

Validation:

- `uv run pre-commit run --all-files`: passed.
- `cargo clippy --workspace --all-targets -- -D warnings`: passed.
- `just docs`: passed.
- Focused sanitizer, observability, adaptive, dynamic-host, and native binding tests: passed.
- `just test-python`: 599 passed; 12 tests were blocked by an ignored local `.nemo-relay/plugins.toml` using unsupported observability config version 2.
- `just test-rust`: 996 passed; 88 tests were blocked by the same ignored local configuration.

Start with `python/nemo_relay/plugin.py` for the public lifecycle contract, then `crates/python/src/py_plugin.rs` for the non-blocking native bridge and `python/tests/test_event_sanitizers.py` for the deadlock regression.

- Resolves [RELAY-599](https://linear.app/nvidia/issue/RELAY-599/nemo-relay070-nemo-relay070-rc2-pluginclear-hangs-in-a-running-asyncio)

## Summary by CodeRabbit

- **New Features**
  - Added asynchronous plugin cleanup with `await plugin.clear_async()`, allowing teardown without blocking Python event loops.
  - Added an awaitable API for clearing global plugin configuration.
  - Async cleanup now reports teardown and cleanup errors to callers.
  - Plugin context managers perform asynchronous cleanup on exit.

- **Bug Fixes**
  - Synchronous `clear()` rejects calls from running event loops and directs callers to the async alternative.

- **Documentation**
  - Updated plugin configuration and graceful-shutdown examples to use asynchronous cleanup where appropriate.


crates/python/src/py_plugin.rs
```diff
@@ -953,6 +953,21 @@ fn clear_plugin_configuration_py(py: Python<'_>) -> PyResult<()> {
     py.detach(clear_plugin_configuration).map_err(to_py_err)
 }
 
+#[pyfunction(name = "clear_plugin_configuration_async")]
+#[pyo3(signature = () -> "object", text_signature = "() -> object")]
+fn clear_plugin_configuration_async_py<'py>(py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
+    pyo3_async_runtimes::tokio::future_into_py(py, async move {
```
ericevans-nv, line 1051: Could we make `clear_async()` cancellation-resistant? Cancelling the Python Future cancels this Rust wrapper, but an already-started `spawn_blocking` teardown continues in the background, so the caller can return before cleanup finishes and its result is lost. The dynamic-plugin close path already keeps shared completion state so another waiter can observe the same teardown. Could this path use the same pattern and add a cancellation test?

willkill07: Done in latest commit.

ericevans-nv (approved):

PR #606, opened 2026-07-31
docs(observability): clarify OTLP batch and endpoint capacity

zhongxuanwang-nv:
#### Overview

Documents OpenTelemetry batch environment settings and endpoint-capacity guidance. This is a docs-only change with no Relay runtime, configuration, or Rust source API change. The settings increase a finite queue threshold; they do not guarantee lossless telemetry.

#### Details

- Document the process-wide `OTEL_BSP_MAX_QUEUE_SIZE`, `OTEL_BSP_MAX_EXPORT_BATCH_SIZE`, and `OTEL_BSP_SCHEDULE_DELAY` settings and their SDK defaults.
- Document that values must be positive integers, malformed values fall back to defaults, and a full queue drops completed spans instead of applying backpressure.
- Document that Relay has no endpoint-count limit, but each endpoint owns its exporter, provider, queue, and runtime resources; memory and export traffic scale with endpoint count and trace payload size.
- Clarify that values must be set before plugin activation and that endpoint-specific batch sizing is not supported.
- No Relay-specific batch configuration or runtime behavior is added.

Validation:
- Manual local OTLP smoke test, not retained as a regression: `OTEL_BSP_MAX_QUEUE_SIZE=16384` delivered all 8,001 spans in the controlled burst. `OTEL_BSP_MAX_EXPORT_BATCH_SIZE=512` was the SDK default.
- Fern documentation check (only the expected unauthenticated redirects warning)

Start with `docs/configure-plugins/observability/opentelemetry.mdx`.

- Relates to [RELAY-602](https://linear.app/nvidia/issue/RELAY-602/nemo-relayv070-rc2observability-config-v3-silently-loses-spans-under)
- Relates to [RELAY-604](https://linear.app/nvidia/issue/RELAY-604/nemo-relayv070-rc2-observability-config-v3-opentelemetryendpoints)


zhongxuanwang-nv:
retargeting & fixing...

crates/core/Cargo.toml
```diff
@@ -84,6 +84,8 @@ spdlog-rs = { version = "0.5", features = ["log", "multi-thread"] }
 tokio = { version = "1", features = ["rt", "macros", "sync", "test-util", "rt-multi-thread", "time"] }
 futures = "0.3"
 opentelemetry_sdk = { workspace = true, features = ["trace", "testing"] }
+opentelemetry-proto = { version = "0.32", default-features = false, features = ["trace", "gen-tonic-messages"] }
```
afourniernv: The dev-dependencies added for the test crate seem reasonable.

crates/core/src/observability/otel.rs
willkill07, line 1: OTel has environment variables... can we just document them?

https://docs.rs/opentelemetry_sdk/latest/opentelemetry_sdk/#environment-variables

| Variable | Description | Default value |
|:---|:----|:----:|
| OTEL_BSP_SCHEDULE_DELAY | Delay interval (in milliseconds) between two consecutive exports. | 5000 |
|OTEL_BSP_MAX_QUEUE_SIZE | Maximum queue size. | 2048 |
|OTEL_BSP_MAX_EXPORT_BATCH_SIZE | Maximum batch size. Must be less than or equal to OTEL_BSP_MAX_QUEUE_SIZE. | 512 |

crates/core/src/observability/otel.rs
```diff
@@ -731,7 +751,34 @@ fn build_tracer_provider(config: &OpenTelemetryConfig) -> Result<SdkTracerProvid
...
+    match max_queue_size {
+        Some(max_queue_size) => {
+            let processor = BatchSpanProcessor::builder(exporter)
+                .with_batch_config(
+                    BatchConfigBuilder::default()
+                        .with_max_queue_size(max_queue_size)
+                        .build(),
+                )
+                .build();
+            builder.with_span_processor(processor).build()
```
willkill07: Unconditionally create the processor always.

```rust
let processor = BatchSpanProcessor::builder(exporter).build();
```

Or if we do have the config, allow control over queue size, batch size, and delay?

bbednarski9:
Is this a breaking change to the rust source API by adding max_queue_size? If so that needs to be documented so we can make a decision on whether it can be tolerated for 0.7

willkill07 (changes requested):

afourniernv (dismissed):
I guess some actionable findings from others :/ ... disregard.

bbednarski9 (changes requested):
Im a bit confused by the title/description here. We're saying that we're preventing OTLP span loss, but the queue size just increases our threshold. It doesnt guarantee telemetry loss. Can you clarify the title and description to match the scope? Additionally see my comment about about a breaking change. If its breaking we need to add the properly formatted disclosure at the top of the PR

crates/core/src/observability/otel.rs
zhongxuanwang-nv: Hmmm.. I did not know this exists!! Yes, I do agree this is much better than the new field that I introduced here. I just started implementing this.

crates/core/src/observability/otel.rs
zhongxuanwang-nv: That's true! Thanks Will! I'll make that construction explicit and keep the batch settings consistent!

zhongxuanwang-nv:
@bbednarski9 Right, I think preventing OTLP is indeed out-of-scope.. Just updated the title! Also Will mentioned about the ENV variables that we support and I ignored — I think with these I agree that we should NOT include the `max_queue_size` field anymore because this also introduces the breaking change like you suggested. Making these changes now!

bbednarski9:
Cool, that sounds good. When you update it, can you also analyze how data can be lost, whether it's easily quantifiable/reportable, and given that analysis we may need to document limitations or best effort mitigation. Lmk when ready for Round 2

bbednarski9:
@coderabbit review

bbednarski9 (dismissed):
From docs it looks like

```
builder.with_batch_exporter(exporter).build()
```

is functionally equivalent to 

```
let processor = BatchSpanProcessor::builder(exporter).build();
builder.with_span_processor(processor).build()
```

Therefore, this fix is looking like a Docs PR, not a functional change. I think we should make this update to the title/description and do some testing to confirm that environment variable changes will fix the QA issue, but also that defaults are reasonable for most deployment.

Additionally, im including a list of edge cases to include below to include in your smoke tests to see if we need to expand any regression tests.


- OTEL_BSP_MAX_QUEUE_SIZE=0 is accepted by the Rust SDK and can cause effectively every completed span to be dropped.
- OTEL_BSP_SCHEDULE_DELAY=0 can make the processor thread repeatedly poll with a zero timeout, potentially creating a CPU-intensive busy loop.
- OTEL_BSP_MAX_EXPORT_BATCH_SIZE=0 produces unusual batching behavior rather than being rejected.
- Malformed values silently fall back to defaults, while zero is successfully parsed. The docs should say these settings must be positive integers.
- Queue size counts spans, not bytes. Because Relay disables per-span attribute caps, 16384 spans can consume substantial memory.
- Capacity is per endpoint. With N endpoints, queue memory, processor threads, and export work scale roughly with N.
- A full queue drops spans; it does not apply backpressure. The 8,001-span test only proves retention when the configured queue is larger than that controlled burst.
- Settings are read when each provider is constructed. Changing the environment after plugin activation has no effect.
- These variables affect other OpenTelemetry SDK batch processors in the process, not exclusively Relay endpoints.
- Provider shutdown() and force_flush() use an SDK-level five-second timeout, while Relay permits exporter timeouts such as the test’s 60 seconds. The test releases its blocked collector before shutdown, so it does not cover a stalled exporter during teardown.

willkill07 (approved):

PR #644, opened 2026-08-03
fix: honor programmatic component enablement

bbednarski9:
#### Overview

Discovered `plugins.toml` files could silently keep a plugin component disabled even when a caller explicitly included that component in its programmatic configuration with `enabled = true`. This happened because the typed overlay removed the default-valued `enabled` field before layering, allowing a lower-precedence file's `enabled = false` to survive.

This change restores the documented precedence contract: declaring a component programmatically applies its `enabled` value over discovered file configuration. When that changes a discovered `false` to `true`, initialization also returns and stores a warning that identifies the component and contributing file.

#### Details

- Preserve `enabled` in the typed programmatic overlay for every plugin component kind, rather than special-casing observability.
- Keep layered configuration behavior for components the caller did not declare: discovered `enabled = false` values remain effective.
- Keep an explicit programmatic `enabled = false` authoritative.
- Emit `plugin.component_reenabled` when a programmatically declared component overrides a discovered `enabled = false`; the diagnostic includes the component kind, the `enabled` field, and the source path when available.
- Propagate resolution diagnostics through regular initialization, dynamic-host activation, the returned `ConfigReport`, and the active report exposed by `report()`.
- Preserve exact/no-discovery initialization semantics and retain the previous report if activation rolls back.
- Document the precedence and warning behavior in plugin configuration guidance and the 0.7 release notes.

Enablement matrix:

| Discovered result | Caller component | Effective value | Warning |
|---|---|---|---|
| `true` | `enabled = true` | `true` | None |
| `true` | `enabled = false` | `false` | None |
| `false` | `enabled = true` | `true` | `plugin.component_reenabled` |
| `false` | `enabled = false` | `false` | None |
| Either | Component omitted | Discovered value | None |

Across discovered files, the highest-precedence layer that explicitly specifies
`enabled` wins; omission inherits the lower-precedence value. Declaring a
component without specifying `enabled` uses the typed default, `true`. To inherit
file enablement, the caller omits that component.

Validation performed:

- Focused Rust unit tests for overlay precedence, diagnostic provenance, report persistence, rollback, and dynamic-host activation passed.
- `cargo fmt --all -- --check` passed.
- Python suite passed: 614 tests.
- Node.js suite passed: 344 tests.
- Go suite passed.
- Documentation build passed (with the expected Fern authentication warning).
- A real Python collector reproduction confirmed `validate()` remains discovery-free, `initialize()` reports the warning, `report()` retains it, and the caller-enabled collector receives events.
- The pre-commit suite passed except that its `python-worker-proto-check` hook could not find `just` on the hook PATH; running the equivalent `just check-python-worker-proto` command directly passed.
- `just test-rust` passed the Rust unit and integration suites, then failed in an unchanged pre-existing doctest in `crates/core/src/api/runtime/scope_stack.rs` because the example references the unavailable `nemo_relay::Result` alias.

Start with `plugin_config_overlay_value()` and `programmatic_enable_override_diagnostics()` in `crates/core/src/plugin.rs`, then review `test_plugin_config_overlay_enables_programmatically_declared_components` and `test_programmatic_enable_override_diagnostic_names_discovered_source` in `crates/core/tests/unit/plugin_tests.rs`.

- Relates to: no public issue

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved plugin configuration diagnostics during initialization and activation.
  - Programmatic component activation settings now correctly override discovered file settings, while other defaults continue to inherit.
  - Added warnings when re-enabling components disabled in configuration files, including the source file.
  - Active plugin reports now include relevant configuration warnings and diagnostic details.

- **Documentation**
  - Documented configuration override behavior and related warnings.
  - Added this behavior to the release notes and configuration guidance.

### DIFF
--- a/crates/core/src/plugin.rs
+++ b/crates/core/src/plugin.rs
@@ -1205,12 +1205,7 @@ fn merge_plugin_components(left: &mut Json, right: Json) {
         *left = right;
         return;
     };
-    let mut base_slots: HashMap<String, Vec<usize>> = HashMap::new();
-    for (index, component) in left_components.iter().enumerate() {
-        if let Some(kind) = component_kind(component) {
-            base_slots.entry(kind.to_string()).or_default().push(index);
-        }
-    }
+    let base_component_count = left_components.len();
     let mut consumed: HashMap<String, usize> = HashMap::new();
     for component in right_components {
         let Some(kind) = component_kind(&component).map(str::to_owned) else {
@@ -1218,10 +1213,7 @@ fn merge_plugin_components(left: &mut Json, right: Json) {
             continue;
         };
         let nth = consumed.entry(kind.clone()).or_insert(0);
-        let slot = base_slots
-            .get(&kind)
-            .and_then(|slots| slots.get(*nth))
-            .copied();
+        let slot = nth_component_by_kind(&left_components[..base_component_count], &kind, *nth);
         *nth += 1;
         match slot {
             Some(index) => merge_plugin_component(&mut left_components[index], component),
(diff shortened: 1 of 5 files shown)


mnajafian-nv (approved):
LGTM, thanks

mnajafian-nv (approved):
LGTM!

PR #652, opened 2026-08-03
fix: resolve bare OTLP HTTP trace endpoints

ericevans-nv:
#### Overview

Complete bare OTLP/HTTP endpoints with the standard `/v1/traces` signal path and make `doctor` report the effective trace destination consistently with the exporter. Explicit HTTP paths and all gRPC endpoints remain unchanged.

#### Details

- Resolve bare `http` and `https` OTLP endpoints to `/v1/traces` while preserving query parameters and explicit paths.
- Use the same endpoint resolution in the exporter and `doctor` diagnostics.
- Treat HTTP 405 from the diagnostic GET probe as evidence that the configured route exists, while retaining warnings for missing routes such as HTTP 404.
- Add focused exporter, endpoint-resolution, and diagnostic regression coverage.
- Document the OTLP/HTTP endpoint behavior.

Validation:

- Focused endpoint-resolution, exporter, and `doctor` tests passed.
- `just docs` passed with the existing FDR 403 redirect-check warning.
- Commit-stage pre-commit hooks passed.
- `just test-rust` reached 1,066 passing tests and 86 failures. The first failure is an untouched pricing test using unsupported observability config version 2; the remaining failures are primarily poisoned-state cascades. No focused test for this change failed.
- The remaining cross-binding matrix was not rerun; remote CI will provide the complete PR validation.

Start with `resolve_http_trace_endpoint` in `crates/core/src/observability/otel.rs`, then review the matching `doctor` behavior in `crates/cli/src/diagnostics/mod.rs` and its regression tests.

- Relates to [RELAY-631](https://linear.app/nvidia/issue/RELAY-631/nemo-relay070-an-endpoint-without-a-signal-path-silently-exports)

## Summary by CodeRabbit

- **Bug Fixes**
  - HTTP OpenTelemetry endpoints without a path now automatically use `/v1/traces`.
  - Existing endpoint paths, query parameters, and gRPC endpoints remain unchanged.
  - Diagnostics now resolve configured HTTP endpoints before testing them.
  - HTTP 405 responses are treated as successful endpoint checks, while missing routes continue to produce warnings.

- **Documentation**
  - Clarified endpoint normalization behavior for HTTP and gRPC OpenTelemetry configuration.

### DIFF
--- a/crates/core/src/observability/otel.rs
+++ b/crates/core/src/observability/otel.rs
@@ -156,7 +156,15 @@ pub fn resolve_http_trace_endpoint(endpoint: &str) -> Cow<'_, str> {
     let Ok(mut parsed) = reqwest::Url::parse(endpoint) else {
         return Cow::Borrowed(endpoint);
     };
-    if !matches!(parsed.scheme(), "http" | "https") || parsed.path() != "/" {
+
+    let has_explicit_root_path = endpoint
+        .split(['?', '#'])
+        .next()
+        .is_some_and(|url| url.ends_with('/'));
+    if !matches!(parsed.scheme(), "http" | "https")
+        || parsed.path() != "/"
+        || has_explicit_root_path
+    {
         return Cow::Borrowed(endpoint);
     }
     parsed.set_path("/v1/traces");
--- a/crates/core/src/observability/otel.rs
+++ b/crates/core/src/observability/otel.rs
@@ -14,6 +14,7 @@
 //! - [`OpenTelemetrySubscriber`] exposes a NeMo Relay [`EventSubscriberFn`] and
 //!   convenience `register` / `deregister` / `force_flush` / `shutdown` methods
 
+use std::borrow::Cow;
 use std::cell::RefCell;
 use std::collections::{HashMap, HashSet, VecDeque};
 use std::sync::mpsc;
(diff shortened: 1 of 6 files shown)


willkill07 (approved):

mnajafian-nv (approved):
LGTM plus alignment on Bryan's feedback

crates/core/src/observability/otel.rs
```diff
@@ -149,6 +150,19 @@ pub enum OtlpTransport {
     Grpc,
 }
 
+/// Completes a bare OTLP/HTTP base URL with the standard trace signal path.
+#[doc(hidden)]
+pub fn resolve_http_trace_endpoint(endpoint: &str) -> Cow<'_, str> {
+    let Ok(mut parsed) = reqwest::Url::parse(endpoint) else {
+        return Cow::Borrowed(endpoint);
+    };
+    if !matches!(parsed.scheme(), "http" | "https") || parsed.path() != "/" {
```
bbednarski9: **P2 — Preserve explicitly configured root OTLP routes.**

`Url::path()` returns `/` for both `https://collector` and `https://collector/`, so this condition rewrites the latter to `/v1/traces` even though it is an explicit path. Before this change, the exporter used it verbatim; a collector or reverse proxy accepting OTLP at `POST /` will now be probed/exported at `/v1/traces` and can fail with 404. This also conflicts with the new documentation's promise that explicit paths are preserved.

Please preserve an explicit `/` (or explicitly document this as a breaking change) and add a regression test for a root-route collector.

bbednarski9 (commented):
One compatibility concern.

crates/core/src/observability/otel.rs
ericevans-nv: A pathless endpoint such as `https://collector` now receives `/v1/traces`, while an explicit root endpoint such as `https://collector/` is preserved. I also added resolver and doctor regression coverage for the root-route case and clarified the behavior in the documentation.

bbednarski9 (approved):

PR #653, opened 2026-08-03
fix: reject colliding typed OTLP destinations

willkill07:
#### Overview

Reject unsafe typed OpenTelemetry fan-out configurations that send different projections to the same OTLP destination.

- [x] I confirm this contribution is my own work, or I have the right to submit it under this projects license.

#### Details

- Reject different endpoint types that share the same trimmed endpoint and transport during validation and activation.
- Keep repeated endpoint types valid, and prevent policy configuration from suppressing the safety error.
- Update the multi-endpoint example and guidance to require independent destinations for different projections.

Validation: focused Rust regressions, Python, Go, and Node binding suites, docs and link checks, Cargo Clippy, and pre-commit checks. The full Rust suite was stopped after an unrelated native-plugin integration test stalled.

Start with crates/core/src/observability/plugin_component.rs, which centralizes the destination-collision invariant used by validation and activation. Then review the regression coverage in crates/core/tests/unit/observability/plugin_component_tests.rs.

- Fixes RELAY-630

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented conflicting telemetry destinations when different endpoint types could produce duplicate trace or span identifiers.
  * Preserved support for repeated destinations using the same endpoint type.

* **Documentation**
  * Clarified endpoint compatibility rules and identifier behavior.
  * Updated the example GenAI endpoint to use port 4319.

### DIFF
--- a/crates/core/src/observability/plugin_component.rs
+++ b/crates/core/src/observability/plugin_component.rs
@@ -18,6 +18,7 @@
 //! metadata; their declared scope type is preserved in the exported event
 //! stream.
 
+use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::future::Future;
 use std::panic::{AssertUnwindSafe, catch_unwind};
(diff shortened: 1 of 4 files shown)


mnajafian-nv (approved):
LGTM!

mnajafian-nv (approved):

PR #665, opened 2026-08-03
fix: recover failed ATIF remote deliveries

willkill07:
#### Overview

Make ATIF metadata-routing and remote delivery failures visible and recoverable.

#### Details

- Add bounded runtime diagnostics to active plugin reports across bindings.
- Retry configured remote ATIF destinations on each trajectory and write a local recovery artifact when every remote upload fails.
- Fail teardown for recorded delivery failures without disabling future plugin configuration changes.
- Reject absolute and traversal-bearing ATIF filename templates; document the recovery and reporting behavior.

Start with `AtifDispatcher` in `crates/core/src/observability/plugin_component.rs`, then the focused ATIF dispatcher tests.

- Fixes RELAY-638

## Summary by CodeRabbit

* **New Features**
  * Added aggregated runtime diagnostics to plugin reports across supported language interfaces.
  * Runtime failures now include component, error code, message, field, session, and occurrence details.
  * Failed remote deliveries are retried and can create local recovery copies.
  * Failed-teardown reports remain available until cleanup succeeds.

* **Bug Fixes**
  * Individual output destinations remain available after another destination fails.
  * Unsafe filename templates skip only the affected trajectory.
  * Delivery and destination-rendering failures are reported during plugin teardown.

* **Documentation**
  * Clarified recovery, retry, validation, diagnostics, and teardown behavior.

### DIFF
--- a/crates/core/src/plugin.rs
+++ b/crates/core/src/plugin.rs
@@ -44,6 +44,7 @@ use crate::api::runtime::{
     ToolExecutionFn, ToolInterceptFn, ToolSanitizeFn,
 };
 use crate::api::subscriber::{deregister_subscriber, register_subscriber};
+use crate::observability::plugin_component::ATIF_RUNTIME_DELIVERY_FAILURE_MARKER;
 pub use nemo_relay_types::plugin::{ConfigDiagnostic, DiagnosticLevel};
 
 pub mod dynamic;
@@ -2056,7 +2057,7 @@ fn clear_plugin_configuration_inner() -> PluginHostClearOutcome {
     // removal itself as unsafe.
     let callbacks_cleared = deregistration_errors
         .iter()
-        .all(|error| error.contains("ATIF runtime delivery failures"));
+        .all(|error| error.contains(ATIF_RUNTIME_DELIVERY_FAILURE_MARKER));
     let deregistration_error = (!deregistration_errors.is_empty()).then(|| {
         PluginError::RegistrationFailed(format!(
             "plugin teardown failed: {}",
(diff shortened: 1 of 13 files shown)


crates/core/src/plugin.rs
```diff
@@ -2025,8 +2052,13 @@ fn clear_plugin_configuration_inner() -> PluginHostClearOutcome {
     let deregistration_errors = previous
         .map(|mut previous_state| rollback_registrations_checked(&mut previous_state.registrations))
         .unwrap_or_default();
-    let callbacks_cleared = deregistration_errors.is_empty();
-    let deregistration_error = (!callbacks_cleared).then(|| {
+    // Runtime delivery failures are reported by an otherwise successful
+    // deregistration callback. They must propagate without treating callback
```
mnajafian-nv, line 2111: Tested on head `1ac7e97`. I reproduced a case where a still-open agent hits its first ATIF HTTP delivery failure during `clear_plugin_configuration()`. Teardown fails, but that failure is not preserved in `active_plugin_report().runtime_diagnostics`. We take `ACTIVE_PLUGIN_CONFIGURATION` before deregistration callbacks run, so the teardown-time call to `record_active_plugin_runtime_diagnostic()` cannot update the active report on this path. Can we preserve teardown-time diagnostics in the active report, or narrow the PR/docs claim to match the current behavior?

mnajafian-nv (approved):
conditional approval upon review of the repro-backed inline suggestion on teardown-time ATIF diagnostics.

crates/core/src/plugin.rs
willkill07: Fixed in [e7c1d10](https://github.com/NVIDIA/NeMo-Relay/pull/665/commits/e7c1d10b13b2f1a361b2df098cdb09703211647d)

mnajafian-nv (approved):

PR #668, opened 2026-08-03
fix: skip disabled plugin configuration layers

willkill07:
#### Overview

Disabled plugin component entries no longer contribute configuration during layered resolution, while programmatic re-enablement remains visible in diagnostics.

#### Details

- Ignore component entries that explicitly set `enabled = false` when merging plugin configuration layers, including the first file-backed layer.
- Retain disabled-state metadata so a programmatic re-enable still emits `plugin.component_reenabled`.
- Add regression coverage for lower-priority configuration, all-disabled inputs, and the diagnostic fallback.
- Document that disabled component entries neither override lower layers nor contribute configuration.

`crates/core/src/plugin.rs`, specifically `merge_plugin_config_documents`, `component_enabled_sources`, and `programmatic_enable_override_diagnostics`.

- Closes: RELAY-633

## Summary by CodeRabbit

* **Bug Fixes**
  * Disabled plugin components are now excluded from configuration layering and cannot contribute settings or override enabled components.
  * Diagnostics identify the configuration source when programmatic settings re-enable a component disabled in a configuration file.
  * Components disabled across all configuration layers are omitted from the merged configuration.

* **Documentation**
  * Clarified how enabled and disabled plugin components are matched and merged across configuration files.

### DIFF
--- a/crates/core/src/plugin.rs
+++ b/crates/core/src/plugin.rs
@@ -1715,7 +1715,12 @@ fn resolve_default_file_plugin_config() -> Result<DiscoveredPluginConfig> {
 
 struct DiscoveredPluginConfig {
     value: Json,
-    enabled_sources: HashMap<String, PathBuf>,
+    enabled_sources: HashMap<String, ComponentEnabledSource>,
+}
+
+struct ComponentEnabledSource {
+    enabled: bool,
+    path: PathBuf,
 }
 
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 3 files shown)


mnajafian-nv (approved):
LGTM! the earlier blocker is addressed thanks. The open CodeRabbit point looks like optional extra coverage, not a blocker.

mnajafian-nv (approved):

crates/core/src/plugin.rs
```diff
@@ -1208,6 +1210,13 @@ fn merge_plugin_components(left: &mut Json, right: Json) {
     let base_component_count = left_components.len();
     let mut consumed: HashMap<String, usize> = HashMap::new();
     for component in right_components {
+        if component
```
bbednarski9: One concern is the case where a higher (system vs. user) config specifies the component = false, lower file = true, caller = omit. I want to make sure we are aligned on the expectation here. Originally it would use the system config and disable it. But i see the rationale for keeping it enabled.

bbednarski9 (approved):
approving as ive thought about it and I think this new precedence is better. That way even if a plugin is disabled at higher level, it wont make it impossible to configure for a project.

crates/core/src/plugin.rs
willkill07: If we need some way to blacklist plugin(s) we can always add a specific configuration option :)

bbednarski9 (approved):

PR #670, opened 2026-08-03
fix: add offline mode to doctor probes

mnajafian-nv:
#### Overview

Add an offline mode to `nemo-relay doctor` so observability diagnostics can validate configured exporters without requiring live network reachability. Live probe output now also makes it explicit when a check is only confirming HTTP reachability or gRPC TCP reachability.

- [x] I confirm this contribution is my own work, or I have the right to submit it under this repository license.

#### Details

- add `--offline` to `nemo-relay doctor` and thread an explicit probe mode through the doctor collection path
- skip live OpenTelemetry, ATOF stream, and non-in-memory response-cache probes in offline mode while preserving config and endpoint validation
- clarify live probe messaging so HTTP checks are described as reachability probes and gRPC checks explicitly say the TCP port connected but the OTLP handshake was not verified
- add regression coverage for offline flag parsing, skipped probe behavior, malformed endpoint handling, and unchanged live probe behavior

**Validation:**
- `just test-rust`
- `cargo fmt --all`
- `uv run pre-commit run --files crates/cli/src/commands/diagnostics.rs crates/cli/src/commands/mod.rs crates/cli/src/commands/root.rs crates/cli/src/diagnostics/mod.rs crates/cli/src/diagnostics/probes.rs crates/cli/tests/coverage/commands/main_tests.rs crates/cli/tests/coverage/shared/doctor_tests.rs crates/cli/tests/coverage/shared/probes_tests.rs`
- `cargo test -p nemo-relay-cli doctor_accepts_offline_flag -- --nocapture`
- `cargo test -p nemo-relay-cli diagnostics::tests -- --nocapture`
- `cargo test -p nemo-relay-cli diagnostics::probes::tcp_tests -- --nocapture`

Start in `crates/cli/src/diagnostics/mod.rs`, especially `observability_http_exporter_checks()` and `probe_atof_stream_sink()`. The most relevant regression coverage is in `crates/cli/tests/coverage/shared/doctor_tests.rs`.

- Relates to: none

## Summary by CodeRabbit

* **New Features**
  * Added an `--offline` option to the `doctor` command.
  * Offline diagnostics skip live network checks while still validating endpoint formats and configuration.
  * Improved diagnostics for reachability, connection failures, timeouts, and invalid endpoints.

* **Bug Fixes**
  * Improved health-check behavior and messaging for HTTP, gRPC, Redis, OpenTelemetry, and stream endpoints.

* **Tests**
  * Added coverage for offline mode, command parsing, endpoint validation, and diagnostic results.

### DIFF
--- a/crates/cli/src/commands/diagnostics.rs
+++ b/crates/cli/src/commands/diagnostics.rs
@@ -21,7 +21,10 @@ pub(crate) struct DoctorCommand {
     pub(crate) install_dir: Option<PathBuf>,
     #[arg(long)]
     pub(crate) json: bool,
-    #[arg(long)]
+    #[arg(
+        long,
+        help = "Validate configuration and endpoint syntax without running live network probes"
+    )]
     pub(crate) offline: bool,
 }
 
--- a/crates/cli/src/commands/diagnostics.rs
+++ b/crates/cli/src/commands/diagnostics.rs
@@ -21,6 +21,8 @@ pub(crate) struct DoctorCommand {
     pub(crate) install_dir: Option<PathBuf>,
     #[arg(long)]
     pub(crate) json: bool,
+    #[arg(long)]
+    pub(crate) offline: bool,
 }
 
 #[derive(Debug, Clone, Args)]
@@ -41,6 +43,7 @@ pub(super) async fn execute(
     crate::diagnostics::run_doctor(
         command.agent.map(Into::into),
         command.json,
+        crate::diagnostics::DoctorProbeMode::from_offline_flag(command.offline),
         &gateway_overrides,
         logging_fallback_error,
     )
--- a/crates/cli/src/diagnostics/mod.rs
+++ b/crates/cli/src/diagnostics/mod.rs
@@ -1124,6 +1124,13 @@ async fn probe_atof_stream_sink(
             };
         }
     };
+    if let Err(details) = validate_atof_stream_probe_target(index, transport, url) {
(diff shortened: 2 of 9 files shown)


crates/cli/src/diagnostics/mod.rs
willkill07, line 1: We don't validate existing timeout/header/header-env values in the `--offline` path.

Fix: only skip payload creation and the actual HTTP/WebSocket/NDJSON request.

crates/cli/src/commands/diagnostics.rs
```diff
@@ -21,6 +21,8 @@ pub(crate) struct DoctorCommand {
     pub(crate) install_dir: Option<PathBuf>,
     #[arg(long)]
     pub(crate) json: bool,
+    #[arg(long)]
+    pub(crate) offline: bool,
```
willkill07: could we add a description for `--offline` ?

crates/cli/src/commands/diagnostics.rs
willkill07, line 1: Could we also update any corresponding documentation? If we don't have anything for doctor, let's not add some now :) we can defer to a new documentation add.

willkill07 (changes requested):
We have a regression in validating other endpoint fields in `--offline` mode. We need to fix that.

Otherwise, just minor nits.

willkill07 (approved):

PR #687, opened 2026-08-04
fix: surface ATIF delivery failures consistently

willkill07:
#### Overview

Reject non-string ATIF filename metadata instead of silently routing trajectories through a fallback, and make remote-delivery teardown failures consistently observable and drainable.

#### Details

- Reject present non-string metadata values even when a filename fallback is configured, while retaining fallback behavior for missing and null values.
- Preserve remote-upload diagnostics in `plugin.report()` after subscriber delivery is flushed.
- Remove the misleading registration-removal wrapper from ATIF delivery errors.
- Reset completed Python asynchronous clear state so a failed clear is drainable and a later clear or initialization succeeds.
- Add Rust and Python regression coverage and document the resulting filename, reporting, and teardown contract.

Validation:

- Focused Rust filename-rendering, runtime-diagnostic, HTTP retry, and Python-binding teardown tests passed.
- The Python end-to-end non-string metadata and drainable-clear regression passed against `release/0.7`.
- `cargo fmt`, workspace clippy, cargo check, changed-file pre-commit checks, Go tests, and documentation checks passed.
- Full Rust, Python, and Node runs were also exercised; local failures were limited to assertions receiving the inherited-configuration warning from `/Users/wkillian/.nemo-relay/plugins.toml`, followed by global-state cascades where those tests aborted before cleanup.

Start with `render_atif_filename` in `crates/core/src/observability/plugin_component.rs`, then the end-to-end regression in `python/tests/test_observability_plugin.py`.

- Closes: RELAY-638

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved observability plugin error reporting for invalid, non-string, or nested metadata in destination filenames.
  - Preserved delivery diagnostics through teardown, recovery, and failed reconfiguration.
  - Prevented overlapping asynchronous cleanup operations from overwriting newer plugin state.
  - Distinguished delivery errors from registration leaks, incomplete removal, and fatal shutdown failures.

- **Documentation**
  - Clarified filename fallback behavior, diagnostic persistence, teardown reporting, and recovery outcomes.

- **Tests**
  - Expanded coverage for invalid metadata, failed delivery cleanup, runtime diagnostics, and asynchronous configuration clearing.

### DIFF
--- a/crates/core/src/plugin.rs
+++ b/crates/core/src/plugin.rs
@@ -337,7 +337,13 @@ pub struct PluginRegistration {
     pub kind: String,
     /// Runtime-qualified registration name.
     pub name: String,
-    deregister: Box<dyn FnMut() -> Result<()> + Send>,
+    deregister: Box<dyn FnMut() -> PluginRegistrationCleanupOutcome + Send>,
+}
+
+pub(crate) enum PluginRegistrationCleanupOutcome {
+    Removed,
+    RemovedWithError(PluginError),
+    NotRemoved(PluginError),
 }
 
 impl fmt::Debug for PluginRegistration {
@@ -354,7 +360,22 @@ impl PluginRegistration {
     pub fn new(
         kind: impl Into<String>,
         name: impl Into<String>,
-        deregister: Box<dyn FnMut() -> Result<()> + Send>,
+        mut deregister: Box<dyn FnMut() -> Result<()> + Send>,
+    ) -> Self {
+        Self {
+            kind: kind.into(),
+            name: name.into(),
+            deregister: Box::new(move || match deregister() {
+                Ok(()) => PluginRegistrationCleanupOutcome::Removed,
+                Err(error) => PluginRegistrationCleanupOutcome::NotRemoved(error),
+            }),
+        }
+    }
+
+    pub(crate) fn new_with_outcome(
+        kind: impl Into<String>,
+        name: impl Into<String>,
+        deregister: Box<dyn FnMut() -> PluginRegistrationCleanupOutcome + Send>,
     ) -> Self {
         Self {
             kind: kind.into(),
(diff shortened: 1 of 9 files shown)


crates/core/src/plugin.rs
```diff
@@ -1553,12 +1573,37 @@ async fn initialize_plugins_exact_inner(
...
+            .take()
+            .map(|state| state.report);
+        if !teardown.errors.is_empty() {
+            if let Some(report) =
+                teardown_report.filter(|report| !report.runtime_diagnostics.is_empty())
+                && let Ok(mut guard) = LAST_FAILED_RUNTIME_DIAGNOSTICS_REPORT.lock()
+            {
+                *guard = Some(report);
+            }
+            record_rollback_failures(rollback_failures.as_ref(), teardown.errors.clone());
```
ericevans-nv: Could we avoid permanently blocking plugin changes when teardown removed the callback but its final delivery failed? `RemovedWithError` means removal succeeded, but recording it in `rollback_failures` retains the mutation lease, so every later plugin operation returns `another plugin configuration mutation is in progress`. Could we use `teardown.callbacks_cleared` so only `NotRemoved` blocks future mutations, while delivery failures still return their error and preserve the diagnostic? Please also have the replacement test attempt another public mutation afterward.

willkill07: I'm confused at the scope of this review comment. 

Most importantly, it's not permanent. It's a short-lived mutex lock.

Second, this is under plugin initialization, which by design should only be called once per runtime.

I don't understand the suggestion.

mnajafian-nv (approved):
lgtm upon reviewing inline comment

crates/core/src/plugin.rs
```diff
@@ -1558,12 +1579,37 @@ async fn initialize_plugins_exact_inner(
...
+        let teardown = rollback_registrations_checked(&mut previous_state.registrations);
+        let teardown_report = ACTIVE_PLUGIN_CONFIGURATION
+            .lock()
+            .map_err(|err| {
+                PluginError::Internal(format!("active plugin configuration lock poisoned: {err}"))
+            })?
+            .take()
+            .map(|state| state.report);
+        if !teardown.errors.is_empty() {
+            if let Some(report) =
```
mnajafian-nv, line 1603: `RemovedWithError` still flows into `rollback_failures` here. In the legacy initialization path, any recorded rollback failure disables later plugin mutations, so a delivery-only teardown error can still block future initialization even when callbacks were removed. Can we carry `callbacks_cleared` through this path and disable later mutations only for`NotRemoved` / panic cases?

mnajafian-nv (approved):

crates/core/src/plugin.rs
ericevans-nv: The mutex lock is short-lived; the concern is that `std::mem::forget(lease)` prevents the owner state from returning from `Legacy` to `Idle`, so later plugin mutations continue to conflict after initialization returns.

mnajafian-nv (approved):

ericevans-nv (approved):

PR #719, opened 2026-08-05
feat: configure OpenTelemetry batching per endpoint

willkill07:
#### Overview

Add endpoint-local OpenTelemetry batch processor configuration for observability config version 3. Omitted values continue to inherit the standard `OTEL_BSP_*` environment settings and SDK defaults.

#### Details

- Add optional `max_queue_size`, `max_export_batch_size`, and `scheduled_delay_millis` fields to each configured OpenTelemetry endpoint.
- Apply endpoint overrides after the environment-aware SDK batch defaults are constructed, preserving independent endpoint behavior and existing diagnostics, flush, and shutdown semantics.
- Reject zero values and explicitly configured export batches larger than their queue with indexed configuration diagnostics.
- Add configuration parity and serialization coverage for Rust, Python, Node.js, and Go while leaving manual subscriber APIs and the C FFI unchanged.
- Preserve the existing Python `OpenTelemetryEndpointConfig` positional constructor order by appending the new fields after all pre-existing fields, with regression coverage for legacy positional calls.
- Document precedence, sizing semantics, queue overflow behavior, scheduled-delay behavior, and the endpoint-local OTLP request timeout.

Validation:

- `just test-go`
- `just docs`
- Full isolated Python suite: 640 passed, 45 skipped
- Focused Python observability suite: 16 passed
- Focused Rust and Node.js observability configuration tests

The canonical Rust, Python, and Node.js commands were also attempted. Configuration-sensitive cases discovered a local `/Users/wkillian/.nemo-relay/plugins.toml`, producing unrelated inherited-configuration warnings and conflicts. The canonical Python run completed with 629 passing tests and 11 affected tests; the full suite passes when run with isolated configuration discovery.

Start with `crates/core/src/observability/plugin_component.rs` for the public endpoint schema and validation, then `crates/core/src/observability/otel.rs` for environment-aware batch processor construction and endpoint override precedence. The Python compatibility follow-up is covered by `python/tests/test_observability_plugin.py`.

- Relates to: RELAY-626

## Summary by CodeRabbit

* **New Features**
  * Added per-endpoint OpenTelemetry settings for queue size, export batch size, and scheduled export delay.
  * Added support across Rust, Node.js, Go, and Python configuration interfaces.
  * Endpoint-specific settings take precedence over process-wide defaults.
  * Added validation for zero values and batch sizes larger than the queue.

* **Documentation**
  * Updated configuration guidance, examples, fallback behavior, and known issues.

* **Tests**
  * Added coverage for serialization, validation, defaults, and runtime propagation.


crates/core/src/observability/plugin_component.rs
```diff
@@ -2000,6 +2027,36 @@ fn build_otel_config(
...
+    index: usize,
+    section: &OpenTelemetryEndpointConfig,
+) -> PluginResult<()> {
+    for (field, value) in [
+        ("max_queue_size", section.max_queue_size),
+        ("max_export_batch_size", section.max_export_batch_size),
+    ] {
+        if value == Some(0) {
+            return Err(PluginError::InvalidConfig(format!(
+                "OpenTelemetry endpoints[{index}].{field} must be greater than 0"
```
zhongxuanwang-nv, line 2040: I guess it's fine if we don't specify a ceiling for the `max_queue_size` or `max_export_batch_size` right? Because rust would preallocate `sync_channel::<SpanData>(config.max_queue_size) (span_processor.rs:358)`, a value too big may make OOM instantly, but I guess it's not up to us to tell the users, aka they should know this? 😅

crates/node/observability.d.ts
```diff
@@ -81,6 +81,9 @@ export interface OpenTelemetryEndpointConfig {
   service_version?: string;
   instrumentation_scope?: string;
   timeout_millis?: number;
+  max_queue_size?: number;
+  max_export_batch_size?: number;
+  scheduled_delay_millis?: number;
```
zhongxuanwang-nv, line 86: Should we mark this PR as a breaking change?

docs/configure-plugins/observability/opentelemetry.mdx
```diff
@@ -90,29 +93,44 @@ work or delivery to the other exporters.
...
-OpenTelemetry's standard batch processor settings apply process-wide to every
-configured endpoint. Set them before the plugin activates; endpoint-specific
-batch sizing is not supported.
+Configure batch processing independently on each endpoint with
+`max_queue_size`, `max_export_batch_size`, and `scheduled_delay_millis`.
+When an endpoint omits a field, the corresponding standard OpenTelemetry
+environment variable applies process-wide. If neither is set, the SDK default
+applies.
+
+The precedence for each setting is endpoint value, then environment variable,
```
zhongxuanwang-nv, line 115: Should we add something like `These endpoint fields apply to plugin-configured endpoints only; subscribers created through Direct Subscriber should use the OTEL_BSP_* environment variables.`?

zhongxuanwang-nv (approved):
Thank you Will! This looks great!!

crates/node/observability.d.ts
willkill07: `?` means nullable / optional -- not breaking :)

crates/core/src/observability/plugin_component.rs
willkill07: Yeah, they should know better

docs/configure-plugins/observability/opentelemetry.mdx
willkill07: No, because the endpoint fields also work for the direct subscribers!

```python
from nemo_relay import OpenTelemetryConfig, OpenTelemetrySubscriber

config = OpenTelemetryConfig(
    "openinference",
    "http://localhost:6006/v1/traces",
)
config.service_name = "agent-service"
subscriber = OpenTelemetrySubscriber(config)
```

crates/node/observability.d.ts
zhongxuanwang-nv: Oops! haha

docs/configure-plugins/observability/opentelemetry.mdx
zhongxuanwang-nv: Oops, I should've known this!

zhongxuanwang-nv (approved):

PR #728, opened 2026-08-06
feat(examples): add repeatable Harbor Hermes Switchyard eval

bbednarski9:
## Summary

Adds a self-contained Harbor Terminal-Bench example that runs Hermes through Relay 0.7.0 and the Switchyard plugin, without a Fabric dependency.

- separates committed `plugins.toml` configuration from an untracked, machine-local Phase 2 environment file
- adds durable, resumable cohort supervision with configurable concurrency and resource admission checks
- adds all-89 dataset wiring smoke coverage, regression/offline compatibility gates, and parallel-isolation validation
- preserves per-task validation, artifact finalization, OpenInference export, and provenance evidence

## Why

Phase 2 runs need to be repeatable on other hosts before scaling evaluation throughput. The example now supports a host-specific private env file while keeping routing/plugin policy in the versioned template.

## Validation

- exercised the Phase 2 runner locally
- `git diff --check origin/main...HEAD`
- credential audit of the proposed commits: no private env, key, PEM, or credential-signature content; committed authorization is a replacement placeholder only

## Notes

This is intentionally a draft while the Phase 2 evaluation and cross-machine run are being completed.


willkill07:
Just dropping a note that I'm not sure this belongs in the repo:
1. Hermes has native Relay integration
2. Switchyard plugin doesn't live in Relay

PR #698, opened 2026-08-04
fix: queue managed sanitizer publication

willkill07:
#### Overview

Queue managed payload sanitization and event publication so observability work does not add latency to application execution.

- [x] I confirm this contribution is my own work, or I have the right to submit it under this project's license.
- [x] I searched existing issues and open pull requests, and this does not duplicate existing work.

#### Details

- Queue managed LLM, tool, and stream payload sanitization on the subscriber publication path.
- Preserve execution-time lifecycle timestamps for ATIF, OpenTelemetry, and OpenInference, including queued stream END events.
- Update Rust, Python, Node.js, and Go test expectations to use the subscriber flush barrier, and document the queued behavior.

#### Where should the reviewer start?

Start with `crates/core/src/api/llm.rs`, `crates/core/src/api/tool.rs`, and `crates/core/src/stream.rs`; the blocked-sanitizer regression coverage is in `crates/core/tests/integration/middleware_tests.rs`.

#### Related Issues: (use one of the action keywords Closes / Fixes / Resolves / Relates to)

- Relates to: none

## Summary by CodeRabbit

* **New Features**
  * Observability sanitization and event publication now run asynchronously, reducing delays to managed and streaming execution.
  * LLM, tool, and stream lifecycle events maintain ordering while being published through the queued pipeline.
  * Sanitized payloads are processed before subscriber and exporter delivery.

* **Documentation**
  * Clarified asynchronous middleware, sanitization, publication, and streaming behavior.

* **Tests**
  * Added coverage for non-blocking sanitization, event ordering, and queued subscriber processing.

PR #750, opened 2026-08-10
perf(observability): queue managed observability sanitizers

willkill07:
#### Overview

Queue managed payload and event sanitization so observability work does not delay tool, buffered LLM, or stream execution.

#### Details

- Queue managed tool and LLM payload sanitizer chains on the serial publication dispatcher before event sanitization and subscriber/exporter delivery.
- Detach streaming LLM END finalization while preserving FIFO delivery, execution-time timestamps, and the `flush_subscribers()` completion barrier.
- Add blocked-sanitizer regression coverage for tool, buffered LLM request/response, stream termination, and delivery ordering; document the queued-observability contract.
- Supersedes #698 with the same runtime goal, plus explicit queued-END construction and compatibility with the current `enable_full_payloads` behavior.

Start with `crates/core/src/api/llm.rs` and `crates/core/src/stream.rs`; `crates/core/tests/integration/middleware_tests.rs` demonstrates the non-blocking and flush-barrier contract.

- Relates to: [RELAY-706](https://linear.app/nvidia/issue/RELAY-706/enhancement-decouple-observability-sanitization-from-managed-execution)
- Supersedes: #698

## Summary by CodeRabbit

* **New Features**
  * Lifecycle and tool events now support asynchronous payload sanitization and publication.
  * Stream completion events include UTC timestamps and use a shared publication process.
* **Performance**
  * Managed calls and stream termination no longer wait for observability processing or subscriber delivery.
  * Event delivery remains ordered while sanitization runs asynchronously.
* **Bug Fixes**
  * Improved handling of sanitized payloads, null tool results, and response-codec failures during event publication.
* **Documentation**
  * Clarified middleware execution, event publication, sanitization, and subscriber flush behavior.

### DIFF
--- a/crates/core/src/api/llm.rs
+++ b/crates/core/src/api/llm.rs
@@ -1638,10 +1638,11 @@ pub async fn llm_call_execute(params: LlmCallExecuteParams) -> Result<Json> {
         let scope_guard = scope_stack.read().expect("scope stack lock poisoned");
         snapshot_event_subscribers(scope_guard.collect_scope_local_subscribers())?
     };
+    let observability_request = intercepted_request.clone();
     inject_traceparent(&mut intercepted_request, handle.uuid)?;
     queue_llm_start_with_subscribers(
         &handle,
-        &intercepted_request,
+        &observability_request,
         annotated_request.clone(),
         request_codec.clone(),
         &lifecycle_subscribers,
@@ -1848,10 +1849,11 @@ pub async fn llm_stream_call_execute(params: LlmStreamCallExecuteParams) -> Resu
         let scope_guard = scope_stack.read().expect("scope stack lock poisoned");
         snapshot_event_subscribers(scope_guard.collect_scope_local_subscribers())?
     };
+    let observability_request = intercepted_request.clone();
     inject_traceparent(&mut intercepted_request, handle.uuid)?;
     queue_llm_start_with_subscribers(
         &handle,
-        &intercepted_request,
+        &observability_request,
         annotated_request,
         request_codec.clone(),
         &lifecycle_subscribers,
--- a/crates/core/src/api/llm.rs
+++ b/crates/core/src/api/llm.rs
@@ -500,20 +500,30 @@ fn queue_llm_start_with_subscribers(
 ) -> Result<()> {
     ensure_runtime_owner()?;
     let scope_stack = handle.captured_scope_stack().clone();
-    let (entries, agent_is_fresh, full_payloads_enabled) = {
-        let mut scope_guard = scope_stack.write().expect("scope stack lock poisoned");
-        let scope_locals = scope_guard.collect_scope_local_registries(|registries| {
-            &registries.llm_sanitize_request_guardrails
-        });
+    let scope_locals = {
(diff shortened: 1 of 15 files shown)


crates/core/src/stream.rs
```diff
@@ -354,22 +365,12 @@ impl LlmStreamWrapper {
...
-            Err(_) => {
-                let _ = subscriber_dispatcher::spawn_background_publication(finalize);
-                None
-            }
-        }
+        // Stream finalization is observability-only. Queue it on the shared
+        // publication executor so stream termination does not await response
+        // or event sanitizers. The registered barrier keeps subscriber flushes
+        // ordered behind this END event.
+        let _ = subscriber_dispatcher::spawn_background_publication(finalize);
```
afourniernv, line 369: Can we close the stream’s bookkeeping before sending this work to the background?

I live-tested the repository’s existing `stream_termination_modes_close_accounting_without_losing_evidence` test against both revisions:

- Base `1e50d67a`: passes; a late contribution is rejected immediately after stream exhaustion.
- PR runtime commit `a1af7958`: fails; the same late contribution is accepted unless `flush_subscribers()` runs first.

The current head only adds tests and documentation on top of that runtime commit. `LlmOptimizationRecorder` is public and cloneable, so a streaming plugin can still hold it during this window. The sanitization and END-event publication can remain queued; only contribution acceptance needs to be closed when the stream ends.

crates/core/src/api/llm.rs
```diff
@@ -1076,41 +1168,65 @@ async fn llm_call_end_with_behavior(
...
+                    return event;
+                };
+                let end_metadata = metadata_with_otel_status(metadata, "OK", None);
+                state.build_llm_end_event(
+                    EndLlmHandleParams::builder()
+                        .handle(&queued_handle)
+                        .data_opt(payload.data)
+                        .metadata_opt(end_metadata)
+                        .annotated_response_opt(payload.annotated_response)
+                        .timestamp_opt(timestamp)
```
afourniernv: Can we choose the END timestamp once, before queueing, and reuse it here?

I live-tested the same managed LLM call with a 200 ms response sanitizer against both revisions:

- Base `1e50d67a`: the published END timestamp was set before `llm_call_execute` returned.
- PR runtime commit `a1af7958`: the published END timestamp was 201 ms after `llm_call_execute` returned.

Managed calls normally leave `timestamp` as `None`. The provisional event therefore picks “now,” and this rebuilt event picks “now” again after response sanitization. That makes the recorded LLM duration include sanitizer delay. CodeRabbit already called out the failed-call path; this is the same problem on the successful managed-call path.

afourniernv (changes requested):
The queued observability direction makes sense. I’m requesting changes for two narrower lifecycle regressions that I reproduced against the PR base and head. Both should be fixable without changing the overall design.

afourniernv (approved):
Verified the two previously blocking regressions on current head ca5acf55: stream accounting closes before queued finalization even without an explicit flush, and delayed sanitizer publication preserves the managed call completion timestamp. Both focused tests pass locally, and the full CI matrix is green.

PR #751, opened 2026-08-10
feat(plugin)!: support async native Rust middleware

willkill07:
#### Overview

Adds typed asynchronous middleware support to the native Rust plugin SDK.

> [!WARNING]
> **BREAKING CHANGE:** **[Rust source compatibility]** Typed native Rust guardrail, sanitizer, and intercept callbacks now return futures and receive owned arguments. Update every typed callback to the asynchronous contract, await continuation calls, return `LlmJsonAsyncStream` from stream intercepts, rebuild with `nemo-relay-plugin` 0.8.0, and set `compat.relay = ">=0.8.0,<1.0"`. Subscribers, validation, registration, scope and mark helpers, raw synchronous ABI callbacks, Go, and worker subscriber contracts remain synchronous.

#### Details

- Run typed native Rust guardrails, sanitizers, and intercepts on an SDK-owned Tokio executor.
- Add ABI v4 completion-scoped codec support and pull-based downstream LLM streams while retaining v2/v3 compatibility.
- Update plugin fixtures, examples, documentation, release notes, and dynamic-plugin guidance for the async API.
- Rebased onto current upstream `main` and reconciled the sanitizer codec and migration-guide changes.

Validation: `just test-rust`, `cargo clippy --workspace --all-targets -- -D warnings`, `just test-python-plugin`, `just test-python`, `just test-go`, `just test-node`, and `just docs` passed. `uv run pre-commit run --all-files` passed every hook except `cargo-deny`, which could not load the local advisory database because it contains a duplicate `RUSTSEC-2026-0244` entry.

Start with `crates/plugin/src/async_sdk.rs` and the native host integration in `crates/core/src/plugin/dynamic/native.rs`; `crates/plugin/tests/typed_callbacks.rs` covers the callback and executor contract.

- Closes RELAY-671

## Summary by CodeRabbit

- **New Features**
  - Added asynchronous Rust native plugin middleware with SDK-managed execution.
  - Added async tool and LLM interception, guardrails, continuations, pull-based streaming, cancellation, and backpressure support.
  - Added configurable execution, concurrent downstream calls, and scope propagation.
  - Introduced native ABI v4 with compatibility for v2 and v3 plugins.

- **Documentation**
  - Updated plugin guides, examples, release notes, and migration guidance.
  - Documented async middleware behavior and Relay 0.8+ compatibility requirements for typed async plugins.


crates/core/src/plugin/dynamic/native.rs
```diff
@@ -2123,33 +2519,15 @@ unsafe extern "C" fn native_async_next_invoke_result(
...
-                    }
-                }
-            }
-        }
-        drop(_library_guard);
+        remove_native_next_operation(&cleanup_owner, tokio::task::id());
+        callback_guard.deliver(result);
     });
+    let abort = task.abort_handle();
+    if !register_native_next_operation(&owner, task.id(), abort.clone()) {
```
afourniernv, line 2652: **Blocking: rejected continuation registration can still invoke the callback.**

In plain terms, Relay says it could not start the operation, so the typed SDK discards the callback. Relay can then invoke that discarded callback anyway.

I reproduced this with a focused test that expired the operation owner before registration. The host returned `InvalidArg`, but the callback was still invoked. I confirmed the same behavior for unary continuations and pull-stream opening.

**Likelihood:** Low, but realistic. It requires cancellation or shutdown during a narrow registration window.

**Impact:** Severe. The callback can access or free memory that has already been released, causing memory corruption or a process crash.

When registration returns an error, the host must not invoke the callback, consistent with the ABI contract.

crates/plugin/src/async_sdk.rs
```diff
@@ -0,0 +1,1526 @@
...
+            () = wait_for_stream_cancellation(&output) => return,
+        };
+        let mut stream = match stream {
+            Ok(stream) => stream,
+            Err(error) => {
+                output.reject(&error).await;
+                return;
+            }
+        };
+        while let Some(item) = futures::StreamExt::next(&mut stream).await {
```
afourniernv: **Blocking: consumer cancellation does not stop a returned stream that is waiting for its next item.**

In plain terms, the user cancels the stream, but Relay can remain stuck waiting for more data from it.

I reproduced this with a focused test using a stream that deliberately remained pending. Cancelling the consumer did not release the stream or output handle. Cleanup only occurred after I manually woke the source.

**Likelihood:** Moderate. Network-backed streams commonly wait for data, and cancellation is a normal operation.

**Impact:** A stalled source can retain plugin resources indefinitely and interfere with shutdown or plugin unloading.

Please check for cancellation while awaiting every `stream.next()` operation, not only while creating the stream.

crates/core/src/plugin/dynamic/native.rs
```diff
@@ -1801,6 +2123,7 @@ unsafe extern "C" fn native_async_stream_push_json(
     if stream.cancelled.load(Ordering::Acquire) {
         return NemoRelayStatus::InvalidArg;
     }
+    stream.backpressured.store(false, Ordering::Release);
```
afourniernv, line 2126: **Blocking before ABI v4 is finalized: backpressure state is shared across producers.**

In plain terms, two producers share one "queue full" light. Producer A encounters a full queue and should retry, but producer B can later succeed and turn the light off. Producer A then incorrectly believes its failed item should not be retried.

I reproduced this exact sequence with a deterministic test: producer A received `Internal` from a full queue, the consumer drained it, producer B pushed successfully, and producer A's backpressure check then returned false.

**Likelihood:** Low with the typed SDK because it serializes forwarding. It is reachable for raw ABI plugins that write concurrently to one output stream.

**Impact:** A retryable stream chunk can be silently lost.

The result should be tied to the individual operation, or the ABI must enforce serialized push-and-check behavior.

afourniernv (dismissed):
Requesting changes for three confirmed callback, cancellation, and backpressure issues. I reproduced each with a focused deterministic test against the current PR head; the existing test suite does not cover these interleavings.

crates/core/src/plugin/dynamic/native.rs
```diff
@@ -2123,33 +2519,15 @@ unsafe extern "C" fn native_async_next_invoke_result(
...
-            }
-        }
-        drop(_library_guard);
+        remove_native_next_operation(&cleanup_owner, tokio::task::id());
+        callback_guard.deliver(result);
     });
+    let abort = task.abort_handle();
+    if !register_native_next_operation(&owner, task.id(), abort.clone()) {
+        abort.abort();
+        return NemoRelayStatus::InvalidArg;
```
mnajafian-nv, line 2655: I reproduced this on the current head with a focused unit repro. If `register_native_next_operation()` rejects the task, this still delivers the result callback once after returning `InvalidArg`. The normal success path still passes, so this looks specific to rejected registration. On the SDK side, `invoke_unary_next()` frees the callback state immediately on non-`Ok`, so this can callback into reclaimed state after reporting failure. This wants the same closed-start path or inactive-guard behavior here, plus a regression for rejected result-callback registration.

crates/core/src/plugin/dynamic/native.rs
```diff
@@ -2238,6 +2616,322 @@ unsafe extern "C" fn native_async_next_invoke_stream(
...
+            Err(payload) => callback_guard.fail(&format!(
+                "native async stream continuation panicked: {}",
+                panic_payload_message(payload.as_ref())
+            )),
+        }
+    });
+    let abort = task.abort_handle();
+    if !register_native_next_operation(&owner, task.id(), abort.clone()) {
+        abort.abort();
+        return NemoRelayStatus::InvalidArg;
```
mnajafian-nv, line 2896: I reproduced the same issue on the pull-stream open path with a focused unit repro. This returns `InvalidArg`, but the open callback still fires once afterward. `LlmStreamNext::call()` frees the callback state immediately on non-`Ok`, so this has the same ownership problem across the ABI boundary. I would use the same fix shape here and add a regression for rejected stream-open registration.

crates/core/src/plugin/dynamic/native.rs
mnajafian-nv: I just saw Alex's review feel free to discard

crates/core/src/plugin/dynamic/native.rs
mnajafian-nv: I just saw Alex's review feel free to discard

crates/plugin/src/async_sdk.rs
afourniernv: **Blocking: preserve the callback scope while polling the returned stream**

I reproduced this with the current Hermes #77915 and Switchyard #270 sources using a real NVIDIA streaming request. I temporarily replaced Switchyard’s Relay 0.7 raw adapter with this typed streaming API.

A mark emitted inside the callback Future was correctly parented to `hermes.logical_llm_call`. Once the Future returned, marks emitted while this loop polled the stream had `parent_uuid` values that did not exist in the ATOF event stream. Switchyard’s real `routing.requested` and `routing.decision` marks were affected. The complete request returned 255 chunks, so this is not specific to cancellation or early stream closure.

Running the same Hermes/Switchyard workload through Relay 0.7.1’s current raw adapter correctly parented both routing marks.

`ScopedFuture` restores the captured scope while creating the stream, but the returned stream is polled outside that binding here. This affects typed streaming plugins that emit marks, open child scopes, or call other scope-dependent Relay APIs from `Stream::next()`.

Please poll the returned stream under the captured scope and add a regression test that emits one mark inside the callback Future and another during stream polling, asserting that both retain the callback’s parent scope.

mnajafian-nv (approved):

mnajafian-nv (approved):

crates/core/src/plugin/dynamic/native.rs
```diff
@@ -2111,9 +2496,21 @@ unsafe extern "C" fn native_async_next_invoke_result(
...
     let user_data = user_data as usize;
-    let _library_guard = next._callback_user_data.clone();
-    next.runtime.spawn(async move {
+    let (start_tx, start_rx) = tokio::sync::oneshot::channel();
+    let cleanup_owner = owner.clone();
+    let task = next.runtime.spawn(async move {
+        if start_rx.await.is_err() {
+            return;
+        }
+        let mut callback_guard = NativeAsyncResultCallbackGuard {
```
afourniernv: **Blocking: successful continuation registration can lose callback ownership during cancellation.**

The failure sequence is:

1. A typed plugin calls `next.call(...)`. The SDK allocates callback state and calls `async_next_invoke_result` or `async_next_open_llm_stream`.
2. The host returns `Ok`, which transfers responsibility for that state to the host callback. From this point, the SDK expects exactly one result or cancellation callback.
3. The owning request is cancelled before the spawned continuation task is first polled.
4. Owner cleanup aborts the task, but the callback guard is only created inside the task after `start_rx.await`. If the task never reaches that line, no callback runs and the SDK cannot reclaim its state.

This can happen when a caller cancels a managed tool or LLM request just after a plugin starts its downstream call, during runtime or plugin shutdown, or when a plugin starts concurrent or detached `next` calls and the owning middleware finishes. Ordinary middleware that awaits `next.call()` is less exposed unless the request is externally cancelled. Scheduler pressure makes the window easier to hit, but is not required.

I tested both unary continuations and pull-stream opening using the real `NativeAsyncWait` and `NativeAsyncStreamReceiver` drop paths:

- With both workers briefly occupied, 512/512 calls returned `Ok`, but the host delivered 0 callbacks, ran 0 callback-state destructors, and retained all 512 plugin allocations.
- On an otherwise idle two-worker runtime, three intentionally adversarial 20,000-operation cancellation runs retained about 90-91% of callback states. That is not an expected production failure rate; it demonstrates that saturation is not required when cancellation happens immediately.

The current impact is a callback-state leak and potentially stranded detached work, not the earlier use-after-free. The contract gap is specific: a successful registration transfers callback-state ownership to the host, but this cancellation path can terminate without returning that ownership through a callback.

crates/plugin/src/async_sdk.rs
```diff
@@ -0,0 +1,1572 @@
...
+        let mut stream = match stream {
+            Ok(stream) => ScopedStream::new(stream, stream_binding),
+            Err(error) => {
+                output.reject(&error).await;
+                return;
+            }
+        };
+        loop {
+            let item = tokio::select! {
+                item = futures::StreamExt::next(&mut stream) => item,
```
afourniernv: **Blocking: a panic while polling a returned typed stream leaves the managed consumer waiting indefinitely.**

The failure sequence is:

1. The typed stream callback Future completes successfully and returns a `Stream`.
2. Relay begins polling that stream and may already deliver one or more chunks.
3. A later `poll_next` panics.
4. The SDK executor task exits because this polling loop is not covered by `catch_unwind`.
5. No `Rejected` or `Finished` terminal event is sent. Releasing the plugin's output handle does not close the host receiver because the host-side stream still retains its sender.

Realistic triggers include plugin transformation code using `unwrap` on an unexpected provider chunk, a dependency panicking inside a custom stream adapter, or an internal scope-binding invariant failing while the SDK polls the stream. Normal stream errors returned as `Err` are already handled correctly; this finding is specifically about panics.

I validated this at two levels:

- In the typed-SDK harness, a stream emitted one valid chunk and panicked on its next poll. The host observed only `[Chunk]`, with no terminal event.
- Through a loaded native CDYLIB and Relay's real `llm_stream_call_execute` path, the first chunk arrived and the next consumer poll timed out. The process stayed alive and the plugin could be cleared after dropping the stream, so this is a per-call hang rather than a process crash.
- As a control, locally translating the polling panic into a rejected output made both tests terminate correctly.

This is unlikely during a healthy stream, but the impact of the failure is unbounded wait time for the affected managed request until its caller applies cancellation or a timeout.

crates/core/src/plugin/dynamic/native.rs
willkill07: Fixed in 8f5155229. Unary results and pull-stream opening now share an internal pending/accepted/task-dropped/completed handoff. A successful registration therefore retains callback ownership and delivers exactly one result or cancellation callback even when owner cleanup aborts the spawned task before its first poll; a rejected registration suppresses delivery. Callback guards retain the plugin library guard, and any callback-triggering drop happens after releasing the handoff mutex. The new `accepted_native_callbacks_settle_when_cancelled_before_first_poll` regression uses an unpolled current-thread runtime for both paths and asserts one callback, one state drop, no provider poll, and empty owner abort maps. The existing rejected-registration regression still asserts zero callbacks.

crates/plugin/src/async_sdk.rs
willkill07: Fixed in 8f5155229. Every `poll_next` of a typed returned stream is now wrapped in `catch_unwind`; a panic becomes the terminal rejection `typed native stream panicked while polling`. The surrounding `ScopedStream`, cancellation selection, backpressure-aware output handling, and output/continuation guards remain intact. The new `typed_async_stream_rejects_poll_panics_and_releases_output` regression emits one chunk, panics on the next poll, and asserts `[Chunk, Rejected]` plus exactly one output and continuation release.

bbednarski9:
Looking now thanks for patience

docs/reference/migration-guides.mdx
```diff
@@ -30,6 +30,65 @@ NeMo Relay is built into Hermes Agent. Do not install Relay separately or
...
+})?;
+```
+
+Event sanitizers now receive `Arc<Event>`. `ToolNext`, `LlmNext`, and
+`LlmStreamNext` are owned, cloneable values with async `call` methods. Stream
+interceptors return a future containing `LlmJsonAsyncStream`. LLM sanitizer
+codec facades remain available before and after an await for the lifetime of
+the middleware completion.
+
+The SDK owns one multi-thread Tokio runtime per exported plugin, starts it
```
bbednarski9: Nit, but this should say "per configured component", right?

bbednarski9:
Will refactor #575 on top of this PR after it merges as that PR also proposed ABI v4

docs/build-plugins/dynamic-plugins/native-dynamic/about.mdx
```diff
@@ -182,9 +227,9 @@ incrementally, so Relay does not buffer the provider stream into an array.
 
 The incremental output queue is bounded. `async_stream_push_json` and
 `async_stream_reject` never block a native callback thread. If either returns
-`Internal` and the host last-error contains `backpressured`, retain the logical
-chunk or rejection and retry after the consumer advances. `InvalidArg` means
-the stream is already closed or cancelled and must not be retried. Check
+`Internal`, call `async_stream_is_backpressured`: retain and retry the logical
```
bbednarski9: I feel like this documentation is inaccurate. Are we still returning 'internal'? Also `async_stream_is_backpressured` seems unreliable for retry per chunk/thread if its being shared across multiple threads. Maybe its just better as a signal for warning/notice rather than a logic flag for retries

crates/plugin/src/async_sdk.rs
```diff
@@ -0,0 +1,1580 @@
...
+    fn drop(&mut self) {
+        self.accepting.store(false, Ordering::Release);
+        self.tracker.close();
+        let runtime = self
+            .runtime
+            .get_mut()
+            .unwrap_or_else(|error| error.into_inner())
+            .take();
+        if let Some(runtime) = runtime {
+            // The host removes callbacks before dropping the plugin descriptor,
```
bbednarski9: **Blocking: executor teardown can abort when callback deregistration runs inside a Tokio runtime.**

I reproduced this on `8f515522` with the typed callback harness. The test entered a current-thread Tokio runtime, registered and invoked a typed async interceptor to start its SDK executor, then released the final executor reference through the registration's actual `extern "C"` `free_fn`. `NativeExecutor::drop` called `runtime.block_on(self.tracker.wait())`; Tokio panicked with `Cannot start a runtime from within a runtime`, and because the panic occurred through `drop_unary_callback`, the test process terminated with `SIGABRT` after reporting `panic in a function that cannot unwind`.

The statement that teardown is not called by an SDK worker is insufficient: host deregistration may release the final callback state from a Relay Tokio worker. Executor draining and destruction should be transferred to a non-runtime OS thread while retaining the native-library guard. Simply using `shutdown_background()` may violate the accepted-task draining and unload-safety contract.

bbednarski9 (changes requested):
One blocking comment reproduced a failure case to check out. Otherwise nits

bbednarski9 (approved):

PR #755, opened 2026-08-11
fix(langchain): close finished scopes from the top of the stack down

SandyChapman:
#### Overview

> Targets `release/0.7` so this can go into a 0.7.3 cut for Fabric 0.2

**What.** When a LangChain chain run ends, the handler records it as completed and closes finished scopes from the top of the stack down, instead of attempting a close that the stack may reject and then abandoning the scope.

**Why.** Relay closes scopes strictly LIFO, but LangGraph schedules sibling chain runs as concurrent asyncio tasks that share one scope stack, so two siblings can finish in an order the stack rejects: A starts, B starts, A ends, B ends. `_pop_scope` removed the handle from `_scope_handles` *before* attempting the pop and then swallowed the rejection, so the scope stayed live on the stack with nothing left able to close it. It remained current for everything that followed, and the caller's own enclosing scope raised on exit:

```
RuntimeError: invalid argument: scope handle is not at the top of the stack
```

Downstream this reported fully successful agent work as failed (NVBug 6562846). The enclosing scope belonged to a NeMo Fabric Deep Agents invocation, whose adapter turned a teardown error into an invocation failure (since contained separately in NVIDIA/NeMo-Fabric#191) — 31 otherwise-successful agent-eval trials were scored as adapter failures in one regression run, including at parallelism 1.

#### Details

Each run's handle is tracked from `on_chain_start`. When a run ends, its prepared output, metadata and end timestamp are recorded as completed state, and finished scopes are closed from the top of the active stack down: read `scope.get_handle()`, close it when its uuid belongs to a completed run of ours, stop as soon as the top is a run still going or a scope this handler does not own.

Only the current top is ever closed, so the stack is never asked to accept a close it would reject. There is no speculative pop, hence no failure to classify and no dependency on the runtime's error text.

**On "drain only the owning stack".** Implemented, then removed, with evidence. A stack-identity check *cannot* be written against the public API: `set_thread_scope_stack` genuinely shares the stack (a push in the thread is visible outside it) but returns a **different Python wrapper**, and `ScopeStack` exposes no id and only identity equality. A handler completing a run on a worker thread would therefore park every completion and strand every scope — the original bug, reintroduced by the guard. Dropping it also removes a strong reference to a `ScopeStack` per retained completion. Covered by `test_a_run_completed_on_a_worker_thread_still_closes`.

Cross-stack safety is therefore structural, not an ownership guard: a completion is matched on the top scope's uuid **and** name. uuid alone is not sufficient — a stack rebuilt from a `PropagationContext` re-creates a scope carrying an existing uuid, so a uuid-only match pops the stand-in, discards the completion and strands the real scope on its own stack (`test_a_propagated_stack_does_not_consume_another_stacks_completion`). The two are distinguishable because the runtime names rebuilt scopes `propagated-parent` and `propagated-root` (`scope_stack.rs`), which `test_propagated_stand_ins_keep_their_reserved_names` pins so a rename fails loudly rather than silently closing the wrong scope. Residual: a chain run named exactly one of those two reserved names would still collide.

**@AjayThorve's findings.** Output is prepared when the callback fires, so later mutation cannot change emitted telemetry. Cross-stack loss is structural rather than guarded, per above.

**@willkill07's second review.** Output is prepared *before* the run is dropped from the handler's map and degrades to `None` on failure — serializing walks caller data and a cyclic output raises `RecursionError`; a run removed with no completion recorded is a scope nothing can close. A completion is deleted only once the stack accepts the close, since the runtime can refuse before mutating the stack. Handler state transitions are serialized under a lock.

Reading the current handle is guarded by `scope_stack_active()`, which does not create a stack, so completing a run in a context without one no longer materialises an empty stack there.

`_close_completed_scopes_locked` in `python/nemo_relay/integrations/langchain/callbacks.py` — the loop that reads the top and stops as soon as it is not a completed run of ours. Note the locked/unlocked split: `_pop_scope` holds the lock and calls the locked form, so the lock is never re-acquired on the normal path. That was a real fragility — an earlier revision nested the acquisition and only worked because the lock was reentrant.

Then the tests, which use a real scope stack. `test_callbacks.py` mocks `nemo_relay`, and its mock accepted any pop in any order, so it could not observe the LIFO rule this fix turns on; it now models the stack and rejects closing anything but the top.

The LangGraph tests drive real compiled graphs with no stubbing: a two-branch fan-out, a nested three-branch fan-out with a subgraph, and a failing node completing through `on_chain_error`.

#### Testing

- `python/tests/integrations` — 64 passed, 2 skipped.
- `python/tests` (excluding `plugin/` and `test_dynamic_plugin_host.py`, which need `just build-test-plugin-fixtures`) — 545 passed, 2 skipped.
- Pre-commit on the changed files: copyright, ruff, ruff-format, ty all pass.
- Mutation checks, each caught by a named test: closing the top regardless of ownership (13 tests fail); a non-reentrant lock; serializing output at close time; letting output serialization escape; deleting a completion before the pop succeeds; dropping the `scope_stack_active` guard; and a handler that pushes no scopes at all.
- Every new test was run against the unpatched upstream handler: 19 of 21 fail, and the 2 that pass are the ones that should — the harness control and the topology test, neither of which depends on this change. Of the 19, nine fail behaviourally (the scope-stack error, or its absence); the rest reference internals that do not exist upstream, so they guard against regressions in this implementation rather than detecting the original defect.
- End to end against the original symptom: built this branch as a wheel and installed it into a NeMo Fabric checkout. Fabric's own tests that assert the leak *exists* flip to failing, and its Deep Agents mitigation for this bug never engages because there is nothing left to mitigate. A wheel built from unpatched `release/0.7` was used as the control, and confirms the flip is caused by this change rather than by version skew.

#### Remaining limitations

- **A scope closed out of band** can never return to the top, so its completion is retained for the life of the handler. That is the trade for never guessing at error text; the entry is inert and everything around it still closes. Documented on the type and in a test.
- **A sibling that never ends at all** — a cancelled task with no `on_chain_error` — still leaves its parent waiting. Unchanged by this PR, and better than an abandoned scope since the state is visible.
- Untested: runs that never end retaining their `_scope_handles` entry, sync (non-async) LangGraph execution, and very wide fan-out as a performance question.

- Relates to internal NVBug

## Bug Fixes
- Improved reliability when concurrent workflow steps complete out of order.
- Preserved outputs, metadata, and completion timestamps during deferred processing.
- Prevented lingering workflow state after concurrent or nested executions.
- Improved handling of failed workflow steps and branch cleanup.
- Increased reliability for concurrent LangGraph branch execution.

## Tests
- Added coverage for nested, overlapping, and out-of-order workflow lifecycles.
- Added regression coverage for concurrent branches, shared handlers, worker-thread completion, and cleanup behavior.

### DIFF
--- a/python/nemo_relay/integrations/langchain/callbacks.py
+++ b/python/nemo_relay/integrations/langchain/callbacks.py
@@ -204,6 +204,13 @@ class NemoRelayCallbackHandler(BaseCallbackHandler):
             completed = self._completed.get(top.uuid)
             if completed is None:
                 return
+            if top.name != completed.handle.name:
+                # Same uuid, different scope. ``create_scope_stack_from_propagation``
+                # rebuilds a parent from a captured context, so our uuid can appear on
+                # another stack as a stand-in. Closing it would discard the completion
+                # and strand the real scope. Handles cannot be told apart by identity --
+                # reading the current one returns a fresh wrapper even on its own stack.
+                return
             try:
                 nemo_relay.scope.pop(
                     completed.handle,
--- a/python/nemo_relay/integrations/langchain/callbacks.py
+++ b/python/nemo_relay/integrations/langchain/callbacks.py
@@ -5,7 +5,9 @@
 
 from __future__ import annotations
 
+import datetime
 import logging
+import threading
 import typing
 
 from langchain_core.callbacks.base import BaseCallbackHandler
@@ -19,6 +21,54 @@ if typing.TYPE_CHECKING:
 _logger = logging.getLogger(__name__)
 
 
+class _CompletedScope(typing.NamedTuple):
+    """A run that has ended, waiting for its scope to reach the top of the stack.
+
+    A scope stack closes strictly LIFO, but concurrent sibling runs finish in the order
+    the graph chooses, so a run can end while scopes opened after it are still open.
+    Recording the completion and closing it once it reaches the top preserves the stack's
+    ordering without ever attempting a close the stack would reject.
+
+    A completion is only ever closed when its scope is the top of the active stack, so
+    ownership needs no separate check: a scope sits on exactly one stack, and its handle
+    uuid can only be the current top on that stack. Comparing ``ScopeStack`` objects
+    would not work anyway -- propagating a stack to a worker thread yields a different
+    Python wrapper for the same stack, with no way to correlate the two.
+
+    ``output`` is already serialized: callers own the mapping they hand to the callback
+    and may mutate it afterwards, so it is snapshotted when the callback fires, for the
+    same reason ``ended_at`` is.
+
+    Entries are removed as their scopes reach the top, so the set is bounded by the runs
+    a handler has open at once. The exception is a scope closed by something other than
+    this handler: it can never return to the top, so its entry is retained for the life
+    of the handler, holding the serialized output.
+    """
+
+    handle: nemo_relay.ScopeHandle
+    output: nemo_relay.Json | None
+    metadata: nemo_relay.Json | None
+    ended_at: datetime.datetime
+
+
+def _current_scope_handle() -> nemo_relay.ScopeHandle | None:
+    """Return the scope on top of the active stack, or ``None`` if there is not one.
+
+    ``get_handle`` creates a stack for the current context when it has none, so it is
+    guarded by a status check: a run completing somewhere without a stack has nothing
+    this handler could close, and should not leave one behind for having asked. The
+    check consults the thread binding as well as the context, so a stack propagated to
+    a worker thread still counts as active.
+    """
+    try:
+        if not nemo_relay.scope_stack_active():
+            return None
+        return nemo_relay.scope.get_handle()
+    except Exception:
+        _logger.debug("NeMo Relay: reading the current scope failed", exc_info=True)
+        return None
+
+
 class NemoRelayCallbackHandler(BaseCallbackHandler):
     """Bridge LangChain chain run IDs to NeMo Relay Agent scopes."""
 
@@ -27,7 +77,14 @@ class NemoRelayCallbackHandler(BaseCallbackHandler):
 
     def __init__(self) -> None:
         super().__init__()
-        self._scope_handles: dict[UUID, typing.Any] = {}
+        # Start, completion, top-check and pop are multi-step transitions over the maps
+        # below. A handler can be reused by callbacks delivered on worker threads, so the
+        # transitions are serialized: a top-check is only meaningful if the pop it
+        # authorizes cannot be overtaken by another thread pushing onto the same stack.
+        # Reentrant because completing a run drains, and both take the lock.
+        self._lock = threading.RLock()
+        self._scope_handles: dict[UUID, nemo_relay.ScopeHandle] = {}
+        self._completed: dict[str, _CompletedScope] = {}
 
     def on_chain_start(
         self,
@@ -68,7 +125,8 @@ class NemoRelayCallbackHandler(BaseCallbackHandler):
                 input=prepared_inputs,
                 metadata=scope_metadata,
             )
-            self._scope_handles[run_id] = handle
+            with self._lock:
+                self._scope_handles[run_id] = handle
         except Exception:
             _logger.error("NeMo Relay: on_chain_start failed", exc_info=True)
 
@@ -101,12 +159,79 @@ class NemoRelayCallbackHandler(BaseCallbackHandler):
     def _pop_scope(
         self, run_id: UUID, *, output: dict[str, typing.Any] | None = None, metadata: nemo_relay.Json | None = None
     ) -> None:
-        handle = self._scope_handles.pop(run_id, None)
-        if handle is None:
-            return
+        # Serialized before the run mappings are touched. Serialization walks
+        # caller-supplied data and can fail on its own -- a cyclic output raises
+        # ``RecursionError`` -- and a run dropped from the maps with no completion
+        # recorded is a scope nothing can ever close.
+        prepared_output = _prepare_output(output)
 
-        try:
-            prepared_outputs = _prepare_lc_payloads(output) if output is not None else None
-            nemo_relay.scope.pop(handle, output=prepared_outputs, metadata=metadata)
-        except Exception:
-            _logger.error("NeMo Relay: scope.pop failed", exc_info=True)
+        with self._lock:
+            handle = self._scope_handles.pop(run_id, None)
+            if handle is None:
+                return
+
+            self._completed[handle.uuid] = _CompletedScope(
+                handle=handle,
+                output=prepared_output,
+                metadata=metadata,
+                ended_at=datetime.datetime.now(datetime.timezone.utc),
+            )
+            self._close_completed_scopes_locked()
+
+    def _close_completed_scopes(self) -> None:
+        """Close finished scopes from the top of the active stack down."""
+        with self._lock:
+            self._close_completed_scopes_locked()
+
+    def _close_completed_scopes_locked(self) -> None:
+        """Close finished scopes from the top of ``stack`` down.
+
+        Only the scope currently on top is ever closed, so the stack is never asked to
+        accept a close it would reject. Draining stops as soon as the top is a run that
+        is still going, or a scope this handler does not own; anything completed
+        underneath it waits, which is what keeps the ordering intact.
+
+        This is also what confines a shared handler to one stack at a time: a completion
+        belonging to another stack cannot be the top here, so it is left alone.
+
+        The caller must hold ``self._lock``: a top-check is only meaningful if the pop it
+        authorizes cannot be overtaken by another thread pushing onto the same stack.
+        """
+        while self._completed:
+            top = _current_scope_handle()
+            if top is None:
+                return
+            completed = self._completed.get(top.uuid)
+            if completed is None:
+                return
+            try:
+                nemo_relay.scope.pop(
+                    completed.handle,
+                    output=completed.output,
+                    metadata=completed.metadata,
+                    timestamp=completed.ended_at,
+                )
+            except Exception:
+                # The runtime can refuse before it mutates the stack, so the scope
+                # may still be live and closable later. Keep the completion -- it is
+                # the only record able to close it -- and stop rather than spin.
+                _logger.error("NeMo Relay: scope.pop failed", exc_info=True)
+                return
+            # Only once the stack has actually accepted the close.
+            del self._completed[top.uuid]
+
+
+def _prepare_output(output: dict[str, typing.Any] | None) -> nemo_relay.Json | None:
+    """Serialize a callback output, degrading to ``None`` if it cannot be serialized.
+
+    Losing an output payload costs one scope's telemetry detail. Letting the failure
+    escape would cost the scope itself, since the run has ended and nothing else will
+    close it.
+    """
+    if output is None:
+        return None
+    try:
+        return _prepare_lc_payloads(output)
+    except Exception:
+        _logger.error("NeMo Relay: preparing scope output failed", exc_info=True)
+        return None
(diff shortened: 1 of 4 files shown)


python/nemo_relay/integrations/langchain/callbacks.py
```diff
@@ -105,8 +134,56 @@ def _pop_scope(
...
+            output=output,
+            metadata=metadata,
+            ended_at=datetime.datetime.now(datetime.timezone.utc),
+        )
+        outcome = self._close_scope(pending)
+        if outcome is _CloseOutcome.DEFERRED:
+            self._pending_pops.append(pending)
+        elif outcome is _CloseOutcome.CLOSED:
+            # Closing this scope may have exposed one that was queued behind it.
+            self._drain_pending_pops()
```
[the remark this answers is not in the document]
AjayThorve: [P1] Confirmed. This fails without worker threads, so a lock alone is not sufficient.

I reproduced it with two real async LangGraph invocations, each bound to a different `ScopeStack`, sharing one handler. Invocation 1 queues A while B is open. Invocation 2 closes X; its drain retries A against invocation 2's stack, gets `NotFound`, and deletes A as terminal. Both graphs return their expected values, but invocation 1 leaves A current and its request close fails with `scope handle is not at the top of the stack`.

The invariant needs to be: a pending pop can only be retried against the stack that owns it. Carry the originating stack on `_PendingPop` (or partition the queue by stack), drain only the active stack, and add this two-context async regression. Synchronization is still needed for threaded reuse, but it does not fix the ownership bug by itself.

AjayThorve (commented):
The reported single-stack failure is fixed. One additional telemetry-fidelity issue remains inline.

python/nemo_relay/integrations/langchain/callbacks.py
```diff
@@ -105,8 +134,56 @@ def _pop_scope(
...
 
+        # A scope stack closes strictly LIFO, but concurrent sibling runs finish in the
+        # order the graph chooses, so this run's scope may not be on top yet. Queue a
+        # rejected close and replay it once the scopes above it have gone, rather than
+        # abandoning a scope that is still live: an abandoned scope stays current for
+        # everything that follows and makes the enclosing scope fail to close. The end
+        # time is captured now so a deferred close still records when the run ended.
+        pending = _PendingPop(
+            handle=handle,
+            output=output,
```
AjayThorve: [P2] Snapshot the output when the callback fires

`_PendingPop` retains the caller-owned `outputs` dict, and `_prepare_lc_payloads` does not run until each close attempt. LangChain passes the same callback arguments through its handler list, so later code can mutate this object while A is deferred. I reproduced `outputs={"done": "A-at-callback"}`, deferred A, mutated it to `"A-mutated-after-callback"`, then closed B; A's emitted end event contained the mutated value.

That makes telemetry depend on replay timing and contradicts the callback-time semantics already preserved for `ended_at`. Prepare or copy the output in `_pop_scope` and queue that snapshot, then assert that the emitted payload remains the callback-time value.

willkill07 (changes requested):
As an alternative to retrying queued handles and classifying exceptions, could we consider an owner-stack-aware, top-driven state machine?

Store:
- each run’s `handle`
- owning `ScopeStack`
- completion payload
- metadata
- timestamp

When a run ends, mark it complete and drain only its owning stack:
- inspect `scope.get_handle()`
- close it when the top UUID belongs to a completed run
- repeat until the top is still running or is not owned by the handler

This would preserve LIFO ordering, prevent cross-stack queue loss, and avoid depending on exception-message matching.

I would also add a regression test using one shared handler across two isolated stacks, plus Ajay’s callback-time output snapshot test.

python/tests/integrations/langchain_tests/test_callbacks_scope_stack.py
```diff
@@ -0,0 +1,564 @@
...
+    _end(handler, run_a, "A")
+    assert len(handler._completed) == 1, "A's completion should be retained"
+
+    # A later run on the same stack is unaffected.
+    _start(handler, run_b, "B")
+    _end(handler, run_b, "B")
+
+    # And the enclosing scope still closes, which is what actually matters.
+    nemo_relay.scope.pop(request)
+    assert handler._scope_handles == {}
```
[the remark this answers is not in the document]
SandyChapman: Documented the bound on `_CompletedScope` and in the test: entries clear as scopes reach the top, except one closed out of band, which is retained with its output and stack reference.

python/tests/integrations/langgraph_tests/test_langgraph_integration.py
```diff
@@ -6,7 +6,8 @@
 from __future__ import annotations
 
 import asyncio
-from typing import TYPE_CHECKING, Any, cast
+import operator
+from typing import TYPE_CHECKING, Annotated, Any, cast
```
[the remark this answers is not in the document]
SandyChapman: Tried it — renaming to `_state` breaks `ty`, since LangGraph's `add_node` stops accepting the node signature. ARG001 is not in this repo's enabled ruff rules (`select = ["E","F","W","I"]`), so I left the parameter named.

python/tests/integrations/langchain_tests/test_callbacks_scope_stack.py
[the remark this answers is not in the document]
SandyChapman: Eviction isn't implementable: neither ScopeStack nor ScopeHandle supports weak references, so the handler can't observe an owning stack becoming unreachable. A cap would risk evicting a completion that is still legitimately waiting — indistinguishable from an orphan without a "is this handle still on the stack" API — which would reabandon the scope this PR exists to fix. Bounding this properly needs a runtime-side accessor; happy to file that separately if maintainers want it.

python/nemo_relay/integrations/langchain/callbacks.py
```diff
@@ -102,11 +154,47 @@ def _pop_scope(
...
 
-        try:
-            prepared_outputs = _prepare_lc_payloads(output) if output is not None else None
-            nemo_relay.scope.pop(handle, output=prepared_outputs, metadata=metadata)
-        except Exception:
-            _logger.error("NeMo Relay: scope.pop failed", exc_info=True)
+        self._completed[handle.uuid] = _CompletedScope(
+            handle=handle,
+            stack=stack,
+            output=_prepare_lc_payloads(output) if output is not None else None,
```
willkill07: The run has already been removed from `_scope_handles` and `_scope_stacks` when `_prepare_lc_payloads` executes here, and this call is outside an exception boundary.

A cyclic output reproduces the failure: `on_chain_end` raises `RecursionError`, `_completed` remains empty, and the enclosing request cannot close because the child scope is still live.

**AR:** Please prepare the output before discarding the run mappings, use a safe fallback when serialization fails, and add a regression proving instrumentation cannot strand the scope.

python/nemo_relay/integrations/langchain/callbacks.py
```diff
@@ -102,11 +154,47 @@ def _pop_scope(
...
+        if stack is None or stack is not _current_scope_stack():
+            # The owning context is not the active one, so its top is not ours to read.
+            # The completion waits for a callback that runs on its own stack.
+            return
+
+        while self._completed:
+            top = _current_scope_handle()
+            if top is None:
+                return
+            completed = self._completed.pop(top.uuid, None)
```
willkill07: Retain the completion until `scope.pop` succeeds

This removes the only completion record before the native pop.

The core operation can fail before mutating the stack. For example: during owner checks, lock acquisition, or subscriber/stack snapshotting. So the comment below that retrying cannot help is not generally true. A failure here again leaves a live scope with no state able to close it.

**AR:** Please delete the completion only after a successful pop (or restore/mark it failed), and change the current drop-on-failure test to cover a transient failure followed by successful cleanup.

python/nemo_relay/integrations/langchain/callbacks.py
```diff
@@ -27,7 +76,9 @@ class NemoRelayCallbackHandler(BaseCallbackHandler):
 
     def __init__(self) -> None:
         super().__init__()
-        self._scope_handles: dict[UUID, typing.Any] = {}
+        self._scope_handles: dict[UUID, nemo_relay.ScopeHandle] = {}
+        self._scope_stacks: dict[UUID, nemo_relay.ScopeStack | None] = {}
+        self._completed: dict[str, _CompletedScope] = {}
```
willkill07, line 87: Synchronize shared-handler state transitions

These maps participate in multi-step start, completion, top-check, and pop transitions but have no synchronization. With one handler reused by worker-thread callbacks, one thread can remove the completion for top `B` while another pushes `C`; the subsequent pop of `B` fails and its completion is lost.

**AR:** Please serialize handler-owned transitions per owning stack (or otherwise make them atomic) and add a deterministic worker-thread regression. The existing asyncio-task tests do not exercise this interleaving.

willkill07 (changes requested):

python/nemo_relay/integrations/langchain/callbacks.py
SandyChapman: Serialized under one `threading.RLock` rather than per-stack: stack identity isn't comparable from Python (`set_thread_scope_stack` hands back a different wrapper, `ScopeStack` has identity-only `__eq__` and isn't weakref-able), so keying by owner stranded every scope completed on a worker thread. Cross-stack safety instead comes from matching uuid **and** name — a stack rebuilt from a `PropagationContext` re-creates a scope under an existing uuid, so uuid alone can match a stand-in on another stack. Regressions: `test_a_run_completed_on_a_worker_thread_still_closes`, `test_a_queued_close_is_not_discarded_by_another_stacks_drain`, `test_a_propagated_stack_does_not_consume_another_stacks_completion`.

python/nemo_relay/integrations/langchain/callbacks.py
SandyChapman: Done — `_pop_scope` now calls `_prepare_output` before it touches either map, and that helper falls back to `None` with an error log if serialization raises. `test_an_unserializable_output_does_not_strand_the_scope` uses your cyclic-dict repro and asserts the enclosing scope still closes.

python/nemo_relay/integrations/langchain/callbacks.py
SandyChapman: Done — the `del self._completed[top.uuid]` now runs only after `scope.pop` returns; a raise leaves the completion in place for the next drain. The drop-on-failure test is replaced by `test_a_transient_pop_failure_keeps_the_completion_for_a_later_close`, which fails one pop then lets the retry succeed.

python/nemo_relay/integrations/langchain/callbacks.py
```diff
@@ -101,12 +159,79 @@ def on_chain_error(
...
+        with self._lock:
+            handle = self._scope_handles.pop(run_id, None)
+            if handle is None:
+                return
+
+            self._completed[handle.uuid] = _CompletedScope(
+                handle=handle,
+                output=prepared_output,
+                metadata=metadata,
+                ended_at=datetime.datetime.now(datetime.timezone.utc),
```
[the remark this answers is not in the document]
SandyChapman: Not applicable to this repo: `UP017` and `G201` aren't in the enabled rule set (`[tool.ruff.lint] select = ["E", "F", "W", "I"]`, pyproject.toml:148), so the lint pass doesn't flag these. Ruff 0.16.1 — the version cited here — reports `All checks passed!` on both files under the repo config; the findings only appear with `--select UP017,G201` forced on. Adopting either is a repo-wide config decision rather than something this PR should change in two files.

python/tests/integrations/langchain_tests/test_callbacks_scope_stack.py
```diff
@@ -0,0 +1,539 @@
...
+    parent_run, run_a, run_b = uuid4(), uuid4(), uuid4()
+    request = nemo_relay.scope.push("request", nemo_relay.ScopeType.Agent)
+
+    _start(handler, parent_run, "parent")
+    _start(handler, run_a, "A", parent_run_id=parent_run)
+    _start(handler, run_b, "B", parent_run_id=parent_run)
+
+    _end(handler, run_a, "A")
+    assert len(handler._completed) == 1, "A should be waiting on B"
+    a_ended_by = datetime.datetime.now(datetime.timezone.utc)
```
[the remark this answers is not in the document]
SandyChapman: Same as the `callbacks.py` thread: `UP017` isn't in this repo's enabled rules (`select = ["E", "F", "W", "I"]`, pyproject.toml:148), and Ruff 0.16.1 passes clean on this file under the repo config.

willkill07 (approved):

PR #767, opened 2026-08-12
fix(cli): avoid bootstrap permission writes during runtime startup

ericevans-nv:
#### Overview

Relay uses a private bootstrap key during local startup. Creating the key and securing its directory and file permissions requires write access to Relay’s persistent bootstrap state.

Before this change, Relay performed that setup during runtime startup. In a restricted environment, the process may be allowed to read existing Relay state but not create files or change permissions, causing startup to fail.

For persistent integrations, this change prepares the bootstrap key during installation. Installation must have access to Relay’s persistent state; if it runs in a fully read-only sandbox before the key exists, it fails early rather than leaving startup to fail later.

Later runtime starts reuse an existing private key without modifying the filesystem.

#### Details

- Prepare the bootstrap key during persistent Relay installation.
- Reuse an existing private key during runtime startup instead of resetting its permissions.
- Retain the existing repair path for missing or unsafe key state.
- Add focused Unix tests for readonly-key reuse and permission repair.

Claude Code and Codex were the environments where this issue was observed, but the fix applies to any restricted runtime that can read an already-provisioned private key but cannot modify its directory or permissions.

Validation:

- Passed: focused bootstrap-key tests (6 passed).
- Passed: `cargo fmt --all`, workspace Clippy, and staged-file pre-commit checks.
- Limitation: the full Rust suite has one failure in an unchanged Python NeMo Guardrails test involving `./rails`; it does not exercise this CLI bootstrap-key behavior.
- Limitation: all-files pre-commit reported external GitHub documentation-link HTTP/2 errors.

Start with the private-key reuse decision in `crates/cli/src/configuration/mod.rs`, then review installation-time preparation in `crates/cli/src/commands/install.rs` and the focused regression tests in `crates/cli/tests/coverage/shared/config_tests.rs`.

- Relates to: none

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bootstrap security key handling during installation.
  * Existing keys are reused only when their files and directories have appropriate private permissions.
  * Insecure permissions are automatically repaired, or keys are safely recreated when necessary.
  * Invalid bootstrap keys are rejected before installation begins.
  * Dry-run installations continue without initializing or modifying the bootstrap key.
  * Permission checks and repairs now work consistently across supported platforms.
  * Valid existing keys retain their contents when permissions are corrected.


willkill07 (approved):
Nice!

some of the coderabbit feedback seems reasonable to address. non-blocking approval

willkill07 (approved):
LGTM

PR #776, opened 2026-08-13
ci: use system Python for uv workflows

willkill07:
#### Overview

Prevent CI from downloading managed Python through uv during workflow startup, keep Rust dependency caches profile-specific, and simplify Linux native-package builds.

#### Details

- Configure uv to use only the interpreter provided by the runner.
- Replace managed `uv python install` steps with pinned `actions/setup-python` v7 using the centralized CI Python version.
- Rely on `setup-python` built-in PATH setup and host-architecture default instead of duplicating either behavior in the workflows.
- Build musllinux Python wheels with pinned `PyO3/maturin-action`, retaining the pinned PyPA images and CI-derived package version.
- Build Linux Node glibc and musl packages from the existing AMD64 and ARM64 runners through one target-driven Zig path; remove the bespoke musllinux Docker bootstrap.
- Split Rust caches by debug and release builds, and save them only from `main` to avoid branch-scoped cache duplication.

Review the Linux package matrix in `.github/workflows/ci_node.yml` and the target-aware `package-node` recipe in `justfile`.

- Relates to: none

## Summary by CodeRabbit

* **Chores**
  * Standardized CI workflows on configured system Python runtimes.
  * Updated Python setup actions and streamlined platform-specific configuration.
  * Improved build caching by separating debug and release artifacts.
  * Limited cache publishing to the main branch where applicable.
  * Streamlined musllinux package preparation for more consistent wheel builds.
  * Improved Node package builds with configurable cross-compilation targets and build strategies.
  * Removed redundant platform-specific packaging steps for simpler, more reliable builds.


.github/workflows/ci_python.yml
```diff
@@ -155,7 +122,7 @@ jobs:
         run: just test-python-plugin-e2e
 
       - name: Run Python LangChain integration tests
-        if: ${{ inputs.run_integration_langchain }}
+        if: ${{ inputs.run_integration_langchain && matrix.platform != 'windows-arm64' }}
```
bbednarski9, line 125: Did we remove Windows ARM64 OpenSSL above in order to make CI pass without adding it back in?

bbednarski9 (dismissed):

.github/workflows/ci_python.yml
willkill07: I removed it because of the amount of time it wasted for little coverage gain. We already test on all other platforms.

mnajafian-nv (approved):
Lgtm

PR #791, opened 2026-08-14
fix: preserve symlinked agent config files

mnajafian-nv:
#### Overview

Preserve original Claude and Codex host symlink state across Relay install, reinstall, restore, and uninstall flows without writing Codex bootstrap secrets through a user-managed symlink target.

#### Details

- record symlink targets in CLI file snapshots so restore flows can recreate the original link instead of replacing it with a regular file
- add a cross-target fallback for snapshot symlink creation and strengthen snapshot restore coverage for recreated and dangling symlinks
- persist the original Codex config symlink target in the Relay-owned backup so uninstall can recreate the original symlink and restore its prior target bytes or dangling state
- materialize the active Codex install as a private local config file instead of writing the bootstrap client token through a symlink into a user-managed target
- preserve the original Codex config symlink target across reinstall backup refreshes so later uninstall still restores the original link
- prune the test-only dangling-hooks marker path and keep symlink coverage on the real legacy Codex hooks migration flow
- keep Claude symlink-preserving behavior unchanged

Validation:
- `cargo fmt --all`
- `cargo test -p nemo-relay-cli snapshot_restore_preserves_symlink_identity_and_target_bytes -- --nocapture`
- `cargo test -p nemo-relay-cli snapshot_restore_recreates_dangling_symlink_and_removes_materialized_target -- --nocapture`
- `cargo test -p nemo-relay-cli preserve_symlinked -- --nocapture`
- `cargo test -p nemo-relay-cli codex_reinstall_restores_original_symlink_after_user_edits_materialized_config -- --nocapture`
- `cargo test -p nemo-relay-cli codex_install_migration_preserves_symlinked_legacy_hooks_path -- --nocapture`
- `cargo test -p nemo-relay-cli --lib -- --nocapture`

Start in `crates/cli/src/agents/codex/host.rs`, especially the Codex config backup, install, and uninstall paths. Then review `crates/cli/tests/coverage/agents/plugin_host_tests.rs` for the new symlinked Codex config and legacy-hooks coverage, plus `crates/cli/src/filesystem/snapshots.rs` for the shared symlink restore helper.

- Relates to: none

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved symbolic links when installing, updating, restoring, and uninstalling configuration files and hooks.
  * Correctly handled dangling links and files that were originally absent.
  * Restored Claude and Codex settings without replacing symlink paths.
  * Prevented backup links from redirecting updates or deletions to unrelated files.

* **Tests**
  * Added Unix coverage for symlink preservation, target updates, dangling links, user edits, and snapshot restoration.


afourniernv:
One suggestion: would it make sense to prune the new dangling-hooks marker path? Current Relay no longer installs hooks into `~/.codex/hooks.json`; it only cleans up legacy entries, and the marker is written only by test code. We could keep the symlink handling for legacy-hook cleanup, remove the marker-specific logic, and test the real `remove_legacy_codex_hooks` path instead. The core fix still looks good either way, but this may keep the PR tighter.

Otherwise, LGTM.

crates/cli/src/agents/codex/host.rs
```diff
@@ -1290,10 +1312,64 @@ pub(crate) fn uninstall_codex_hooks(path: &Path, _gateway_url: &str) -> Result<b
...
+fn codex_backup_marks_original_config_absent(backup: &DocumentMut) -> bool {
+    backup
+        .get(ABSENT_CONFIG_BACKUP_KEY)
+        .and_then(Item::as_value)
+        .and_then(TomlValue::as_bool)
+        == Some(true)
+}
+
+#[cfg(test)]
+fn write_absent_codex_hooks_backup(path: &Path) -> Result<(), String> {
```
yczhang-nv: Codex hooks dangling-symlink support still never runs in release: write_absent_codex_hooks_backup and its only caller install_codex_hooks (:1280) are both #[cfg(test)]. Untouched by the new commits.

crates/cli/src/agents/codex/host.rs
```diff
@@ -791,7 +797,7 @@ pub(crate) fn install_codex_config(path: &Path, gateway_url: &str) -> Result<(),
         merge_codex_provider_extensions(&mut doc, extensions);
     }
 
-    if let Err(error) = atomic_write_private(path, doc.to_string().as_bytes()) {
+    if let Err(error) = atomic_write_private_preserving_symlink(path, doc.to_string().as_bytes()) {
```
yczhang-nv: The bearer token in http_headers is now written through the symlink, i.e. into the user's dotfiles repo. Location, not permissions (still 0600).

yczhang-nv (approved):
Approved with comments.

crates/cli/src/agents/claude/host.rs
```diff
@@ -121,7 +123,7 @@ pub(crate) fn restore_claude_provider(gateway_url: &str) -> Result<(), String> {
                 }
             }
         } else {
-            write_json(&path, &settings)?;
+            write_json_preserving_symlink(&path, &settings)?;
```
yczhang-nv, line 134: Reverting Claude restore to plain fs::remove_file(&path) regressed the dangling-symlink case. I ran it: with ~/.claude/settings.json a dangling symlink, enable follows the link and creates the target, then restore deletes the symlink and leaves the target behind still holding {"env":{"ANTHROPIC_BASE_URL":"..."}}. The user loses their symlink and gains a stray relay-written file in their dotfiles — main never created that file at all. Root cause: the ABSENT marker doesn't record whether the path was a symlink, so restore can't distinguish this from the case claude_restore_absent_original_does_not_delete_symlink_target protects, and takes the conservative branch for both. Codex already solves this with path_is_dangling_symlink + write_absent_codex_config_backup; Claude needs the same signal in its marker. Also worth noting the PR overview still claims dangling-state preservation for Claude.

yczhang-nv (approved):

crates/cli/src/agents/codex/host.rs
mnajafian-nv: Addressed in the latest updates. I removed the dead test-only dangling-hooks marker path and kept the symlink handling focused on the legacy hooks cleanup flow that still runs in release.

crates/cli/src/agents/codex/host.rs
mnajafian-nv: Addressed in the latest updates. Relay-owned Codex config writes no longer follow the user symlink target, so the bootstrap header stays in the managed config path instead of being written into the linked dotfiles target.

mnajafian-nv:
Addressed. I removed the dead dangling-hooks marker path and kept the symlink handling focused on the legacy Codex hooks cleanup flow.

crates/cli/src/agents/claude/host.rs
mnajafian-nv: Addressed in the latest update. Claude now records when the original settings path was a dangling symlink, and restore uses that marker to preserve the symlink path while removing the Relay-created target. I also added a regression for the exact dangling-symlink sequence you described.

crates/cli/src/agents/codex/host.rs
```diff
@@ -734,6 +737,60 @@ fn restore_codex_install_snapshots(snapshots: &[FileSnapshot]) -> Result<(), Str
...
+) -> Result<(), String> {
+    let target = symlink_target.to_str().ok_or_else(|| {
+        format!(
+            "failed to persist original Codex config symlink target {}: path is not valid UTF-8",
+            symlink_target.display()
+        )
+    })?;
+    backup[CONFIG_SYMLINK_TARGET_BACKUP_KEY] = value(target);
+    Ok(())
+}
```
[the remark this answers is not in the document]
mnajafian-nv: Addressed in the latest update. Codex now skips the optional symlink-target backup marker when the target path is not valid UTF-8, so install continues instead of failing. I also added a regression that exercises the non-UTF-8 dangling-symlink path.

yczhang-nv (approved):

PR #797, opened 2026-08-17
chore: install native Node packages on runtimes below the engines floor

yczhang-nv:
#### Overview

`npm install nemo-relay-node` on Node.js 22 or 23 exits 0 with only an `EBADENGINE` warning, silently omits the platform-specific native package, and then fails at `require()` with `Cannot find module 'nemo-relay-node-<triple>'`. npm omits an optional dependency whose `engines` do not match the running Node.js, and `scripts/package-node-bin.py` copied the metapackage `engines` into all seven native sub-package manifests.

This drops `engines` from the native sub-package manifests only. The metapackage keeps `"node": ">=24.0.0"`, which degrades to an advisory `EBADENGINE` warning. **Node.js 24 remains the supported runtime**; this only stops the install from producing a package that cannot load.

The same script also declared the `.node` binary as each native package's `main`. Runtimes that do not resolve a `.node` file as a package entry point cannot load `nemo-relay-node` by name at all, so native packages now ship a generated JavaScript entry point and declare that instead.

#### Details

* `scripts/package-node-bin.py`: `repository_metadata()` takes an `include_engines` flag; `build_native_package()` passes `False`. The metapackage is unchanged.
* `scripts/package-node-bin.py`: each native package gains a generated `index.js` that requires its `.node` binary, `main` points at it, and `files` carries both entries.
* `scripts/tests/test_package_node_bin.py`: asserts the native manifest has no `engines`, that the metapackage keeps its range, and that the generated entry point requires the binary.
* `.github/ci-tool-versions.env` and `.github/actions/load-ci-tool-versions/action.yml`: add `NEMO_RELAY_CI_NODE_COMPAT_VERSION=22`, exposed as `node_compat_version`. It is a compatibility target, not a supported floor.
* `.github/workflows/ci_node.yml`: `PackageSmoke` adds an install-and-load pass below the declared floor for the five `setup-node` rows, and the two musl rows now loop over a second digest-pinned `node:22-alpine3.22` image. New assertions cover both mechanisms this PR depends on: native packages must not declare `engines`, and native `main` must stay a JavaScript entry point.
* `crates/node/README.md` and `docs/about-nemo-relay/release-notes/index.mdx`: document that the `EBADENGINE` warning on the metapackage is the only signal a user gets on Node.js 22 or 23, with the warning quoted verbatim.

Validation, on `darwin-arm64`:

* Reproduced the original failure from the public registry on Node 22.23.2 — `npm install nemo-relay-node@0.7.3` warned `EBADENGINE`, installed 1 package, and `require()` threw at `index.js:308`.
* With this change, packed artifacts installed through the real `optionalDependencies` path add 2 packages on Node 22.23.2 and 23.11.0, all seven entry points `require()`, and a live `pushScope` / `event` / `popScope` plus `flushSubscribers` emits 3 events. Node 24.19.0 installs with no warning.
* `just test-node` on Node 24.19.0: 375/375.
* `scripts/tests/test_package_node_bin.py` passes; `pre-commit` is clean on every changed file, including actionlint.
* The other six platform packages are unexercised locally and rely on `PackageSmoke`.

`scripts/package-node-bin.py`, at `repository_metadata()` and `build_native_package()`. Both behaviors this PR changes live there, and the trade-off worth reviewing is deliberate: the metapackage now declares a floor it does not enforce, so anyone using `engine-strict=true`, or a package manager stricter than npm about a parent's `engines`, still cannot install on Node.js 22 or 23. The alternative considered was lowering the published floor to `>=22.0.0`, which was rejected to keep the supported runtime at Node.js 24.

Then `.github/workflows/ci_node.yml`, where the two new assertions are what stop either mechanism from silently regressing.

- Relates to: RELAY-241 (internal tracker). No public GitHub issue exists for this work.


yczhang-nv:
Superseded by #798, which is the same engines fix on a branch without an internal tracker identifier.

Changes from this PR, all from review feedback:
- The native entry-point change is split out. The Bun rationale did not hold as written, and moving `main` off the binary changes `require.resolve()` for consumers that use it to locate the native file. It will be pursued separately with a concrete failing runtime and its own smoke test.
- The new PackageSmoke pass now installs only the metapackage and resolves the platform package through `optionalDependencies`, so it actually exercises the path this bug lives on. Verified to fail on pre-fix artifacts.
- Retitled to `fix(node):`, since this is a user-visible packaging bug.

PR #798, opened 2026-08-17
fix(node): install platform packages on runtimes below the engines floor

yczhang-nv:
#### Overview

`npm install nemo-relay-node` on Node.js 22 or 23 exits 0 with only an `EBADENGINE` warning, silently omits the platform-specific package, and then fails at `require()` with `Cannot find module 'nemo-relay-node-<triple>'`. npm omits an optional dependency whose `engines` do not match the running Node.js, and `scripts/package-node-bin.py` copied the metapackage `engines` into all seven platform manifests.

This drops `engines` from the platform manifests only. The metapackage keeps `"node": ">=24.0.0"`, which degrades to an advisory `EBADENGINE` warning. **Node.js 24 remains the supported runtime**; this only stops the install from producing a package that cannot load.

The trade-off, stated plainly: the metapackage now declares a floor it does not enforce. Anyone using `engine-strict=true`, or a package manager stricter than npm about a parent's `engines`, still cannot install on Node.js 22 or 23. The alternative considered was lowering the published floor to `>=22.0.0`, which was rejected to keep the supported runtime at Node.js 24.

**Merge timing: this PR is intentionally not urgent.** It is install ergonomics, not a capability, and nothing in the dependent integration work is blocked by it. Consumers on Node.js 22 or 23 can get a working install today by naming the platform package explicitly, which makes it a direct dependency that npm does not omit:

```bash
npm install nemo-relay-node nemo-relay-node-linux-x64-gnu
```

Verified against the published `0.7.3` on Node 22.23.2: two `EBADENGINE` warnings, both packages installed, `require()` succeeds. Running on Node.js 24 avoids the issue entirely. This PR can therefore sit until the end of the integration work it came from, and should be reviewed on its merits rather than on schedule pressure.

#### Details

* `scripts/package-node-bin.py`: `repository_metadata()` takes an `include_engines` flag; `build_native_package()` passes `False`. The metapackage manifest is unchanged.
* `scripts/tests/test_package_node_bin.py`: asserts the platform manifest has no `engines` and that the metapackage keeps its range.
* `.github/ci-tool-versions.env`, `.github/actions/load-ci-tool-versions/action.yml`: add `NEMO_RELAY_CI_NODE_COMPAT_VERSION=22`, exposed as `node_compat_version`. It is a compatibility target, not a supported floor.
* `.github/workflows/ci_node.yml`: `PackageSmoke` gains a pass on the compatibility version. It installs **only the metapackage** and lets npm reach the platform package through `optionalDependencies`, with the packed tarball wired in via an `overrides` entry. Installing both tarballs directly, as the existing pass does, makes the platform package a direct dependency and cannot detect an `engines` regression. The two musl rows run the same check against a second digest-pinned `node:22-alpine3.22` image.
* `crates/node/README.md`, `docs/about-nemo-relay/release-notes/index.mdx`: document the install behavior on Node.js 22 and 23, with the `EBADENGINE` warning quoted verbatim. The warning is the only signal under npm's default `engine-strict=false`; with `engine-strict=true`, or a package manager that enforces the parent package's engine range, installation fails instead. Both versions remain unsupported either way.

Validation, on `darwin-arm64`:

* Reproduced the original failure from the public registry on Node 22.23.2 — `npm install nemo-relay-node@0.7.3` warned `EBADENGINE`, installed 1 package, and `require()` threw at `index.js:308`.
* With this change, packed artifacts installed through `optionalDependencies` add 2 packages on Node 22.23.2 and 23.11.0, all seven entry points `require()`, and a live `pushScope` / `event` / `popScope` plus `flushSubscribers` emits 3 events. Node 24.19.0 installs with no warning.
* The new CI pass was run locally against both artifact shapes: it exits 0 on the fixed artifacts and fails with `nativePackages.length actual: 0, expected: 1` on pre-fix artifacts, so it is a real install-path regression test rather than a manifest assertion. The musl variant was executed under POSIX `sh` against the same artifacts.
* `just test-node` on Node 24.19.0: 375/375. `scripts/tests/test_package_node_bin.py` passes. `pre-commit` is clean on every changed file, including actionlint.
* The other six platform packages are unexercised locally and rely on `PackageSmoke`.

`scripts/package-node-bin.py`, at `repository_metadata()` and `build_native_package()` — the whole behavior change is those two call sites.

Then the new `PackageSmoke` pass in `.github/workflows/ci_node.yml`, which is the part worth scrutiny: it deliberately installs the metapackage alone so that the `optionalDependencies` resolution path is exercised, because that is the path this bug lives on.

- Relates to: none. Tracked internally; no public issue exists for this work.


yczhang-nv:
Closing: the integration that motivated this no longer needs the Node.js binding at all. The pi plugin will talk to the NeMo Relay CLI over the gateway, the same way Codex and Claude Code do, so `nemo-relay-node` is off that path entirely and the Node 22 install problem stops being a blocker for it.

The underlying bug is still real for anyone installing `nemo-relay-node` on Node.js 22 or 23: npm omits the platform package as an unmatched optional dependency, `npm install` exits 0, and `require()` then fails with `Cannot find module 'nemo-relay-node-<triple>'`. Workaround without any package change — name the platform package explicitly so it is a direct dependency:

```bash
npm install nemo-relay-node nemo-relay-node-linux-x64-gnu
```

Verified against the published 0.7.3 on Node 22.23.2.

The branch `fix/node-native-package-engines` keeps the work if anyone wants to revive it: platform manifests drop `engines`, plus a PackageSmoke pass that resolves through `optionalDependencies` and was confirmed to fail on pre-fix artifacts.

PR #800, opened 2026-08-17
feat(core): add Event metadata injection middleware

ericevans-nv:
#### Overview

Relay Events currently have no general middleware point for adding cross-cutting metadata after an Event is constructed. This change adds an insert-only Event metadata injection stage so Rust callers and native plugins can attach validated attributes to every Scope-start, Scope-end, and Mark Event before subscriber and exporter delivery.

Injector callbacks receive the constructed Event as immutable context and return proposed metadata additions. Relay validates and merges those additions into `Event.metadata` without allowing existing metadata or other Event fields to be overwritten.

Follow-up documentation: [PR #801](https://github.com/NVIDIA/NeMo-Relay/pull/801).

- [x] I confirm this contribution is my own work, or I have the right to submit it under this project's license.
- [x] I searched existing issues and open pull requests, and this does not duplicate existing work.

#### Details

This PR:

- Adds an asynchronous `EventMetadataInjectorFn` callback that returns metadata key/value additions.
- Supports global, scope-local, and native-plugin injector registration.
- Snapshots the injectors visible to an Event when that Event is emitted.
- Runs injectors sequentially on the queued publication path before Event sanitizers.
- Applies injectors to every canonical Scope-start, Scope-end, and Mark Event.
- Orders injectors deterministically by numeric priority and then registration name.
- Preserves existing metadata when an injector returns a duplicate key. When multiple injectors propose a new key, the first successful insertion wins.
- Accepts regular or dotted keys containing ASCII letters, numbers, underscores, and hyphens.
- Accepts strings, numbers, booleans, empty lists, and homogeneous primitive lists.
- Rejects malformed keys, nested objects, `null`, nested lists, and mixed-type lists.
- Isolates callback errors, panics, and invalid output so the original Event continues through sanitization and export.
- Cleans up plugin-owned and scope-local registrations through the existing ownership lifecycle.

Focused validation covered the callback contract, key and value validation, deterministic ordering, insert-only merging, callback failure isolation, plugin cleanup, scope cleanup, and all canonical Event records.

This PR does not add language-binding or gRPC registration surfaces, and it does not promote selected metadata into top-level OpenTelemetry span attributes.

Validation performed:

- `cargo test -p nemo-relay event_metadata_injection`
- `cargo test -p nemo-relay --test api_surface_integration event_metadata_injectors_are_insert_only_ordered_and_failure_safe -- --exact`
- `cargo fmt --all --check`
- `cargo clippy -p nemo-relay --all-targets --all-features -- -D warnings`
[...]

PR #802, opened 2026-08-18
feat(observability): promote selected Event metadata to OTel attributes

ericevans-nv:
#### Overview

Relay already exports Event metadata under lifecycle-specific Relay attribute names. This change adds an opt-in `promote_metadata_prefixes` setting that also exposes matching metadata under its original key as a top-level OpenTelemetry attribute.

The option defaults to an empty list, so existing OpenTelemetry output remains unchanged unless promotion is explicitly configured.

Follow-up documentation: [PR #803](https://github.com/NVIDIA/NeMo-Relay/pull/803).

#### Details

- Adds `promote_metadata_prefixes` to each OpenTelemetry trace endpoint.
- Treats every configured value as a case-sensitive literal string prefix with no implicit dot or metadata-key segment boundary. For example, `nv.` selects keys beginning with `nv.`, `nv_` selects keys beginning with `nv_`, and `user` selects keys such as `username` and `user_api_key`.
- Copies matching metadata into top-level attributes across Full, GenAI, and OpenInference projections.
- Preserves start-only Scope metadata. When the same key appears at Scope start and end, Scope-end metadata is authoritative.
- If the authoritative Scope-end value cannot be represented as an OpenTelemetry attribute, omits that promoted attribute and reports a diagnostic instead of falling back to the earlier Scope-start value.
- Promotes Mark metadata for projections that emit Marks.
- Supports strings, booleans, signed integers, floating-point numbers, and homogeneous primitive arrays.
- Omits unsupported values without dropping the Event or exposing the rejected value.
- Prevents promoted metadata from replacing projection-owned attributes or configured aliases.
- Leaves the original Event, ATOF output, and existing Relay-prefixed attributes unchanged.
- Exposes the option through Rust, plugin configuration, Python, Node.js, Go, and C/FFI.

Validation completed:

- Focused Rust promotion tests passed.
- Focused C/FFI tests passed.
- `just test-python` passed 682 tests.
- `just test-node` passed 386 tests.
- `just test-go` passed.
- Changed-file pre-commit validation passed after rebasing onto current `main`, including Ruff, Python type checking, FFI header synchronization, Cargo formatting, Clippy, Cargo check, Go formatting and vet, and Node formatting.

Start with `crates/core/src/observability/mod.rs` for prefix validation and typed conversion, followed by `crates/core/src/observability/otel.rs` for Scope and Mark lifecycle behavior and collision handling.

- Relates to: none

## Summary by CodeRabbit

- **New Features**
  - Added configurable promotion of selected event metadata into OpenTelemetry span attributes across Python, Node.js, Go, and native integrations.
  - Final-event values take precedence while preserving projection-owned and protected attributes.
  - Supports scalar and homogeneous primitive-array metadata values.

- **Bug Fixes**
  - Added diagnostics for unsupported, reserved, conflicting, or unrepresentable metadata without dropping events or spans.

- **Validation**
  - Rejects blank, malformed, duplicate, whitespace-padded, and wildcard prefixes.


bbednarski9:
if we are going to do the language bindings async from this PR, we might want to include docs in this PR to make sure that lands for the configuration changes in time for 0.8.0

crates/core/src/observability/otel.rs
```diff
@@ -1200,11 +1239,30 @@ impl OtelEventProcessor {
         if self.otel_type != OpenTelemetryType::GenAi && is_trace_root {
             push_session_identity_attributes(&mut attributes, event);
         }
+        // Snapshot keys claimed by the projection so promoted metadata cannot
+        // replace them when the span completes.
+        let mut projection_attribute_keys = attributes
```
bbednarski9, line 1244: I think we need to improve the validation of metadata keys versus projection/attribute keys that are both namespaced and borrowed by OTEL. In this code, metadata wins over projection attributes IFF they are missing on a per-span basis. However, the reserved namespace is significant. Claude found:

```
nemo_relay.*          (13 keys — Relay owns the whole namespace)
gen_ai.*              (21 keys — GenAI projection owns the whole namespace)
error.type            exception.type
input.mime_type       input.value
output.mime_type      output.value
llm.cost.total        llm.invocation_parameters   llm.model_name   llm.provider
llm.token_count.{prompt,completion,total}
metadata                                    ← bare key, no dot
openinference.span.kind
service.{name,namespace,version}
session.id            user.id
tool.name             tool.parameters
tool_call.id          tool_call.function.{name,arguments}
server.address        server.port           (via semconv)
```

In fact, this same issue might exist for validate_attribute_mappings (mod.rs:336)... So im not sure this is a blocker or a follow-up as we dont have a great pattern in place to date for namespace validation

crates/core/src/observability/otel.rs
```diff
@@ -1282,6 +1342,25 @@ impl OtelEventProcessor {
...
         }
+        // Preserve every projection-owned start/end key while promoting the
+        // final metadata carried by the scope-end Event.
+        active_span.projection_attribute_keys.extend(
+            attributes
+                .iter()
+                .map(|attribute| attribute.key.as_str().to_string()),
+        );
+        let end_metadata = event.metadata().and_then(crate::json::Json::as_object);
+        active_span.start_promoted_metadata.retain(|attribute| {
```
bbednarski9, line 1354: Codex comment: In the scope-end path (otel.rs:1345-1360), start_promoted_metadata.retain(...) drops any key present in the end Event's metadata map, then promote_metadata runs on the end Event. If the end value is a null/object/empty-array/oversized-u64, the conversion fails — so the key is dropped from the start set and fails to promote from the end. The attribute vanishes entirely and emits a warning. The PR body promises "uses the final Scope-end value when the same key appears at both stages," which is ambiguous about this case; either fall back to the start value or state the drop is intentional. No test covers it.

bbednarski9, line 1354: **A promoted key is dropped entirely when the Scope-end value is unsupported.**

`retain` removes any start-promoted key that appears in the end Event's metadata map, on the assumption that `promote_metadata` just below will re-promote it from the end Event. But `contains_key` only proves the key is *present* -- not that its value converts. If the end value is null, an object, an empty array, or a u64 above `i64::MAX`, `metadata_value_to_otel` returns `Err`, so the key is dropped from the start set *and* fails to promote from the end. The attribute disappears from the span, and the only trace is a `metadata_promotion_value_unsupported` warning.

Concretely: `{"nv.tenant": "acme"}` at Scope start and `{"nv.tenant": null}` at Scope end produces no `nv.tenant` attribute at all, even though a valid value was available at start.

The PR description says promotion "uses the final Scope-end value when the same key appears at both stages," which is ambiguous for this case. If the intent is last-write-wins among *promotable* values, the retain should be gated on convertibility rather than presence. If the drop is intentional, it's worth a doc note -- it's surprising that adding an unpromotable value at Scope end deletes a valid attribute that was set at Scope start.

Either way this path currently has no test, which is part of the 80.69% patch-coverage gap.

crates/core/src/observability/mod.rs
```diff
@@ -485,6 +528,150 @@ fn push_top_level_json_value(
...
+        return Vec::new();
+    };
+    let mut existing_keys = attributes
+        .iter()
+        .map(|attribute| attribute.key.as_str().to_string())
+        .chain(protected_keys.iter().cloned())
+        .collect::<std::collections::HashSet<_>>();
+    let mut issues = Vec::new();
+    for (key, value) in metadata {
+        if !prefixes.iter().any(|prefix| key.starts_with(prefix)) || existing_keys.contains(key) {
```
bbednarski9: **`starts_with` has no segment boundary, so a prefix matches more than it looks like it matches.**

`is_valid_metadata_promotion_prefix` makes the trailing dot optional, so `"user"` is a valid configured prefix -- and this line then promotes `username`, `user_id`, and `user_api_key` alongside the intended `user.*` keys. Same for `"session"` picking up `session_token`.

This is new surface relative to `attribute_mappings`, which maps an exact key to an exact alias and can't over-match. Given that the feature's job is copying caller-controlled metadata into exported telemetry, an operator who reasons "I want my `user.` namespace promoted" and writes `user` exports more than intended, and the over-match is silent.

Suggest requiring a segment boundary: match when `key == prefix_without_dot` or `key.starts_with(&format!("{prefix_without_dot}."))`. That keeps the intended `["nv."]` / `["nv"]` usage working identically while making over-match impossible.

If literal substring matching is deliberate, that's a defensible call -- but it should be stated explicitly in the endpoint docs, since the field name reads as namespace selection.

bbednarski9 (commented):
Two notes on the promotion logic itself. Separately tracking the failing Codecov statuses and the missing endpoint docs.

bbednarski9 (changes requested):

crates/core/src/observability/plugin_component.rs
```diff
@@ -731,6 +734,12 @@ impl EditorConfig for OpenTelemetryEndpointConfig {
                 ),
                 otel_editor_field("mark_exclude_names", EditorFieldKind::Json, &[], false),
                 otel_editor_field("attribute_mappings", EditorFieldKind::List, &[], false),
+                otel_editor_field(
+                    "promote_metadata_prefixes",
+                    EditorFieldKind::List,
+                    &[],
+                    true,
+                ),
```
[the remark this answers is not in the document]
ericevans-nv: Added explicit editor-schema assertions for the field’s `List` type and optional status.

crates/core/src/observability/otel.rs
ericevans-nv: Good catch. Metadata promotion now rejects keys owned by Relay or the supported OTel projections, even when a particular span does not emit that key. This protects `nemo_relay.*`, `gen_ai.*`, the other borrowed semantic namespaces, and the bare `metadata` key while continuing to allow configured custom namespaces such as `nv.*`.

I added regression coverage for every reserved namespace and documented the behavior. I left the existing `attribute_mappings` behavior unchanged because applying this restriction there requires a separate compatibility decision.

crates/core/src/observability/otel.rs
ericevans-nv: This behavior is intentional. Scope-end metadata is authoritative when the same key appears at Scope start and end. If the final value cannot be represented as an OpenTelemetry attribute, Relay omits the promoted attribute and reports a diagnostic rather than restoring a stale Scope-start value.

I added regression coverage across Full, GenAI, and OpenInference and clarified this behavior in the PR description.

ericevans-nv: Thanks for raising this. This is the same Scope-end case addressed in [the earlier thread](https://github.com/NVIDIA/NeMo-Relay/pull/802#discussion_r3808153003), where we clarified the intended behavior and added regression coverage across Full, GenAI, and OpenInference. I’m resolving this duplicate in favor of that discussion.

crates/core/src/observability/mod.rs
ericevans-nv: Thanks for calling out the potential ambiguity for operators. Literal string-prefix matching is intentional here and is not limited to dotted namespaces. For example, `nv.` selects keys beginning with `nv.`, `nv_` selects keys beginning with `nv_`, and a broader prefix such as `user` intentionally selects all keys beginning with `user`.

I kept the existing matching behavior, added regression coverage for dotted, underscored, and broader prefixes, and clarified that matching is literal and case-sensitive.

bbednarski9:
could you check the validation of metadata between #800 and #802? I think they might disagree in some cases:

1) Opposite precedence. #800 is first-wins, #802 is last-wins (Scope-end beats Scope-start). Both hardcoded, neither configurable, no shared rationale stated anywhere.

2) Contradictory type contracts. #800's validator is literally named is_otel_compatible_attribute_value (state.rs:1796):

Json::Array(values) => match values.first().and_then(primitive_kind) {
    None => values.is_empty(),   // <- empty arrays ARE OTel-compatible

#802's metadata_array_to_otel disagrees:

let Some(first) = values.first() else {
    return Err("empty arrays do not declare an OTLP element type");
};

3) Same for large unsigned integers — #800 accepts any Json::Number; #802 rejects u64 above i64::MAX.

So an injector that adds nv.tags = [] passes #800's validation as OTel-compatible, lands in Event.metadata, and then #802 refuses to promote it and fires otel_metadata_promotion_value_unsupported on every event carrying it. One merged feature blesses a value the other treats as a warning-worthy error, and the operator gets recurring log noise for doing exactly what #800 permits.

bbednarski9:
If comment above can be resolved, then can we share this helper? 

#800 merged a key-syntax validator, and #802 duplicates it rather than calling it:

│ state.rs:1787 (#800, merged) │ is_valid_event_metadata_attribute_key │
│ mod.rs:372 (#802, open)      │ is_valid_metadata_promotion_prefix    │

ericevans-nv:
@bbednarski9 Thanks for the detailed review.

> if we are going to do the language bindings async from this PR, we might want to include docs in this PR to make sure that lands for the configuration changes in time for 0.8.0

Agreed that the configuration documentation needs to land with the feature. I’ve updated the [OpenTelemetry metadata promotion documentation](https://github.com/NVIDIA/NeMo-Relay/pull/803/files#diff-e1b9e6a1678fb21cc178023af6fef4586269199de15a632046760929c452db76) in [PR #803](https://github.com/NVIDIA/NeMo-Relay/pull/803) to reflect the finalized behavior. When you have a chance, could you please review the changes for inclusion in 0.8.0?

> could you check the validation of metadata between #800 and #802? I think they might disagree in some cases

Thanks for catching these inconsistencies. I aligned the contracts in [`496a8fe0`](https://github.com/NVIDIA/NeMo-Relay/commit/496a8fe078b6d6b34ae89d6e470ed84d51ddc64d):

- Empty arrays are now promoted successfully.
- Injector validation rejects unsigned scalar and array values above `i64::MAX` before insertion.
- Focused positive and negative tests cover both boundaries.

The precedence rules apply at different stages. “First wins” determines which injector can insert a key into one Event. Scope-end precedence applies later, when OpenTelemetry combines the separate Scope-start and Scope-end Events into one span. I’ll make that distinction explicit in [PR #803](https://github.com/NVIDIA/NeMo-Relay/pull/803).

> If comment above can be resolved, then can we share this helper?

Agreed. [`496a8fe0`](https://github.com/NVIDIA/NeMo-Relay/commit/496a8fe078b6d6b34ae89d6e470ed84d51ddc64d) moves the shared metadata-key syntax validation into one core helper. Exact Event metadata keys and promotion prefixes now use the same validation logic; prefix validation only removes the optional trailing dot before calling it.

bbednarski9 (approved):

PR #840, opened 2026-08-21
fix: report metadata promotion diagnostics per key

ericevans-nv:
#### Overview

Preserve the name of every metadata attribute rejected during OpenTelemetry promotion. Runtime reports now expose one stable diagnostic per rejected key instead of collapsing all failures into one entry whose message names only the last key.

#### Details

- Suffix the existing `otel.metadata_promotion_value_unsupported` diagnostic prefix with the rejected metadata key.
- Record rejected keys in deterministic order.
- Allow runtime diagnostic storage to own dynamically generated diagnostic codes.
- Add regression coverage proving that four rejected keys remain independently identifiable across scope start and end events.
- Update the existing unsupported-value test for the key-specific diagnostic code.

This changes exact diagnostic-code matching from:

`otel.metadata_promotion_value_unsupported`

to:

`otel.metadata_promotion_value_unsupported.<metadata-key>`

Consumers that match diagnostic codes exactly should use the key-specific code or match the existing prefix.

Validation:

- Cursor public Python smoke passed with four independently named diagnostics.
- Focused metadata-promotion regression tests passed.
- `cargo fmt --all` passed.
- File-scoped pre-commit checks and commit hooks passed.
- `just test-rust` was not fully green in the local environment:
  - One run had two CLI host-status failures because Claude was observed as running.
  - A later aggregate run had shared plugin-registration failures; 1,502 core tests passed and 24 failed. The metadata-promotion-adjacent failure passed when run independently.
- The Python suite was stopped before completion at the requested push boundary.
- Go, Node.js, and all-file pre-commit validation were not run.

Start with `OtelEventProcessor::promote_metadata` in `crates/core/src/observability/otel.rs`, then review `reports_each_unsupported_metadata_key_with_a_deterministic_diagnostic_code` in `crates/core/tests/unit/observability/otel_tests.rs`.

- Fixes RELAY-763

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved observability diagnostics for unsupported metadata.
  * Diagnostics are now consistently ordered and identify the specific metadata key involved.
  * Unsupported metadata continues to be excluded from span attributes while producing clearer diagnostic details.

* **Documentation**
  * Documented diagnostic codes, messages, occurrence counts, and monitoring guidance for rejected metadata values.

### DIFF
--- a/crates/core/src/observability/otel.rs
+++ b/crates/core/src/observability/otel.rs
@@ -1485,16 +1485,19 @@ impl OtelEventProcessor {
         event: &Event,
         protected_keys: &HashSet<String>,
     ) {
-        let issues = promote_event_metadata_attributes(
+        let mut issues = promote_event_metadata_attributes(
             attributes,
             event,
             &self.promote_metadata_prefixes,
             protected_keys,
         );
 
+        issues.sort_by(|left, right| left.key.cmp(&right.key));
         for issue in issues {
+            let diagnostic_code =
+                format!("otel.metadata_promotion_value_unsupported.{}", issue.key);
             let diagnostic_count = self.runtime_diagnostics.record(
-                "otel.metadata_promotion_value_unsupported",
+                diagnostic_code,
                 format!(
                     "OpenTelemetry metadata attribute {:?} was not promoted: {}",
                     issue.key, issue.reason
--- a/crates/core/src/observability/otel_signal.rs
+++ b/crates/core/src/observability/otel_signal.rs
@@ -57,7 +57,7 @@ impl OpenTelemetryRuntimeDiagnostics {
 
 #[derive(Debug, Default)]
 struct RuntimeDiagnosticState {
-    diagnostics: BTreeMap<&'static str, OpenTelemetryRuntimeDiagnostic>,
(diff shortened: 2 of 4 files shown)


crates/core/src/observability/otel.rs
```diff
@@ -1485,16 +1485,19 @@ impl OtelEventProcessor {
...
             protected_keys,
         );
 
+        issues.sort_by(|left, right| left.key.cmp(&right.key));
         for issue in issues {
+            let diagnostic_code =
+                format!("otel.metadata_promotion_value_unsupported.{}", issue.key);
             let diagnostic_count = self.runtime_diagnostics.record(
-                "otel.metadata_promotion_value_unsupported",
+                diagnostic_code,
```
willkill07, line 1500: With this change, does any documentation need to be adjusted?

willkill07 (approved):
Approving with one comment

crates/core/src/observability/otel.rs
ericevans-nv: Good catch. Yes, the OpenTelemetry documentation should be updated to describe the new key specific diagnostic code format.

willkill07 (approved):

PR #842, opened 2026-08-21
fix(observability): isolate OpenTelemetry endpoint activation

bbednarski9:
#### Overview

Keep valid OpenTelemetry trace endpoints active when another endpoint has invalid configuration or cannot construct its exporter, while retaining indexed diagnostics for invalid batch settings.

#### Details

- Include endpoints[index] in invalid batch-setting diagnostics and classify them as observability.invalid_otel_endpoint warnings.
- Skip trace endpoints that fail batch validation or exporter construction; valid trace endpoints continue to activate.
- Resolve header_env values at activation: an endpoint with a missing value is skipped when a valid peer remains, while an all-invalid configuration is rejected.
- Reject activation only when no trace, log, or metric endpoint can be registered.
- Update regression coverage for mixed-validity, all-invalid, malformed-endpoint, and missing-header configurations.
- Document endpoint-isolation behavior.
- Validation: cargo fmt --all, git diff --check, and five focused Rust regressions passed. The GitHub Actions matrix is running for the latest commit.

Review the partial-activation path in crates/core/src/observability/plugin_component.rs, then the endpoint-isolation regressions in crates/core/tests/unit/observability/plugin_component_tests.rs.

- Relates to: RELAY-766

## Summary by CodeRabbit

* **Bug Fixes**
  * OpenTelemetry now skips invalid endpoints while continuing to activate valid trace, log, and metric endpoints.
  * Invalid batch settings, malformed destinations, and headers no longer prevent other valid endpoints from activating.
  * Delivery failures and exporter errors remain isolated between endpoints.
  * Activation fails clearly when no valid endpoints remain.
  * Warnings identify affected endpoints by index and provide specific diagnostics.

* **Documentation**
  * Updated observability guidance to describe endpoint-level failures, warnings, and activation behavior.


willkill07 (changes requested):
These changes look good but don't address:

> _With atomic activation, one bad endpoint disables all observability._

Can we still address this?

bbednarski9:
@willkill07  yes, it will be a breaking behavior change but new feature so all good

willkill07 (approved):

willkill07 (approved):

bbednarski9:
[P1] Endpoint isolation breaks when logs or metrics derive from traces. Signal endpoints are derived from the full trace config before failed traces are skipped ([plugin component (line 1386)](NeMo-Relay/crates/core/src/observability/plugin_component.rs:1386), [derivation (line 1698)](/NeMo-Relay/crates/core/src/observability/plugin_component.rs:1698)). A malformed second trace endpoint therefore aborts activation if logs or metrics is enabled without explicit endpoints—despite a valid first trace endpoint. This contradicts the documented endpoint-isolation contract ([docs (line 102)]/NeMo-Relay/docs/configure-plugins/observability/configuration.mdx:102)).

[P2] Endpoint indexes drift after a skip. The subscriber is built with the original index, but later resource-access and callback-panic logs enumerate the compacted subscriber list . With endpoints [valid 0, invalid 1, valid 2], operational logs can identify endpoint 2 as index 1.

crates/core/src/observability/plugin_component.rs
```diff
@@ -1364,6 +1364,22 @@ fn build_atif_storage(
     ))
 }
 
+#[derive(Clone)]
+struct IndexedOpenTelemetryResource<T> {
+    index: usize,
+    value: T,
```
willkill07: I feel like `value` should be `Arc<T>` rather than `T`

Alternatively, why not just have an `enum` on active vs skipped?

```rust
enum OpenTelemetryResource<T> {
    Active(Arc<T>),
    Skipped(String),
}
```

Then `IndexedOpenTelemetryResource` can just hold `OpenTelemetryResource<T>` instead of `T`.

Using two different lists to split doesn't make sense when we have pattern matching at our disposal.

willkill07 (changes requested):
Ownership semantics should be cleaned up a bit here. Mandating `Arc` as part of the type parameter is pretty ugly.

crates/core/src/observability/plugin_component.rs
ericevans-nv: Agreed. I replaced the separate active and skipped lists with one enum-based list. Each endpoint now keeps its original index and is marked as either active or skipped.

Implemented in `2c364a3c`.

willkill07 (approved):

PR #933, opened 2026-08-27
enhancement(observability): promote metadata to OTLP resources

willkill07:
#### Overview

#### Details

- Add configurable root Scope-start metadata promotion to OTLP resources, with retained pipelines per effective resource.
- Remove successfully promoted span and resource metadata from OTLP metadata projections while preserving unsupported and configured-resource-collision values. Relay Events and ATOF payloads remain unchanged.
- Expose endpoint configuration consistently in Rust, Python, Node.js, and the observability plugin; add tests and documentation.

- Review `crates/core/src/observability/otel.rs`, especially root-resource pipeline selection and metadata projection filtering.
- `crates/core/tests/unit/observability/otel_tests.rs` covers resource isolation, lifecycle flushing, and metadata filtering.

- Relates to: none

## Summary by CodeRabbit

* **New Features**
  * OpenTelemetry can promote selected root-level metadata to OTLP resource attributes.
  * Added resource metadata prefix configuration for Node.js and Python APIs.
  * Resource attributes persist across related telemetry with defined precedence and fallback behavior.
  * Promoted metadata is no longer duplicated in span or event attributes.

* **Documentation**
  * Added guidance on configuration, precedence, failure behavior, and cardinality.

* **Tests**
  * Expanded coverage for promotion, serialization, validation, and exported OTLP resource attributes.


crates/core/src/observability/otel.rs
```diff
@@ -1404,17 +1524,117 @@ impl OtelEventProcessor {
...
+            let tracer = provider.tracer(self.instrumentation_scope.clone());
+            self.dynamic_pipelines.insert(
+                key.clone(),
+                DynamicTracePipeline {
+                    provider,
+                    tracer,
+                    _runtime: runtime,
+                },
+            );
+        }
```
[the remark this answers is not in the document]
willkill07: This is intentionally non-bounded

mnajafian-nv (approved):
LGTM! Thanks for addressing CR comments

crates/core/src/observability/otel.rs
```diff
@@ -1404,17 +1531,117 @@ impl OtelEventProcessor {
...
+            &protected_keys,
+        );
+        self.record_metadata_promotion_issues(promotion.issues, "resource_metadata");
+        let key = canonical_resource_key(&attributes);
+        let base_key = canonical_resource_key(&configured_resource_attributes(config));
+        if key == base_key {
+            return self.tracer.clone();
+        }
+        if !self.dynamic_pipelines.contains_key(&key) {
+            let (provider, runtime) = match build_owned_tracer_provider_with_resource(
```
bbednarski9: Suspected locking condition here under higher cardinality...

When promote_resource_metadata_prefixes is configured, the first event carrying a new distinct root value triggers tracer_for_root → build_owned_tracer_provider_with_resource, which spawns an OS thread, boots a Tokio runtime, and blocks on a channel — all while holding the global processor Mutex acquired earlier in the callback (lines ~745-750). With any real cardinality of root values (which the docs explicitly warn about), every other concurrent request's trace events block on that same lock until the new pipeline finishes initializing. This produces recurring latency spikes across unrelated traffic, not just the triggering trace.

crates/core/src/observability/otel.rs
```diff
@@ -775,10 +803,25 @@ impl OpenTelemetrySubscriber {
     /// After a successful flush, runtime diagnostics include queue drops observed so far.
     pub fn force_flush(&self) -> Result<()> {
         flush_subscribers()?;
```
bbednarski9, line 816: Another possible issue from claude...

The ? on the main provider's force_flush() returns early before the loop over dynamic resource pipelines (line 819) runs — so if the main flush fails, or any one dynamic pipeline fails mid-loop, all remaining pipelines are skipped without being flushed. Any caller relying on force_flush before shutdown will lose buffered spans from otherwise-healthy pipelines.

crates/core/src/observability/otel.rs
```diff
@@ -1404,17 +1531,117 @@ impl OtelEventProcessor {
...
 
+    fn tracer_for_root(&mut self, event: &Event) -> SdkTracer {
+        let Some(config) = self.owned_config.as_ref() else {
+            return self.tracer.clone();
+        };
+        if config.promote_resource_metadata_prefixes.is_empty() {
+            return self.tracer.clone();
+        }
+
+        let mut attributes = configured_resource_attributes(config);
```
bbednarski9, line 1645: Can we reuse resource_metadata_protected_keys here instead of recomputing?

bbednarski9:
Kind of nit, but i think we should document the semantic change in the PR description now that we are removing the duplicate from the metadata if a specific field gets promoted

bbednarski9 (changes requested):

mnajafian-nv (approved):

bbednarski9 (approved):
tested this E2E in our use case. its working

PR #934, opened 2026-08-27
fix: harden security boundaries without API changes

willkill07:
#### Overview

Implements API-, ABI-, configuration-, and data-compatible hardening for the gateway, permission, dynamic-plugin, observability, and adaptive-cache threat-model findings.

#### Details

- Add hidden authentication for managed, transparent, and generated hook clients; bind sessions to authenticated client identities; strip internal credentials before events and forwarding; and reject unsafe explicit-daemon exposure.
- Route final permission requests through the existing tool conditional-execution chain and bind decisions to the exact active tool call, name, and arguments. Claude Code permission requests intentionally omit the call identifier, so Relay resolves it only when exactly one active pre-tool record matches and denies ambiguous or changed requests.
- Preserve Claude Code host permission semantics: `PreToolUse` no longer pre-authorizes the action, and Relay returns Claude's event-specific permission decision shape only after final evaluation.
- Preserve user settings while ensuring the transparent Relay gateway overlay has final precedence for the launched Claude process.
- Require signed, trusted dynamic plugins at runtime startup by default while preserving existing manifest schemas, lifecycle management APIs, and conditional-middleware fail-open semantics.
- Minimize inherited worker-process environment variables.
- Confine ATOF and ATIF output beneath configured destinations, reject traversal and symlink targets, and use private files, private directories, and atomic ATIF replacement.
- Document cache namespace tenancy and add partitioning coverage for namespaces, providers, and allowlisted tenant headers.
- Preserve all Rust, Python, Node.js, Go, C, hook, provider, event, cache, ATOF, and ATIF interfaces and schemas. There are no breaking API or ABI changes.

Validation:

- `RELAY_E2E_TRANSPARENT_ONLY=1 just test-codex-plugin-e2e` — real Codex CLI tool-call round trip, second provider turn, lifecycle hooks, and balanced ATOF tool scope
- `RELAY_E2E_TRANSPARENT_ONLY=1 just test-claude-plugin-e2e` — real interactive Claude Code `PreToolUse` → `PermissionRequest` allow → tool execution → `PostToolUse` round trip and balanced ATOF tool scope
- `just test-rust` — 4,407 workspace tests plus native-plugin, worker-plugin, and language-binding example suites
- `just test-python`
- `just test-node`
- `just test-go`
- `just docs`
- Targeted changed-file pre-commit validation after the E2E follow-up

Start with `crates/cli/src/server/mod.rs` for the hidden ingress-authentication boundary and Claude permission response, `crates/cli/src/sessions/mod.rs` for session ownership and exact permission binding, and `crates/core/src/observability/private_file.rs` for trace-path confinement and private writes. The live client scenarios are in `scripts/test-codex-plugin-e2e.sh` and `scripts/test-claude-plugin-e2e.sh`. The dynamic-plugin compatibility boundary is in `crates/cli/src/plugins/policy.rs` and `crates/cli/src/plugins/lifecycle/mod.rs`.

- Relates to: none

## Summary by CodeRabbit

* **Security**
  * Strengthened gateway and hook authentication with client ownership checks, browser-origin blocking, and secure client tokens.
  * Restricted gateway binding and dynamic-plugin runtime environments with secure defaults.
  * Protected local observability files with private permissions, safe paths, symlink rejection, and atomic writes.

* **New Features**
  * Added validated permission-request handling and session-scoped authorization for Claude and Codex.
  * Improved Claude plugin argument handling and transparent tool-use workflows.

* **Documentation**
  * Clarified response-cache tenant configuration and dynamic-plugin runtime behavior.


crates/core/src/observability/private_file.rs
```diff
@@ -0,0 +1,215 @@
...
+        }
+        Ok(metadata) if !metadata.is_dir() => {
+            return Err(io::Error::other(format!(
+                "observability output directory '{}' is not a directory",
+                path.display()
+            )));
+        }
+        Ok(_) => {}
+        Err(error) if error.kind() == io::ErrorKind::NotFound => {
+            fs::create_dir_all(path)?;
```
ericevans-nv: `create_private_dir_all` validates only the complete configured root. If an intermediate component is a symlink and the final root does not exist—for example `link/output`, where `link` points outside—`create_dir_all` follows it and the later lexical `strip_prefix` still succeeds. Please reject symlinks/reparse points in every configured-root component and add ATOF and ATIF regressions for this case.

mnajafian-nv: Addressed in `703893d6f`. Configured roots and nested output paths are now traversed component by component through no-follow directory handles, and file creation and atomic rename stay relative to the held parent handle. Added ATOF and ATIF regressions for a symlinked configured-root ancestor.

mnajafian-nv (approved):
Investigating CI now and will add a comment

mnajafian-nv (changes requested):
Request changes: Windows ATOF and ATIF local-file output is broken for normal absolute paths. The Windows amd64 Python, Node.js, and Go jobs reproduce the same failure.

crates/core/src/observability/private_file.rs
```diff
@@ -0,0 +1,144 @@
...
+fn open_or_create_private_dir(path: &Path) -> io::Result<ConfinedDir> {
+    let absolute = if path.is_absolute() {
+        path.to_path_buf()
+    } else {
+        std::env::current_dir()?.join(path)
+    };
+    let mut anchor = PathBuf::new();
+    for component in absolute.components() {
+        match component {
+            Component::Prefix(_) | Component::RootDir if anchor.as_os_str().is_empty() => {
```
mnajafian-nv: [P1] Windows absolute output paths are rejected before confined open. Windows yields `Prefix(C:)` followed by `RootDir`; after the prefix is added to `anchor`, the following root component takes the unsafe-traversal branch because this guard only permits it while `anchor` is empty. I confirmed this in the Windows amd64 Python, Node.js, and Go jobs: ordinary temporary directories all fail. Accept the root following a prefix as part of the trusted anchor, and add a Windows regression for an absolute temp path.

mnajafian-nv (approved):
Great work, LGTM

PR #938, opened 2026-08-28
fix: preserve ATIF metadata across interleaved LLM ends

### LINKED ISSUE #937: [Bug]: ATIF export drops metadata across interleaved LLM responses
### Affected area

Observability or exporters

### Current behavior

ATOF-to-ATIF conversion keeps one pending agent-step enrichment. When two LLM end events are interleaved, processing the second end replaces the first pending enrichment before it is committed. The earlier normalized agent step can therefore contain `extra: null`, losing its `llm_request`, `llm_response`, invocation metadata, and Responses API message phase even though the original ATOF response remains available.

### Expected behavior

Each normalized agent step should retain the request, response, ancestry, and invocation metadata derived from its corresponding LLM lifecycle event. Processing a later LLM end must not overwrite enrichment already pending for an earlier step.

### Steps to reproduce

1. Buffer two LLM lifecycle pairs belonging to different agent scopes.
2. Order the events as start A, start B, end A, agent end A, end B.
3. Export the buffered ATOF events to ATIF.
4. Observe that the first agent step has no `extra`, while the second response retains enrichment.

A focused Rust regression test reproduces this ordering without relying on wall-clock timing.

### Environment

- NeMo Relay commit: `ea471f0e`
- OS and architecture: macOS arm64
- Rust: repository-pinned 1.96.1 toolchain
- Python: 3.14.7, exercising the supported abi3 Python 3.11+ binding
- Node.js: repository test environment
- Go: 1.27.0
- Build and test commands: repository `just` recipes

### Impact

Concurrent root and subagent activity can produce incomplete ATIF trajectories. Downstream analysis loses response phase and correlation metadata for affected agent messages. The original ATOF response is retained, but the normalized ATIF association is lost. There is no reliable downstream workaround using ATIF alone.

owenniles:
#### Overview

Preserve pending ATIF agent-step enrichment when buffered ATOF events contain interleaved LLM end events. Previously, a later response could replace the pending request, response, ancestry, and invocation metadata for an earlier agent step, leaving that step with `extra: null`.

#### Details

- Finalize the current pending agent enrichment before processing the next LLM end event.
- Add a focused flat-stream regression test for the overwrite sequence.
- Add two minimized ATOF JSONL fixtures: a filtered flat regression and a synthetic agent-scoped control.
- Replace all recorded identifiers, timestamps, model names, payloads, paths, and metadata with deterministic synthetic values.
- Document each fixture's provenance, event sequence, expected behavior, and attribution limits.
- Verify response IDs, response-item IDs, final-answer phase, invocation IDs, and ancestry parent IDs; verify scoped responses remain isolated.
- Keep the change internal to ATIF normalization; there are no public API, binding, configuration, documentation, or breaking changes.

Validation:

- `cargo test -p nemo-relay test_exporter_preserves_extras_for_interleaved_llm_ends -- --nocapture`
- `just test-rust` — 4,441 workspace tests plus Rust example/plugin suites passed
- `cargo fmt --all`
- `uv run cargo clippy --workspace --all-targets -- -D warnings`
- `just test-python` — 80 native-binding, 707 package, and 24 plugin-example tests passed
- `just test-go`
- `just test-node` — 409 binding and 26 plugin-example tests passed

Start with `StepConversionState::handle_llm_end` in `crates/core/src/observability/atif.rs`, then review `test_exporter_preserves_extras_for_interleaved_llm_ends` in `crates/core/tests/unit/atif_tests.rs`, which demonstrates the failing event order and expected metadata retention.

- Fixes #937

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tracking of overlapping and nested agent interactions.
  * Preserved the correct request, response, reasoning, invocation, and ancestry details for each LLM operation.
  * Ensured continuation requests are associated with the correct agent step.
  * Prevented payloadless events from disrupting subsequent tool-result associations.

* **Tests**
  * Added coverage for interleaved and nested LLM responses to help prevent metadata mismatches.


crates/core/tests/fixtures/atif/synthetic_agent_scoped_interleaving.atof.jsonl
willkill07, line 1: Do not commit this file

crates/core/tests/fixtures/atif/README.md
willkill07, line 1: Do not commit this file

crates/core/tests/fixtures/atif/filtered_flat_interleaved_llm_ends.atof.jsonl
willkill07, line 1: do not commit this file

crates/core/tests/unit/atif_tests.rs
```diff
@@ -1889,6 +1889,193 @@ fn test_exporter_openai_responses_lifecycle_extracts_messages() {
...
+                .subagent_trajectories
+                .as_deref()
+                .unwrap_or_default()
+                .iter()
+                .find_map(|child| agent_step_by_response_id(child, response_id))
+        })
+}
+
+#[test]
+fn test_filtered_flat_interleaving_fixture_preserves_final_answer_extra() {
```
willkill07: I don't think a test relying on a materialized JSONL is reasonable. If it can be done programmatically, it should be.

crates/core/tests/unit/atif_tests.rs
```diff
@@ -1889,6 +1889,193 @@ fn test_exporter_openai_responses_lifecycle_extracts_messages() {
...
+    assert_eq!(response["output"][0]["id"], json!("msg_child"));
+    assert_eq!(response["output"][0]["phase"], json!("final_answer"));
+    assert_eq!(
+        extra.invocation.unwrap().invocation_id.as_deref(),
+        Some("00000000-0000-7000-8000-000000000211")
+    );
+}
+
+#[test]
+fn test_synthetic_agent_scoped_fixture_isolates_interleaved_responses() {
```
willkill07: I don't think a test relying on a materialized JSONL is reasonable. If it can be done programmatically, it should be.

willkill07 (changes requested):
Testing should avoid static files which can drift.

willkill07:
@owenniles there is also a larger issue where we need to update the ATIF exporter to handle simultaneous LLM Requests -- this PR doesn't address that.

We currently only permit one outstanding request at any time.

owenniles:
Thanks for your feedback @willkill07. I'll address all of those points as soon as I can.

Implementation aside, does this seem like a legitimate bug to you and one worth fixing?

crates/core/src/observability/atif.rs
```diff
@@ -2587,6 +2587,7 @@ impl StepConversionState {
 
     fn handle_llm_end(&mut self, event: &Event, lookups: &EventLookupMaps) {
         self.flush_observations();
+        self.finalize_agent_extra();
```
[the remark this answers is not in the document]
owenniles: Addressed in 5cdd8cc9. Pending continuation requests and reasoning effort are now stored by LLM UUID and consumed only by the matching end event. The regression uses distinct requests and reasoning values and verifies both agent steps retain the matching metadata.

crates/core/tests/fixtures/atif/filtered_flat_interleaved_llm_ends.atof.jsonl
```diff
@@ -0,0 +1,4 @@
...
+{"atof_version":"0.1","attributes":[],"category":"llm","category_profile":{"model_name":"test-model"},"data":{"id":"resp_child","output":[{"content":[{"text":"Child task complete.","type":"output_text"}],"id":"msg_child","phase":"final_answer","role":"assistant","type":"message"}]},"data_schema":null,"kind":"scope","metadata":{},"name":"openai.responses","parent_uuid":"00000000-0000-7000-8000-000000000210","scope_category":"end","timestamp":"2026-01-01T00:00:04.000000+00:00","uuid":"00000000-0000-7000-8000-000000000211"}
+{"atof_version":"0.1","attributes":[],"category":"llm","category_profile":{"model_name":"test-model"},"data":{"id":"resp_parent","output":[{"arguments":"{}","call_id":"call_wait","name":"wait_agent","type":"function_call"}]},"data_schema":null,"kind":"scope","metadata":{},"name":"openai.responses","parent_uuid":"00000000-0000-7000-8000-000000000110","scope_category":"end","timestamp":"2026-01-01T00:00:04.544000+00:00","uuid":"00000000-0000-7000-8000-000000000111"}
```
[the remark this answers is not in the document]
owenniles: Addressed in 5cdd8cc9 by removing both materialized JSONL fixtures. The scenarios are now constructed programmatically in the Rust unit tests, so fixture SPDX headers are no longer applicable.

crates/core/tests/fixtures/atif/README.md
```diff
@@ -0,0 +1,39 @@
+<!--
+SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# ATIF interleaving fixtures
+
+These minimized ATOF `0.1` fixtures model two LLM calls whose end events occur
```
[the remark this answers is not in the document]
owenniles: Addressed in 5cdd8cc9 by removing the fixture README together with the materialized fixtures; the coverage now lives directly in programmatic unit tests.

crates/core/tests/fixtures/atif/synthetic_agent_scoped_interleaving.atof.jsonl
owenniles: Removed in 5cdd8cc9. The agent-scoped scenario is now built programmatically in atif_tests.rs.

crates/core/tests/fixtures/atif/README.md
owenniles: Removed in 5cdd8cc9 along with the static fixtures.

crates/core/tests/fixtures/atif/filtered_flat_interleaved_llm_ends.atof.jsonl
owenniles: Removed in 5cdd8cc9. The flat interleaving scenario now uses Rust event builders.

crates/core/tests/unit/atif_tests.rs
owenniles: Addressed in 5cdd8cc9. This coverage is now programmatic and the static JSONL dependency has been removed.

crates/core/tests/unit/atif_tests.rs
owenniles: Addressed in 5cdd8cc9. The agent-scoped interleaving events are now constructed directly with the existing test builders.

owenniles:
@willkill07 Addressed in 5cdd8cc9. The exporter now tracks pending continuation requests and reasoning effort by LLM UUID, so simultaneous starts cannot overwrite one another before their matching ends. I also removed the materialized JSONL fixtures and rebuilt both flat and agent-scoped coverage programmatically. The focused regression reproduced the old cross-assignment (the first response received the second request) before the fix. Validation passed: 104 ATIF tests, just test-rust, cargo clippy, just test-python, just test-go, just test-node, and pre-commit --all-files.

willkill07 (changes requested):
This looks much better!

One more round:

The UUID-keyed pending request/reasoning state looks right. Could we add coverage for these interleavings?

 1. Matching `reasoning_effort` per LLM lifecycle:

    ```text
    LLM start A: reasoning_effort = "low"
    LLM start B: reasoning_effort = "high"
    LLM end A
    LLM end B
    ```

    Assert A’s agent step has `"low"` and B’s has `"high"`. A singleton “current reasoning effort” would incorrectly assign B’s value to A.

 2. Agent-scope isolation:

    ```text
    root agent starts
    subagent A starts → LLM start A
    subagent B starts → LLM start B
    LLM end A
    LLM end B
    ```

    Assert response A appears only in subagent A’s trajectory with A’s ancestry parent, and likewise for B. This ensures interleaved events do not mix metadata across scoped trajectories.

Also, please move `finalize_agent_extra()` and the UUID-map removals below the `event.data()` guard. A payload-less interleaved LLM end should not finalize the prior agent step, because its tool result may still arrive afterward.

willkill07:
@owenniles can you also cherry-pick these commits and have your PR branch target be `release/0.8` instead of `main` ?

owenniles:
I believe I've addressed all feedback from the current review cycle @willkill07!

willkill07:
/ok to test [485f744](https://github.com/NVIDIA/NeMo-Relay/pull/938/commits/485f744cc0ea8ef3976c6a13445f09277caa9d06)

willkill07:
> Implementation aside, does this seem like a legitimate bug to you and one worth fixing?

Yes, this is a legitimate bug

willkill07 (approved):

PR #947, opened 2026-08-28
feat: add gateway lifecycle controls

willkill07:
#### Overview

Add explicit gateway lifecycle commands and configurable health monitoring for persistent MCP gateways.

#### Details

- Add `nemo-relay gateway start` as the explicit gateway daemon entry point.
- Add `nemo-relay gateway stop`, which resolves the process listening at the exact configured loopback address and verifies that it is a `nemo-relay` executable before signaling it.
- Request graceful shutdown with `SIGINT` on Unix. `--force` terminates the Unix process tree, while Windows always terminates the process tree.
- Preserve the private token-authenticated shutdown path used by managed MCP sidecars.
- Add `NEMO_RELAY_PLUGIN_HEARTBEAT_INTERVAL_SECS`. The default is three seconds and is shortened when necessary to remain below the configured idle timeout; explicit values must be positive and shorter than that timeout.
- Forward the heartbeat setting through persistent MCP configuration and document the lifecycle behavior for Codex and Claude Code.
- Use `listeners` for cross-platform listener ownership resolution and `sysinfo` on Unix for process-tree cleanup. Regenerate the Rust attribution file for the resulting dependency changes.

Validation:

- `just test-rust`
- `just docs`
- Real gateway start, request and trace export, graceful stop, forced stop, foreign-listener refusal, and descendant-cleanup smoke coverage on macOS

Start with `crates/cli/src/commands/gateway.rs` for the process-selection and shutdown contract, then review `crates/cli/src/bootstrap/mod.rs` for heartbeat interval validation. The end-to-end lifecycle coverage is in `crates/cli/tests/cli_tests.rs`.

- Relates to: none


crates/cli/src/commands/gateway.rs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+/// Executes a gateway lifecycle command.
+pub(crate) async fn execute(
+    command: GatewayCommand,
+    server: &ServerArgs,
+    bootstrap_shutdown_token: Option<String>,
+) -> Result<ExitCode, CliError> {
+    match command.command {
+        GatewaySubcommand::Start => super::serve_gateway(server, bootstrap_shutdown_token).await,
```
mnajafian-nv, line 51: [P1] I think `gateway start` and `gateway stop` need one lifecycle contract. I started a gateway with `nemo-relay --bind 127.0.0.1:48765 gateway start`, then stopped that same endpoint. `gateway stop` exited successfully, but the process remained running. Start creates a regular daemon; stop only targets an owned MCP sidecar. Can we either make them operate on the same managed gateway or split the commands/docs so they are not presented as a start/stop pair? Please add a start-then-stop regression for the chosen contract.

willkill07: The purpose of stop is to find the process that is advertising the specific bound port for the configured gateway and stop/kill the process.

It should not be tied to any specific instance.

Think of it as a user friendly version of pkill or something similar on Windows

mnajafian-nv: Thanks, I’ll keep gateway start as the explicit replacement for the bare invocation and make gateway stop reliably stop the Relay gateway at the configured address. I’ll add an end-to-end start/stop test too.

mnajafian-nv (approved):

crates/cli/src/commands/mod.rs
```diff
@@ -60,6 +61,10 @@ struct LoggingSetup {
 
 fn configure_logging(cli: &Cli) -> Result<LoggingSetup, error::CliError> {
     let initialize = match cli.command.as_ref() {
+        Some(Command::Gateway(command)) if !command.is_stop() => {
```
ericevans-nv, line 64: Could you consider initializing logging for `gateway start` even when no configuration file or daemon-specific override is present? This would ensure explicit logging settings are honored. It would also be helpful to cover the configless default-bind path with an explicit logging setting.

ericevans-nv (approved):
The gateway lifecycle implementation and supporting configuration changes look sound overall. I left one optional nit: consider ensuring configless `gateway start` honors explicit logging settings.

ericevans-nv (approved):

crates/cli/src/commands/gateway.rs
```diff
@@ -0,0 +1,226 @@
...
+    select_relay_listener(bind, listeners)
+}
+
+pub(super) fn select_relay_listener(
+    bind: SocketAddr,
+    listeners: impl IntoIterator<Item = Listener>,
+) -> Result<Option<Process>, CliError> {
+    let mut matches = listeners
+        .into_iter()
+        .filter(|listener| listener.socket == bind && listener.protocol == Protocol::TCP);
```
mnajafian-nv, line 123: I think the dependency consolidation dropped an important safety check here. `listeners` 0.4 reports TCP sockets in all states, not only listeners. I reproduced `gateway stop --force` terminating a process named `nemo-relay` that had only an established client socket at the requested address. Can we require TCP `LISTEN` state before signaling and add a regression where an established-only socket is never selected?

mnajafian-nv (approved):
LGTM, This is a low-probability P3 edge case and i would not hold the PR on it: a custom gateway port would need to collide with the local port of another nemo-relay process’s outbound connection.

Salonijain27 (approved):

ericevans-nv (approved):

mnajafian-nv (approved):

PR #951, opened 2026-08-30
fix(cli): keep Codex routing in exec scope

### LINKED ISSUE #950: Codex transparent runs can bypass Relay when exec has command-scoped config
#### Problem

The Codex launcher inserts Relay provider and hook overrides immediately after the host executable. When the caller launches `codex exec` with additional `--config` values in the `exec` command scope, Codex resolves that nested scope independently and can fall back to the default OpenAI provider. The agent run succeeds, but its provider traffic bypasses Relay.

#### Reproduction

Prepare a transparent Codex launch whose original argv includes `codex exec --config <caller-setting>`. The prepared argv currently places `model_provider="nemo-relay-openai"` before `exec`, while the caller setting remains after it. The active `exec` scope consequently does not retain the Relay provider override. The same problem applies to `codex exec resume`.

#### Expected behavior

Relay-generated provider and hook overrides should be inserted into the active Codex command scope: after `exec`, or after `exec resume` for resumed sessions. Caller-supplied command-scoped configuration must not cause transparent provider traffic to bypass Relay.

#### Proposed validation

Add launcher regressions for both `exec --config ...` and `exec resume --config ...`, asserting that Relay overrides and caller configuration occupy the same active command scope.

ventali-fleet:
#### Overview

Keep NeMo Relay Codex provider and hook overrides in the active `exec` command scope so caller-supplied command configuration cannot cause transparent runs to bypass Relay.

- [x] I confirm that I have read the [Contributing Guidelines](https://github.com/NVIDIA/NeMo-Relay/blob/main/CONTRIBUTING.md) and followed the requirements.
- [x] I searched existing issues and pull requests before opening this contribution.

#### Details

Codex scopes `--config` overrides to the active clap subcommand. Relay previously inserted its generated provider override immediately after the host executable, while callers can append MCP configuration after `exec` or `exec resume`. In that shape, the inner execution can lose the Relay provider and use the normal provider instead.

This change inserts Relay generated configuration after `exec`, or after `exec resume`, while preserving the existing placement for other command shapes. It adds regressions for both forms with caller-supplied command configuration.

Validation:

- `cargo clippy -p nemo-relay-cli --lib -- -D warnings`
- focused Codex launcher tests: 6 passed
- pre-commit hooks for both changed files

No breaking changes.

#### Where should reviewer start?

Start with `insert_config_in_command_scope` in `crates/cli/src/agents/codex/launch.rs`, then the two new launcher tests.

#### Related Issues and/or Pull Requests

- Closes #950

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Codex launch handling for `exec`, `exec fork`, `exec resume`, and `exec review`.
  * Ensured configuration overrides are applied within the correct command scope, including nested commands.
  * Preserved proper ordering before MCP configuration, session arguments, and other command options.
  * Prevented unrelated commands containing `exec` arguments from receiving incorrectly scoped configuration.
  * Improved handling of commands using `--full-auto` and other preceding options.

### DIFF
--- a/crates/cli/src/agents/codex/launch.rs
+++ b/crates/cli/src/agents/codex/launch.rs
@@ -9,7 +9,7 @@ use crate::agents::CodingAgent;
 use crate::configuration::{RELAY_PLUGIN_ID, RELAY_SOURCE_PLUGIN_ID};
 use crate::error::CliError;
 use crate::hooks::{generated_policy_hooks, transparent_hook_forward_commands};
-use crate::process::{PreparedAgentLaunch, insert_after_host};
+use crate::process::PreparedAgentLaunch;
 
 pub(crate) fn prepare(launch: &mut PreparedAgentLaunch, gateway_url: &str) -> Result<(), CliError> {
     let has_openai_key = std::env::var("OPENAI_API_KEY")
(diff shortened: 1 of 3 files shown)


crates/cli/src/agents/codex/launch.rs
```diff
@@ -49,10 +49,34 @@ pub(crate) fn prepare(launch: &mut PreparedAgentLaunch, gateway_url: &str) -> Re
...
+    // Codex accepts configuration at each clap command level, but a nested command's
+    // overrides replace the parent scope. Keep Relay's provider and hook overrides in the
+    // same scope as caller-supplied `exec`/`exec resume` configuration.
+    let command = argv[host_index + 1..]
+        .iter()
+        .take_while(|arg| arg.as_str() != "--")
+        .position(|arg| arg == "exec")
+        .map(|index| host_index + 1 + index);
+    let insert_at = match command {
+        Some(exec_index) if argv.get(exec_index + 1).is_some_and(|arg| arg == "resume") => {
```
willkill07:
```
Run Codex non-interactively

Usage: codex exec [OPTIONS] [PROMPT]
       codex exec [OPTIONS] <COMMAND> [ARGS]

Commands:
  resume  Resume a previous session by id or pick the most recent with --last
  review  Run a code review against the current repository
  help    Print this message or the help of the given subcommand(s)
```

Wouldn't we need to handle any possible subcommand, not just resume? How do you expect to generalize this?

crates/cli/src/agents/codex/launch.rs
```diff
@@ -49,10 +49,34 @@ pub(crate) fn prepare(launch: &mut PreparedAgentLaunch, gateway_url: &str) -> Re
...
+    values: impl IntoIterator<Item = String>,
+) {
+    debug_assert!(host_index < argv.len());
+    // Codex accepts configuration at each clap command level, but a nested command's
+    // overrides replace the parent scope. Keep Relay's provider and hook overrides in the
+    // same scope as caller-supplied `exec`/`exec resume` configuration.
+    let command = argv[host_index + 1..]
+        .iter()
+        .take_while(|arg| arg.as_str() != "--")
+        .position(|arg| arg == "exec")
```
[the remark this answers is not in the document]
willkill07: This is valid feedback and should be addressed

willkill07 (changes requested):
Right overall approach, but implementation needs hardened. I'm also wondering if this logic should be extended to the equivalent for Claude Code.

crates/cli/src/agents/codex/launch.rs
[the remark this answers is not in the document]
ventali-fleet: Addressed in 86f4677d. The launcher now parses the first command position after consuming known option values instead of scanning arbitrary tokens. `codex mcp remove exec` therefore keeps Relay configuration immediately after the host, and the exact shape has a regression test. I also covered a top-level `--profile <value>` before `exec`.

crates/cli/src/agents/codex/launch.rs
ventali-fleet: Addressed in 86f4677d. The deepest model-running `exec` subcommands are now handled generically for both current forms: `resume` and `review`, including options before the subcommand. `help` does not run a model, so retaining the configuration in the parent `exec` scope is sufficient. A future model-running `exec` subcommand can be added to the explicit set with the same regression pattern.

ventali-fleet:
Rebased onto current `main` and pushed 86f4677d with both requested hardening changes. I also checked the Claude Code launcher: it constructs one merged `--settings` overlay rather than stacking clap-scoped `--config` values across nested commands, so this specific scope-loss condition does not apply there. Validation is green locally: six focused launcher tests, strict Clippy, formatting, and all changed-file pre-commit hooks.

crates/cli/src/agents/codex/launch.rs
willkill07: @ventali-fleet `fork` is now valid in newer CLIs as well.

ventali-fleet:
@willkill07 Addressed the remaining command-grammar cases in 68d7273d: `--full-auto` is now consumed as a pre-command flag and `exec fork` receives Relay configuration in the deepest model-running scope. Added exact regressions for both forms. Focused launcher tests pass (7/7), strict Clippy and formatting pass, and all changed-file pre-commit hooks pass.

willkill07:
/ok to test [aadf29c](https://github.com/NVIDIA/NeMo-Relay/pull/951/commits/aadf29c1c0e75b3b850a89500b22463e479e943f)

willkill07:
@ventali-fleet there is a CI failure that seems to correlate to your changes:
https://github.com/NVIDIA/NeMo-Relay/actions/runs/33511374067/job/99877296155?pr=951

willkill07:
/ok to test [763ff93](https://github.com/NVIDIA/NeMo-Relay/pull/951/commits/763ff93a0a24f29927f7fd8b42cd9d23ddcedb5f)

willkill07 (approved):

PR #953, opened 2026-08-31
fix: refresh managed integrations after Relay upgrade

willkill07:
#### Overview

Refresh Relay-managed Codex and Claude Code integrations after an upgrade, while preserving the existing safe stale-gateway recovery path.

#### Details

- Add `nemo-relay integrations refresh` to discover managed installations, validate all selected targets, retire their old MCP generations together, and force-replace each integration from the current Relay binary.
- Add MCP version-aware gateway handoff: when a new MCP process encounters an authenticated Relay-owned gateway from a different Relay version, it safely shuts it down so the host can start the new binary; foreign, invalid, and same-version endpoints are not replaced.
- Persist and serialize managed custom install directories for future refresh discovery.
- Keep registry-record failures non-fatal after a completed install or uninstall, with actionable warnings.
- Preserve authenticated replacement of a version-mismatched Relay-owned gateway and correct its stale-sidecar regression fixture.
- Document the refresh workflow for Codex, Claude Code, and plugin installation.

Start with `crates/cli/src/commands/integrations.rs`, `crates/cli/src/bootstrap/mod.rs`, and `crates/cli/src/installation/marketplace/mod.rs`: together they define the version-aware gateway handoff plus validation, generation-retirement, sidecar-refresh, and replacement lifecycle.

- Relates to: none

#### Validation

- Focused refresh-preflight regressions, `cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings`, and `just test-rust` passed.
- `just docs` passed with no documentation errors; its external redirect check emitted an FDR 403 warning.
- `uv run pre-commit run --all-files` passed.

## Summary by CodeRabbit

* **New Features**
  * Added `integrations refresh` for Relay-managed Codex and Claude Code integrations, with dry-run support and multiple installation directories.
  * Newly launched MCP clients can replace verified Relay-owned gateways running a different version.
  * Heartbeat intervals support environment-based configuration with validation and safe defaults.
* **Bug Fixes**
  * Foreign, invalid, or same-version gateways are no longer stopped or removed automatically.
  * Refresh operations validate all targets and preserve installations when prerequisites fail.
* **Documentation**
  * Updated gateway replacement, integration refresh, and compatibility guidance.

### DIFF
--- a/crates/cli/src/bootstrap/state.rs
+++ b/crates/cli/src/bootstrap/state.rs
@@ -225,9 +225,33 @@ pub(crate) fn stop_owned_and_reset(url: &str) -> Result<(), String> {
         return Ok(());
     }
     let _lock = lock_endpoint(&state, url)?;
-    let path = owner_path(&state, url);
+    stop_owned_and_reset_locked(&state, url).map(|_| ())
+}
+
+/// Stops a version-mismatched, Relay-owned gateway while the caller holds the
+/// endpoint startup lock.
+///
+/// Returns `true` when an owned gateway was stopped or a stale owned record was
+/// removed. Same-version, invalid, and absent records return `false` so callers
+/// can preserve the original incompatible-gateway error.
+pub(crate) fn stop_version_mismatched_owned_gateway_locked(
+    state: &Path,
+    url: &str,
+) -> Result<bool, String> {
+    let path = owner_path(state, url);
     let Some(owner) = read_owner_record(&path)? else {
-        return Ok(());
+        return Ok(false);
+    };
+    if !owner.valid_for(url) || owner.version == env!("CARGO_PKG_VERSION") {
+        return Ok(false);
+    }
+    stop_owned_gateway_locked(&path, &owner, url)
+}
+
+fn stop_owned_and_reset_locked(state: &Path, url: &str) -> Result<bool, String> {
+    let path = owner_path(state, url);
+    let Some(owner) = read_owner_record(&path)? else {
+        return Ok(false);
     };
     if !owner.valid_for(url) {
         return Err(format!(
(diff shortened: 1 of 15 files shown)


willkill07:
@coderabbitai full review

crates/cli/src/bootstrap/state.rs
```diff
@@ -225,20 +225,48 @@ pub(crate) fn stop_owned_and_reset(url: &str) -> Result<(), String> {
...
+) -> Result<bool, String> {
+    let path = owner_path(state, url);
     let Some(owner) = read_owner_record(&path)? else {
-        return Ok(());
+        return Ok(false);
+    };
+    if !owner.valid_for(url) || owner.version == env!("CARGO_PKG_VERSION") {
+        return Ok(false);
+    }
+    stop_owned_gateway_locked(&path, &owner, url)
```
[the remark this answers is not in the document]
willkill07: Verified against the current authenticated owner flow and regression fixture. Health compatibility checks bootstrap protocol, proof, and instance ID—not the package-version field—so a listener reporting `previous-version` is authenticated with its persisted owner fingerprint and receives shutdown. The fixture now models that prior version and asserts the shutdown request plus owner-record cleanup; no lifecycle change is required.

crates/cli/src/commands/integrations.rs
```diff
@@ -0,0 +1,146 @@
...
+    let targets = refresh_targets(command.install_dir.as_deref())?;
+    let managed_targets = targets
+        .iter()
+        .filter(|(agent, install_dir)| {
+            crate::installation::marketplace::persisted_state_exists(*agent, install_dir)
+        })
+        .cloned()
+        .collect::<Vec<_>>();
+    if !command.dry_run {
+        crate::installation::marketplace::prepare_integrations_for_refresh(&managed_targets)?;
```
mnajafian-nv: **[P1]** One recovery gap here: refresh retires every managed generation before replacement begins. If a later forced install fails, its rollback restores the retired marker, leaving the previous integration disabled. I reproduced this with a preflight retirement followed by a forced gateway-refresh failure; capturing the previous generation then returns `has been retired`. Could we restore the original marker for failed targets and add that regression?

mnajafian-nv (approved):
LGTM upon addressing inline suggestion

crates/cli/src/commands/integrations.rs
willkill07: Addressed in d22f1f60a2cf8eff09439cf320b66838aba73979

mnajafian-nv (approved):

PR #954, opened 2026-08-31
fix: avoid duplicate ATIF turns for response-less retries

afourniernv:
#### Overview

Prevent a failed, response-less LLM attempt from becoming an extra user turn in ATIF when a later response is provably a retry of the same logical request.

Targets `main` for Relay 0.9 development; no 0.8 backport is proposed.

#### Details

Before this change, ATIF projected every LLM start as a user step. An end event with `otel.status_code = "ERROR"` and no response produced no agent step, but a later successful retry projected the same request again. A single logical call could therefore look like multiple user turns.

The exporter now suppresses the failed ATIF span only when all of the following are true:

- the failed and successful spans each contain exactly one start and one end event;
- start and end identity agree within each span;
- both attempts have the same parent, operation, requested model, full unwrapped request, and exact non-empty `llm_correlation_request_id`;
- the failed end has exact `ERROR` status and no raw or annotated response;
- exactly one later, non-overlapping candidate has exact `OK` status and a non-null raw response.

Missing, conflicting, concurrent, response-bearing, or ambiguous evidence preserves the existing ATIF output. The requested model is compared rather than the provider-reported response model so a provider resolving an alias does not break valid retry correlation.

This is intentionally narrower than selecting an accepted response or hiding response-bearing attempts. ATOF remains unchanged and retains every physical failure and retry. There are no public API, binding, configuration, packaging, or documentation changes.

Live replay used six captured Hermes + Relay ATOF runs:

- all three baseline runs projected `3` steps: `1` user, `2` agent, `1` tool call;
- all three treatment runs projected `4` steps: `1` user, `3` agent, `2` tool calls;
- runs containing one to three response-less provider failures no longer gained duplicate user turns.

Validation:

- `cargo test -p nemo-relay --lib observability::atif::tests::` — 111 passed
- `cargo test -p nemo-relay --lib --quiet -- --test-threads=1` — 1,583 passed
- `uv run pre-commit run --files crates/core/src/observability/atif.rs crates/core/tests/unit/atif_tests.rs` — passed
- six real ATOF files replayed through the final exporter implementation with the expected step, source, and tool counts
- `git diff --check` — passed

The full core suite was run serially because an existing test temporarily changes the process working directory and can race with another test under parallel execution; that test also passed by itself.

Start with `suppress_response_less_retries` in `crates/core/src/observability/atif.rs`, then review `test_exporter_suppresses_response_less_retry_user_step` and the conservative negative cases in `crates/core/tests/unit/atif_tests.rs`.

- No tracking issue; discovered while qualifying the Hermes → Relay → ATIF → Gym flow.

### DIFF
--- a/crates/core/src/observability/atif.rs
+++ b/crates/core/src/observability/atif.rs
@@ -1727,7 +1727,162 @@ fn build_llm_dedupe(events: &[&Event]) -> LlmDedupeLookups {
         }
     }
 
+    suppress_response_less_retries(events, &mut lookups);
+
     lookups
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum LlmRetryTerminalState {
+    ResponseLessError,
+    SuccessfulResponse,
+    Other,
+}
+
+#[derive(Debug, Clone)]
+struct LlmRetryCandidate {
+    uuid: Uuid,
+    parent_uuid: Uuid,
+    start_ts: DateTime<Utc>,
+    end_ts: DateTime<Utc>,
+    operation: String,
+    model_name: String,
+    request: Json,
+    correlation_id: String,
+    terminal_state: LlmRetryTerminalState,
+}
+
+#[derive(Default)]
+struct LlmRetrySpanParts<'a> {
+    starts: Vec<&'a Event>,
+    ends: Vec<&'a Event>,
+}
+
+/// Avoid projecting a transport failure as a second user turn when a later
+/// response is provably a retry of the same request.
+///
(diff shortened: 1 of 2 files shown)


afourniernv:
Closing this implementation. The underlying bug is real: a failed model request followed by a successful retry can appear in ATIF as though the user sent the same message twice. This version fixes the sample by inferring retry groups from correlation metadata, but Relay does not guarantee that metadata identifies one logical call, and the approach can hide real attempts. We should revisit this with an explicit logical-call and attempt contract.

PR #965, opened 2026-09-01
fix(install): recover from orphaned plugin state in install, uninstall, and refresh

### LINKED ISSUE #963: Codex install fails with false-positive "missing generation marker" when local state file is stale
## Summary

`nemo-relay install codex --force` fails with a "missing generation marker" error even when there is no actual pre-existing install — because a stale local state file causes the installer to believe an install exists when neither Codex's own plugin registry nor the filesystem has any trace of one.

## Steps to reproduce

1. Have a `codex.json` at `~/Library/Application Support/nemo-relay/plugins/codex.json` left over from a prior install/uninstall cycle, where the recorded state (e.g. `hostMarketplaceRemoved: false`) is stale relative to reality (no marketplace/plugin actually present on disk, and `codex plugin marketplace list` shows nothing for `nemo-relay-local`).
2. Run `nemo-relay install codex --force`.

## Observed

```
installer error: cannot safely replace or uninstall an existing Codex plugin because its MCP generation marker is missing at
/Users/.../nemo-relay/plugins/codex-marketplace/plugins/nemo-relay-plugin/.nemo-relay-generation;
close all Codex clients and standalone `nemo-relay mcp` processes, run `codex plugin remove nemo-relay-plugin@nemo-relay-local` and
`codex plugin marketplace remove nemo-relay-local`, remove the stale marketplace and state from the selected install directory,
then run `nemo-relay install codex --force` to create a fenced install (and `nemo-relay uninstall codex` afterward if removal
was intended)
```

The suggested remediation (closing clients, removing marketplace/plugin via `codex plugin remove`) does not address the actual problem, since none of those things exist. The referenced generation-marker path's entire parent directory tree doesn't exist either.

## Root cause (as far as we could tell without deep tracing)

The installer's own local state file (`~/Library/Application Support/nemo-relay/plugins/codex.json`) can end up out of sync with actual on-disk/registered state (e.g. after a partial/failed prior uninstall, or manual cleanup of the marketplace directory without updating this state file). When that happens, `--force` install treats the phantom "existing install" as something it must safely replace, looks for a generation marker to prove safety, finds none (correctly, since nothing is there), and fails with an error that blames the wrong cause.

## Expected

If the tracked marketplace/plugin root doesn't exist on disk, the installer should either (a) treat this as "nothing to replace" and proceed with the force install, or (b) report the actual condition (state-file/disk desync) rather than pointing at running processes and generation markers.

## Environment

- `nemo-relay` 0.8.0 (built from `bbednarski/preserve-codex-openai-identity` @ `a10e7e4530bcfc30e00677dbe29ed0958eb5fbb6`, but this looks unrelated to that branch's changes)
- macOS 26.6.1
- Codex CLI 0.146.1

bbednarski9:
#### Overview

An orphaned plugin state file — one left behind by a partial/manual cleanup or an interrupted uninstall, which still parses while the install it describes is gone — was treated as proof of an existing install. That made `install`, `uninstall`, and `integrations refresh` all fail with a "missing generation marker" error whose suggested remediation (close clients, remove the marketplace/plugin via the host CLI) doesn't apply, because none of those things exist.

Fixes #963.

#### Details

All three commands now decide what to retire from what is actually on disk and registered with the host, using one shared predicate (`MarketplaceHost::local_install_exists`) instead of three subtly different checks.

**`install`** — `previous_install_exists` no longer counts `state_bytes.is_some()`.

Removing the state file from that flag would also have relaxed the non-forced path, which shares it to decide whether a plain `install` must refuse. That would let `install` silently overwrite the cleanup-retry state `force_uninstall_host_locked` writes back after a failed host-setup removal, which `installed_integrations` reads to keep offering cleanup. `previous_state_exists` is tracked separately so the require-force refusal still sees the state file while generation retirement does not.

**`uninstall`** — `uninstall_host_locked` derived `local_install_exists` from `state.is_some()`. This path matters because the install-side remediation text tells the user to run exactly that command afterward. `retire_installed_generation` still re-checks host registration when the fence is missing, so an install that lost its marketplace root but stayed registered is unaffected.

**`integrations refresh`** — `retire_integrations_for_refresh` demanded a generation fence for every target carrying a state file. Since refresh selects targets purely by "a state file exists", one orphaned file aborted the whole command for every host, including correctly installed ones, before any install could recover it. Such a target is now skipped and left to the forced install that follows; a target with artifacts still on disk keeps failing loudly, which is the genuine corruption case.

**Generation locks.** The free `local_install_exists` helper also counts `generation_lock.exists()`. That lock sits outside the marketplace root and survives a manual cleanup exactly like the state file, so it is no more evidence of an install — and a lone lock cannot retire anything without its fence. The host-aware predicate used here ignores it.

`prepare_plugin_install` in `crates/cli/src/installation/marketplace/mod.rs`, then the two sibling call sites (`uninstall_host_locked`, `retire_integrations_for_refresh`).

The design decision worth a second opinion is `force_install_recovers_from_a_bare_marketplace_root_for_claude_code`. Claude Code deliberately does not count a bare marketplace root as an install — it requires a plugin manifest, an `.mcp.json`, or a generation fence — so that test asserts a bare root plus a state file should recover. If the intent is that a bare root *should* be protected for Claude Code, the test and the condition both need to flip.

#### Test plan

- [x] `force_install_recovers_from_orphaned_state_with_no_matching_disk_or_host_install` — the original #963 scenario.
- [x] `force_install_recovers_from_a_bare_marketplace_root_for_claude_code` — pins the host-aware predicate; the Codex-only test above passes with or without it.
- [x] `install_without_force_still_refuses_to_overwrite_an_orphaned_state_file` — asserts both the error and that the state file is left byte-identical.
- [x] `uninstall_recovers_from_orphaned_state_with_no_matching_disk_or_host_install`
- [x] `refresh_preflight_tolerates_an_orphaned_state_file` and `refresh_preflight_retires_healthy_targets_alongside_an_orphaned_one` — the latter pins that one orphaned target does not stop healthy targets from being retired.
- [x] `cargo test -p nemo-relay-cli --lib installation::marketplace` — 140 passed, 0 failed, 0 ignored.
- [x] `cargo test -p nemo-relay-cli --lib` — 25 pre-existing failures (mostly `sessions::tests::*atif*`), byte-identical to the set on a clean baseline at this PR's original head; unrelated to this change.
- [x] `cargo fmt -p nemo-relay-cli -- --check`
- [x] `cargo clippy -p nemo-relay-cli --lib --tests -- -D warnings`

- Closes #963

### DIFF
--- a/crates/cli/src/installation/marketplace/mod.rs
+++ b/crates/cli/src/installation/marketplace/mod.rs
@@ -51,7 +51,7 @@ use state::{
 pub(super) use crate::bootstrap::DEFAULT_URL as DEFAULT_GATEWAY_URL;
 pub(super) const MARKETPLACE_NAME: &str = "nemo-relay-local";
 pub(super) const PLUGIN_NAME: &str = "nemo-relay-plugin";
-pub(crate) const RELAY_COMMAND: &str = "nemo-relay";
+pub(super) const RELAY_COMMAND: &str = "nemo-relay";
 
 fn default_operation_lock_dir() -> Result<PathBuf, String> {
     std::env::var_os("HOME")
@@ -605,9 +605,9 @@ fn prepare_install_transaction<H: MarketplaceHost>(
         None
     };
     if !options.force
-        && plugin_preflight
-            .as_ref()
-            .is_some_and(|preflight| preflight.previous_install_exists)
+        && plugin_preflight.as_ref().is_some_and(|preflight| {
+            preflight.previous_install_exists || preflight.previous_state_exists
+        })
     {
         return Err(existing_plugin_install_requires_force_error(host));
     }
@@ -1940,6 +1940,11 @@ struct PluginInstallPreflight {
     marketplace_registered: bool,
     previous_setup_installed: bool,
     previous_install_exists: bool,
+    // Tracked separately from `previous_install_exists`: an orphaned state file is not an install
+    // to retire, but it is still state a non-forced install must refuse to overwrite. Notably
+    // `force_uninstall_host_locked` writes one back as the cleanup-retry marker that
+    // `installed_integrations` relies on to keep offering cleanup.
+    previous_state_exists: bool,
     generation_retirement: Option<GenerationRetirement>,
 }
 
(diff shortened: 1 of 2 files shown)


willkill07 (dismissed):

willkill07 (approved):

willkill07 (approved):

mnajafian-nv (approved):
LGTM no p0 or p1

PR #967, opened 2026-09-01
fix(codex): preserve OpenAI session identity when routing through Relay

bbednarski9:
#### Overview

Preserve Codex conversation identity when installing Relay: keep the built-in `openai` provider, route it through the local gateway, and deliver gateway authentication through a persistent startup header instead of a credential-bearing URL.

#### Details

`nemo-relay install codex` configures a plain `openai_base_url` and appends an owned assignment to `$CODEX_HOME/.env`. Codex loads this at startup, including the tested desktop embedding. The namespaced `OpenAI-Project` value carries Relay's existing per-user proof and the caller's original project metadata. Relay validates and removes the proof before request preparation, observability, or forwarding. Normal project values and upstream Authorization credentials retain their existing semantics.

The installer preserves other dotenv settings, includes the dotenv file in install/trust rollback, supports existing dotenv symlinks, and removes only its managed block on uninstall. Config backup/restore and readiness detection cover the plain URL. Existing capability routes and dedicated Relay headers remain accepted, and the legacy `nemo-relay-openai` alias remains available for older sessions. The Responses WebSocket probe still receives HTTP 426 for immediate HTTP/SSE fallback.

Compatibility: existing gateway authentication mechanisms remain supported. The gateway's existing ChatGPT token classification and caller-credential routing policy are unchanged.

Validation at `792092fa`:

- Local formatting, file hygiene, documentation links, lockfile consistency, dependency policy, and generated attribution checks passed. CI check/clippy, all five Linux/macOS/Windows canonical Rust suites, docs preview, all seven package builds, and all seven package smoke jobs passed in [CI](https://github.com/NVIDIA/NeMo-Relay/actions/runs/35283721643).
- Added regressions for proof consumption, invalid/duplicate proofs, browser-origin rejection, original provider metadata, nested inherited carriers, dotenv preservation, permissions/symlinks, installation rollback, migration, and verified ownership before uninstall restoration.
- User-run Codex Desktop fixture received two successful header/auth requests with credentials loaded from persistent files. Its 1,044-entry diagnostic database contained neither synthetic credential.
- The exact-head CI-built macOS binary passed real install/uninstall in an isolated home. One Codex conversation resumed through caller-owned and gateway-held credentials, an app-server restart, and restored direct access after uninstall. Previous assistant content and thread identity were preserved. Relay stripped the proof and preserved project metadata upstream. WebSocket fallback took 34–91 ms in this run. A flushed 79-entry diagnostic database contained none of the synthetic proof/API keys.
- Live InferenceHub Nano, Super, and Ultra requests authenticated through the persistent header with `model_provider = "openai"`; fallback took 21–30 ms. Super and Ultra passed a shell-tool file-reading task. Nano answered incorrectly without a tool on its initial attempt, then passed on a bounded retry; model tool-following remains variable. Saved outputs contained no actual API key.
- Manual desktop validation passed: install using the PR binary, restart the app, and continue a conversation created before installation. Follow-up checks confirmed the built-in `openai` provider, plain gateway URL, managed persistent header, and exact-head PR binary serving port 47632. All installed Codex doctor checks passed with the installer's binary PATH. This confirms one desktop restart; a second restart was not separately reported.

Start with [environment.rs](https://github.com/NVIDIA/NeMo-Relay/blob/792092fa5dbcddd762f07d499aef7cfaf960c941/crates/cli/src/agents/codex/environment.rs) and [host.rs](https://github.com/NVIDIA/NeMo-Relay/blob/792092fa5dbcddd762f07d499aef7cfaf960c941/crates/cli/src/agents/codex/host.rs) for installation, persistence, and restoration. Then inspect [provider_auth.rs](https://github.com/NVIDIA/NeMo-Relay/blob/792092fa5dbcddd762f07d499aef7cfaf960c941/crates/cli/src/provider_auth.rs) and [server/mod.rs](https://github.com/NVIDIA/NeMo-Relay/blob/792092fa5dbcddd762f07d499aef7cfaf960c941/crates/cli/src/server/mod.rs) for authentication consumption before forwarding. The added coverage is under `crates/cli/tests/coverage/agents/` and `crates/cli/tests/coverage/shared/server_tests.rs`.

- Relates to #965
- Relates to #966

### DIFF
--- a/crates/cli/src/agents/codex/host.rs
+++ b/crates/cli/src/agents/codex/host.rs
@@ -799,6 +799,8 @@ pub(crate) fn install_codex_config(path: &Path, gateway_url: &str) -> Result<(),
     let gateway_url = super::versioned_gateway_url(gateway_url);
     let challenge = BootstrapChallengeKey::load().map_err(|error| error.to_string())?;
     let client_token = challenge.client_token();
+    let openai_base_url =
+        crate::configuration::persistent_openai_base_url(&gateway_url, &client_token);
     let raw = read_optional_text(path)?;
     let mut doc = raw
         .parse::<DocumentMut>()
@@ -827,7 +829,7 @@ pub(crate) fn install_codex_config(path: &Path, gateway_url: &str) -> Result<(),
             )),
         };
     }
-    doc["model_provider"] = value("nemo-relay-openai");
+    doc["openai_base_url"] = value(&openai_base_url);
     ensure_table(&mut doc, "features")["hooks"] = value(true);
     set_multi_agent_v2_enabled(&mut doc, false);
 
@@ -878,6 +880,7 @@ fn refresh_codex_config_backup(
     let mut baseline = current.clone();
     let preserved_provider = codex_extended_provider_without_proof(&baseline, gateway_url);
     let provider_is_managed = codex_provider_item_is_managed(&baseline, gateway_url);
+    restore_managed_openai_base_url(&mut baseline, previous, gateway_url, Some(challenge));
     restore_codex_config_from_backup(&mut baseline, previous, provider_is_managed, false);
     restore_codex_client_proof_from_backup(&mut baseline, previous, Some(challenge));
     if let Some(provider) = preserved_provider {
@@ -998,6 +1001,7 @@ fn sanitize_codex_backup_doc(
 
     if provider_is_managed {
         let preserved_provider = codex_extended_provider_without_proof(&backup, gateway_url);
+        remove_managed_openai_base_url(&mut backup, gateway_url, challenge);
         if top_level_item_is_str(&backup, "model_provider", "nemo-relay-openai") {
             backup.as_table_mut().remove("model_provider");
         }
@@ -1231,6 +1235,7 @@ pub(crate) fn uninstall_codex_config(
     let provider_is_managed = codex_provider_item_is_managed(&doc, &gateway_url);
     match backup_doc.as_ref() {
         Some(backup_doc) => {
+            restore_managed_openai_base_url(&mut doc, backup_doc, &gateway_url, challenge.as_ref());
             restore_codex_config_from_backup(
                 &mut doc,
                 backup_doc,
@@ -1238,7 +1243,13 @@ pub(crate) fn uninstall_codex_config(
                 preserve_hooks,
             );
         }
-        None => remove_codex_config_without_backup(&mut doc, provider_is_managed, preserve_hooks),
+        None => remove_codex_config_without_backup(
+            &mut doc,
+            &gateway_url,
+            provider_is_managed,
+            preserve_hooks,
+            challenge.as_ref(),
+        ),
     }
     if let Some(provider) = preserved_provider {
         ensure_table(&mut doc, "model_providers")
@@ -1302,10 +1313,66 @@ fn restore_codex_config_from_backup(
     restore_multi_agent_v2_enabled(doc, backup_doc);
 }
 
+fn codex_openai_base_url_is_managed(doc: &DocumentMut, gateway_url: &str) -> bool {
+    let Some(client_token) = codex_provider_client_token(doc) else {
+        return false;
+    };
+    let expected = crate::configuration::persistent_openai_base_url(gateway_url, client_token);
+    doc.get("openai_base_url")
+        .and_then(Item::as_value)
+        .and_then(TomlValue::as_str)
+        == Some(expected.as_str())
+}
+
+fn codex_openai_base_url_has_verified_capability(
+    doc: &DocumentMut,
+    gateway_url: &str,
+    challenge: Option<&BootstrapChallengeKey>,
+) -> bool {
+    let Some(challenge) = challenge else {
+        return false;
+    };
+    let expected_prefix = crate::configuration::persistent_openai_base_url(gateway_url, "");
+    doc.get("openai_base_url")
+        .and_then(Item::as_value)
+        .and_then(TomlValue::as_str)
+        .and_then(|url| url.strip_prefix(&expected_prefix))
+        .is_some_and(|token| {
+            !token.is_empty() && !token.contains('/') && challenge.verify_client_token(token)
+        })
+}
+
+fn restore_managed_openai_base_url(
+    doc: &mut DocumentMut,
+    backup: &DocumentMut,
+    gateway_url: &str,
+    challenge: Option<&BootstrapChallengeKey>,
+) {
+    if codex_openai_base_url_is_managed(doc, gateway_url)
+        || codex_openai_base_url_has_verified_capability(doc, gateway_url, challenge)
+    {
+        restore_top_level_item(doc, backup, "openai_base_url");
+    }
+}
+
+fn remove_managed_openai_base_url(
+    doc: &mut DocumentMut,
+    gateway_url: &str,
+    challenge: Option<&BootstrapChallengeKey>,
+) {
+    if codex_openai_base_url_is_managed(doc, gateway_url)
+        || codex_openai_base_url_has_verified_capability(doc, gateway_url, challenge)
+    {
+        doc.as_table_mut().remove("openai_base_url");
+    }
+}
+
 fn remove_codex_config_without_backup(
     doc: &mut DocumentMut,
+    gateway_url: &str,
     provider_is_managed: bool,
     preserve_hooks: bool,
+    challenge: Option<&BootstrapChallengeKey>,
 ) {
     if !provider_is_managed {
         return;
@@ -1313,6 +1380,7 @@ fn remove_codex_config_without_backup(
     if top_level_item_is_str(doc, "model_provider", "nemo-relay-openai") {
         doc.as_table_mut().remove("model_provider");
     }
+    remove_managed_openai_base_url(doc, gateway_url, challenge);
     if let Some(providers) = doc.get_mut("model_providers").and_then(Item::as_table_mut) {
         providers.remove("nemo-relay-openai");
     }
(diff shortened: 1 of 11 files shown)


docs/nemo-relay-cli/plugin-installation.mdx
```diff
@@ -101,7 +101,10 @@ it adds the provider route. The plugin also declares the native
 
 For Codex, `nemo-relay install codex` registers the local marketplace, installs
 `nemo-relay-plugin@nemo-relay-local`, enables Codex hooks, and configures the
-`nemo-relay-openai` provider alias at `http://127.0.0.1:47632/v1`. The plugin's
+Codex built-in `openai` provider to use the local Relay sidecar. The installed
+base URL contains a per-user capability that Relay validates and removes before
+forwarding a request. The installer retains a `nemo-relay-openai` provider alias
+only so sessions created by earlier Relay versions remain resumable. The plugin's
```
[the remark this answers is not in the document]
mnajafian-nv: +1. I reproduced the legacy upgrade path: an active `nemo-relay-openai` selection remains active after install, so new sessions still use the synthetic identity. Please migrate the active selection to `openai` while retaining the alias for old-session resume.

crates/cli/src/agents/codex/host.rs
```diff
@@ -1302,17 +1331,74 @@ fn restore_codex_config_from_backup(
...
+}
+
+fn restore_managed_openai_base_url(
+    doc: &mut DocumentMut,
+    backup: &DocumentMut,
+    gateway_url: &str,
+    challenge: Option<&BootstrapChallengeKey>,
+) {
+    if codex_openai_base_url_is_managed(doc, gateway_url)
+        || codex_openai_base_url_has_verified_capability(doc, gateway_url, challenge)
```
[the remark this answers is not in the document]
mnajafian-nv: Non-blocking +1. I reproduced this cleanup path with matching unverified header and base-URL values: uninstall removes the user URL. Requiring the verified capability for these mutations and adding the regression would make the ownership boundary safer.

crates/cli/src/agents/codex/host.rs
```diff
@@ -827,7 +829,15 @@ pub(crate) fn install_codex_config(path: &Path, gateway_url: &str) -> Result<(),
             )),
         };
     }
-    doc["model_provider"] = value("nemo-relay-openai");
+    doc["openai_base_url"] = value(&openai_base_url);
```
ericevans-nv, line 832: Setting `openai_base_url` on the built-in `openai` provider makes Codex automatically try WebSockets before HTTPS. Relay rejects this WebSocket path, so normal requests go through several failed connection attempts and retries before falling back. This adds avoidable delay and repeated error output to every request, with no setting here to disable the WebSocket attempt.

bbednarski9:
Confirming @ericevans-nv's concern with measurements, and adding a second problem with this approach that is independent of the transport issue.

## 1. The WebSocket cost, measured

Tested against Codex CLI `0.152.0` using the exact config shape `install_codex_config` writes (`openai_base_url` = capability URL, built-in `openai` provider active), pointed at a loopback listener:

```
7 x GET /v1/nemo-relay/<cap>/responses   Upgrade: websocket
backoff: 0.120 -> 0.290 -> 0.544 -> 0.941 -> 1.690 -> 3.202     sum ~6.8s
session start 20:11:00.288 -> HTTPS fallback 20:11:07.434        = 7.146s
```

Every session pays this before its first real request, with `ERROR codex_api::endpoint::responses_websocket` lines and a warning that attributes our missing transport to OpenAI: *"Falling back from WebSockets to HTTPS transport. We're currently experiencing high demand..."*. Against real upstream, add 7x RTT.

It cannot be turned off from config:

| Attempted | Result |
|---|---|
| `[model_providers.openai] supports_websockets = false` | `model_providers contains reserved built-in provider IDs: 'openai'. Built-in providers cannot be overridden.` |
| Same via `-c` on the command line | Identical rejection — the guard runs on the merged config |
| `[features] responses_websockets = false` (and `_v2`) | Loads without error, no effect |
| Answering the upgrade `404` instead of `500` | Still all 7 attempts; only the warning text changes |

Also worth noting: the handshake `GET` matches the wildcard route this PR adds in `server/mod.rs`, `post(gateway::passthrough).get(gateway::models)`. `gateway::models` hardcodes `ProviderRoute::OpenAiModels` regardless of path, so each of the 7 attempts becomes an **authenticated proxied GET to `api.openai.com/v1/responses`**. Same wildcard makes `GET /v1/nemo-relay/<cap>/anything` return the model list.

## 2. Second blocker: the capability token lands in logs

See openai/codex#37988 — Codex logs the full connection URL at `error!` level, at default verbosity, on every WebSocket fallback. Our capability URL carries the token in the path, so each session prints it to stderr seven times:

```
ERROR codex_api::endpoint::responses_websocket: failed to connect to websocket:
  ..., url: ws://127.0.0.1:PORT/v1/nemo-relay/<capability-token>/responses
```

That is CI logs, shared terminals, and pasted bug reports. It is inherent to putting the capability in the path, which this PR does because `openai_base_url` cannot carry a custom header. It stands independently of whether the latency issue is ever fixed.

## 3. Scope correction on what this PR buys

Worth restating, because it changes the trade. Pre-install conversations are **not lost** under the legacy provider. Verified against `0.152.0`:

| Attempt | Active provider | Result |
|---|---|---|
| `codex exec resume --last` | `nemo-relay-openai` | Misses it; starts a new thread |
| `codex exec resume <old-id>` | `nemo-relay-openai` | Works — same thread, history intact, continues through Relay |
| `codex exec resume --last` | `openai` | Picks the **older** `openai` thread over a **newer** `nemo-relay-openai` one |

Codex stores threads in `~/.codex/state_5.sqlite` with `model_provider` and an index on it; selection filters by the active provider. So the regression is **discoverability in the picker**, not preservation — pre-Relay threads stay resumable by id either way. (CLI verified directly; the desktop app is inferred to use the same query.)

A Codex user hit the same trade independently in openai/codex#27381: they disabled WebSockets with a custom provider and reported *"it messes with `codex resume` TUI — it doesn't display my previous sessions"*.

## 4. Upstream status

The blocker is already filed: openai/codex#27381 (feature request, proposes `supports_websockets = false` by name) and openai/codex#38638 (documents the reserved-provider rejection verbatim). No open PR addresses either.

I prototyped two fixes against `openai/codex` tag `rust-v0.152.0` and verified both with the probe above — with either applied, Codex keeps `provider: openai` **and** makes zero WebSocket attempts. Happy to share, but they depend on an upstream release, so they do not unblock 0.8.

## Suggestion

For 0.8 I think @ericevans-nv is right and we should defer: revert install to the legacy `nemo-relay-openai` provider, document that pre-install sessions resume by id, and keep the capability-URL gateway work for when the transport question is settled. The identity fix is worth doing, just not at the cost of a ~7s stall, misleading errors, and a token in stderr on every session.

Independent of direction, two things in this diff are worth fixing since the route ships either way: narrow `.get(gateway::models)` so it matches the models paths rather than every GET, and reject `Upgrade: websocket` locally instead of spending upstream credentials on 7 bogus GETs per session.

crates/cli/src/agents/codex/host.rs
```diff
@@ -838,7 +868,17 @@ pub(crate) fn install_codex_config(path: &Path, gateway_url: &str) -> Result<(),
...
     }
-    doc["model_provider"] = value("nemo-relay-openai");
+    super::environment::install(path, &client_token)?;
+    install_codex_tool_environment_exclusion(&mut doc)?;
+    doc["openai_base_url"] = value(&openai_base_url);
+    if doc
+        .get("model_provider")
+        .and_then(Item::as_value)
+        .and_then(TomlValue::as_str)
+        .is_some_and(|provider| provider != "openai" && provider != "nemo-relay-openai")
```
mnajafian-nv: I reproduced this with the existing legacy-install fixture: after installation, `model_provider` remains `nemo-relay-openai`. That means an upgraded user still uses the legacy provider alias, so their existing pre-Relay history is not discoverable alongside the built-in `openai` history this PR is intended to preserve. Can we migrate that selected provider to `openai` while keeping the alias available for explicitly resuming legacy sessions? Please add install and uninstall coverage for this migration.

crates/cli/src/agents/codex/host.rs
```diff
@@ -1313,17 +1402,194 @@ fn restore_codex_config_from_backup(
     restore_multi_agent_v2_enabled(doc, backup_doc);
 }
 
+const CODEX_PROJECT_ENVIRONMENT_VARIABLE: &str = "OPENAI_PROJECT";
+
+fn install_codex_tool_environment_exclusion(doc: &mut DocumentMut) -> Result<(), String> {
+    let policy = ensure_table(doc, "shell_environment_policy");
```
mnajafian-nv: I reproduced this with `shell_environment_policy = { inherit = "all", exclude = ["USER_SECRET"] }`. Installation replaces the inline table and drops both existing settings. That silently removes user environment-policy configuration, including exclusions, and uninstall cannot restore it. Can we preserve table-like TOML values here and add an install/uninstall round-trip test?

bbednarski9: same as CR comment, i addressed this in [6161ff0](https://github.com/NVIDIA/NeMo-Relay/pull/967/commits/6161ff0bdc02f793be2401d14e2005028377f42d)

crates/cli/src/agents/codex/host.rs
bbednarski9: Thanks for calling this out. Its relevant for migrating relay installs that use the old provider name. I addressed it in [ce23e9c](https://github.com/NVIDIA/NeMo-Relay/pull/967/commits/ce23e9c21753cbc60856d18ff28f10a4bf93a317)

mnajafian-nv (approved):
LGTM! Thanks

ericevans-nv (approved):

willkill07 (approved):
Approving dependencies + logic :)

bbednarski9:
Following up with daemon-mode PR that either patches it or docs that we dont support attribution mode changes

PR #971, opened 2026-09-01
fix: recover unhealthy managed MCP gateways

willkill07:
#### Overview

Recover managed MCP gateways that stop answering health checks, while making port conflicts actionable in operational logs.

#### Details

- Terminate an unhealthy managed gateway, wait one second, then force-kill its detached process tree before bootstrap starts a replacement.
- Preserve unmanaged-listener protection and add endpoint, health classification, remediation, and detailed failure context to bootstrap conflict logs.
- Add stale-owner and forced-termination regression coverage.
- Validated with `just test-rust`, focused bootstrap tests, commit hooks, and `cargo clippy -p nemo-relay-cli --all-targets -- -D warnings`.

Start with `crates/cli/src/bootstrap/mod.rs` for bootstrap classification and conflict events, then `crates/cli/src/bootstrap/state.rs` for process termination.

- Relates to: none

## Summary by CodeRabbit

* **Bug Fixes**
  * Unhealthy managed gateways are now stopped before replacement gateways start.
  * Stale ownership records for unreachable gateways are cleaned up safely.
  * Managed gateway cleanup verifies ownership before stopping processes, leaving unowned gateways unaffected.
  * Gateway processes and related child processes are fully terminated when necessary.
  * Gateway conflicts and recovery failures now provide clearer diagnostic details, including failure classification and remediation information.

### DIFF
--- a/crates/cli/src/bootstrap/mod.rs
+++ b/crates/cli/src/bootstrap/mod.rs
@@ -90,7 +90,8 @@ impl GatewaySpec {
                 log::error!(
                     target: "nemo_relay.bootstrap",
                     event = "gateway_acquisition_failed",
-                    bind = self.bind.to_string().as_str();
+                    bind = self.bind.to_string().as_str(),
+                    error = error.as_str();
                     "Gateway acquisition failed"
                 );
                 Err(error)
@@ -120,7 +121,8 @@ impl GatewaySpec {
                 log::error!(
                     target: "nemo_relay.bootstrap",
                     event = "gateway_recovery_failed",
-                    instance_id = expected_instance;
+                    instance_id = expected_instance,
+                    error = error.as_str();
                     "Gateway recovery failed"
                 );
                 Err(error)
@@ -200,8 +202,17 @@ fn acquire_gateway(spec: &GatewaySpec) -> Result<GatewayEndpoint, String> {
                 Err(incompatible_relay_error(&url))
             }
         }
-        (RelayHealth::Foreign, _) => Err(foreign_listener_error(&url)),
-        (RelayHealth::Unavailable, _) => start_gateway(spec, &state),
+        (RelayHealth::Foreign, _) => {
+            if state::stop_unhealthy_owned_gateway_locked(&state, &url)? {
+                start_gateway(spec, &state)
+            } else {
+                Err(foreign_listener_error(&url))
+            }
+        }
+        (RelayHealth::Unavailable, _) => {
+            state::stop_unhealthy_owned_gateway_locked(&state, &url)?;
+            start_gateway(spec, &state)
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


crates/cli/src/bootstrap/mod.rs
```diff
@@ -221,7 +232,9 @@ fn recover_gateway(spec: &GatewaySpec, expected_instance: &str) -> Result<Gatewa
             }
             (RelayHealth::Incompatible, _) => return Err(incompatible_relay_error(&requested_url)),
             (RelayHealth::Foreign, _) => return Err(foreign_listener_error(&requested_url)),
-            (RelayHealth::Unavailable, _) => {}
+            (RelayHealth::Unavailable, _) => {
```
ericevans-nv, line 239: A managed gateway that still accepts TCP but times out or returns a malformed health response is classified as `Foreign`, so recovery returns on the preceding branch and never reaches this cleanup. That leaves the unhealthy owned process holding the port and prevents the coordinated restart. Could the `Foreign` branch use the same identity-checked termination as acquisition, preserving the conflict only when no owned process was stopped?

crates/cli/src/bootstrap/state.rs
```diff
@@ -304,6 +334,143 @@ fn stop_owned_gateway_locked(path: &Path, owner: &OwnerRecord, url: &str) -> Res
...
+    let Ok(pid) = i32::try_from(pid) else {
+        return Ok(false);
+    };
+    if !process_is_running(pid) {
+        return Ok(false);
+    }
+
+    log_unhealthy_gateway_termination_started(pid);
+    signal_gateway_process_group(pid, libc::SIGTERM)?;
+    if wait_for_process_exit(pid, UNHEALTHY_GATEWAY_TERMINATION_TIMEOUT) {
```
mnajafian-nv: [P1] I tested the case where the gateway leader exits on `SIGTERM` but one child ignores it. The current wait sees the leader exit, skips `SIGKILL`, and leaves the child running. Because this path signals the process group, success should wait for the group to exit, with PID-only waiting reserved for the direct-PID fallback.

crates/cli/src/bootstrap/state.rs
```diff
@@ -304,6 +334,143 @@ fn stop_owned_gateway_locked(path: &Path, owner: &OwnerRecord, url: &str) -> Res
...
+    !process_is_running(pid)
+}
+
+#[cfg(windows)]
+fn terminate_owned_gateway_process(pid: u32) -> Result<bool, String> {
+    log_unhealthy_gateway_termination_started(pid as i32);
+    let graceful = std::process::Command::new("taskkill")
+        .args(["/PID", &pid.to_string(), "/T"])
+        .status();
+    if graceful.is_ok_and(|status| status.success()) {
```
bbednarski9: for windows can we make sure that termination exited first?

crates/cli/src/bootstrap/state.rs
```diff
@@ -304,6 +334,247 @@ fn stop_owned_gateway_locked(path: &Path, owner: &OwnerRecord, url: &str) -> Res
...
+}
+
+#[cfg(not(any(target_os = "linux", target_os = "macos", windows)))]
+fn process_identity(_pid: u32) -> Option<u64> {
+    None
+}
+
+fn owner_process_identity_matches(owner: &OwnerRecord) -> bool {
+    owner.process_identity.is_some_and(|identity| {
+        process_identity(owner.pid).is_some_and(|current| current == identity)
```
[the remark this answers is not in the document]
willkill07: I don't think this is feasibly addressable. I think this is okay to leave alone.

Reason: in the worst case we kill an innocent process.

mnajafian-nv (approved):
LGTM!Thanks

PR #978, opened 2026-09-02
fix: harden PII redaction safety

willkill07:
#### Overview

Harden built-in PII redaction against false positives, unsafe default removal, and silent sanitization omissions.

#### Details

- Require an explicit target selector for the destructive `remove` action.
- Avoid API-key matches embedded in ordinary words without bypassing valid AWS secret shapes.
- Emit structured warnings when annotated LLM payload or metric-envelope redaction omits a value after a failed round trip.
- Add focused regression coverage for the detector and configuration behavior.

Start with `crates/pii-redaction/src/builtin.rs`, which centralizes matcher filtering, omission diagnostics, and the unscoped-removal guard.

- Relates to: none

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved API key detection to avoid matching keys embedded within other words.
  * Git commit SHAs are no longer incorrectly redacted as AWS secret keys.
  * Invalid or failed payload and metric processing is now logged and safely omitted.
  * Improved sanitization failure reporting.

* **Validation**
  * Removal rules now require explicit target paths or globs, preventing unintended removal of all string values.
  * Updated configuration guidance clarifies how empty selectors behave.

### DIFF
--- a/crates/pii-redaction/src/builtin.rs
+++ b/crates/pii-redaction/src/builtin.rs
@@ -209,6 +209,27 @@ impl BuiltinMatcher {
         output.push_str(&text[cursor..]);
         output
     }
+
+    fn replace_all_expanded(&self, text: &str, replacement: &str) -> String {
+        let mut output = String::with_capacity(text.len());
+        let mut cursor = 0;
+        for captures in self.regex.captures_iter(text) {
+            let matched = captures
+                .get(0)
+                .expect("regular-expression captures always include the full match");
+            let matched_text = matched.as_str();
+            if self.detector == Some(BuiltinDetector::AwsSecretAccessKey)
+                && is_hex_git_sha(matched_text)
+            {
+                continue;
+            }
+            output.push_str(&text[cursor..matched.start()]);
+            captures.expand(replacement, &mut output);
+            cursor = matched.end();
+        }
+        output.push_str(&text[cursor..]);
+        output
+    }
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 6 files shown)


sylvesterkaczmarek (commented):
The Git-SHA exemption creates a redaction false negative: the AWS detector accepts any 40 characters in `[A-Za-z0-9/+=]`, so a syntactically valid secret composed entirely of hex digits also satisfies `is_hex_git_sha()` and is left unredacted. Could SHA preservation be scoped to fields known to carry commit IDs rather than globally bypassing the AWS secret detector?

bbednarski9:
Should we update configuration.mdx to include the new runtime requirement for action == "remove" to have >0 selectors?

crates/pii-redaction/src/builtin.rs
```diff
@@ -959,6 +1010,35 @@ fn log_llm_payload_omitted(direction: &str, codec: &LlmCodecIdentity, reason: &s
...
+fn sanitize_category_profile_with_backend(
+    backend: &CompiledBuiltinBackend,
+    profile: CategoryProfile,
+) -> Option<CategoryProfile> {
+    match sanitize_serializable_with_backend(backend, profile) {
+        Ok(profile) => Some(profile),
+        Err(_) => {
+            log::warn!(
+                target: "nemo_relay.plugin",
+                event = "pii_llm_payload_omitted",
```
bbednarski9: An omission here will show up in logs as an "annotated LLM payload" omission even when it isn't one, which will mislead anyone debugging redaction gaps. Suggest a distinct event name, e.g. pii_category_profile_omitted.

See lines 842 and 869 as well.

bbednarski9 (changes requested):

bbednarski9 (approved):

PR #979, opened 2026-09-02
fix: drain adaptive telemetry during shutdown

willkill07:
#### Overview

Preserve queued adaptive telemetry during graceful runtime shutdown with a bounded drain window.

#### Details

- Unregister telemetry producers and close the channel before shutdown waits for the drain task.
- Bound the graceful wait to five seconds, then log and abort only remaining work.
- Let synchronous deregistration detach a naturally closing drain instead of aborting it.
- Log receiver-loss drops and add shutdown-drain regression coverage.

Start with `crates/adaptive/src/runtime/features.rs`, specifically `AdaptiveRuntime::shutdown()` and `deregister()`.

- Relates to: none

## Summary by CodeRabbit

- **Bug Fixes**
  - Queued telemetry events now drain during shutdown and deregistration, improving recorded run reliability.
  - Shutdown and deregistration wait up to five seconds for telemetry processing before terminating stalled work.
  - Telemetry registrations are safely rolled back during shutdown.
  - Failed telemetry deliveries now generate a warning, making dropped events easier to identify.
  - Telemetry registration reports an error when delivery is unavailable.
  - Pending event counts are restored when telemetry delivery fails.

### DIFF
--- a/crates/adaptive/src/runtime/features.rs
+++ b/crates/adaptive/src/runtime/features.rs
@@ -526,10 +526,37 @@ impl AdaptiveRuntime {
             guard.clear();
         }
         // Remove every sender so the drain can process the queued events and
-        // finish naturally. Dropping the JoinHandle detaches that work because
-        // this synchronous API cannot await it.
+        // finish naturally. The synchronous API cannot await it, so supervise
+        // the detached drain with the same bounded timeout as shutdown.
         self.event_tx.take();
-        self.drain_handle.take();
+        if let Some(handle) = self.drain_handle.take() {
+            match tokio::runtime::Handle::try_current() {
+                Ok(runtime) => {
+                    runtime.spawn(async move {
+                        let mut handle = handle;
+                        if tokio::time::timeout(TELEMETRY_DRAIN_TIMEOUT, &mut handle)
+                            .await
+                            .is_err()
+                        {
+                            log::warn!(
+                                target: "nemo_relay.runtime",
+                                event = "adaptive_telemetry_drain_timeout";
+                                "Adaptive runtime deregistration timed out while draining telemetry; aborting remaining work"
+                            );
+                            handle.abort();
+                        }
+                    });
+                }
+                Err(_) => {
+                    log::warn!(
+                        target: "nemo_relay.runtime",
+                        event = "adaptive_telemetry_drain_aborted";
+                        "Adaptive runtime deregistration has no Tokio runtime; aborting telemetry drain"
+                    );
+                    handle.abort();
+                }
+            }
+        }
         self.registered = false;
         Ok(())
     }
--- a/crates/adaptive/src/runtime/features.rs
+++ b/crates/adaptive/src/runtime/features.rs
@@ -540,7 +540,8 @@ impl AdaptiveRuntime {
     /// A [`Result`] that is `Ok(())` when shutdown completes.
     ///
     /// # Errors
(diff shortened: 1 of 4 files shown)


crates/adaptive/src/runtime/features.rs
```diff
@@ -523,9 +525,11 @@ impl AdaptiveRuntime {
...
             guard.clear();
         }
-        if let Some(handle) = self.drain_handle.take() {
-            handle.abort();
-        }
+        // Remove every sender so the drain can process the queued events and
+        // finish naturally. Dropping the JoinHandle detaches that work because
+        // this synchronous API cannot await it.
+        self.event_tx.take();
+        self.drain_handle.take();
```
sara-tadayon-nv: What happens if the drain task hangs? Is that a concern in this `deregister()` path after the `JoinHandle` is dropped?

willkill07: Currently, it will hang 😅 -- addressing in a follow-up commit

crates/adaptive/src/runtime/features.rs
```diff
@@ -52,6 +52,8 @@ use crate::subscriber::create_subscriber_with_counter;
 use crate::tool_parallelism_learner::ToolParallelismLearner;
 use crate::types::cache::HotCache;
 
+const TELEMETRY_DRAIN_TIMEOUT: Duration = Duration::from_secs(5);
```
bbednarski9, line 55: Can we add a regression test to make sure we get the expected exception upon timeout? It's only 5 seconds, should be tolerable for our runners?

crates/adaptive/src/runtime/features.rs
```diff
@@ -523,8 +533,34 @@ impl AdaptiveRuntime {
...
             guard.clear();
         }
+        // Remove every sender so the drain can process the queued events and
+        // finish naturally. The synchronous API cannot await it, so supervise
+        // the detached drain with the same bounded timeout as shutdown.
+        self.event_tx.take();
         if let Some(handle) = self.drain_handle.take() {
-            handle.abort();
+            if let Some(runtime) = self.drain_runtime.take() {
+                runtime.spawn(async move {
```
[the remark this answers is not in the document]
bbednarski9: CR wants better documentation for this... I think its valid because if the runtime dies, the the handle isnt very much use. It's an edge case that could lead to the warning on 557 not actually firing. It might not be worth the solution - sync polling?

crates/adaptive/src/runtime/features.rs
```diff
@@ -523,8 +533,34 @@ impl AdaptiveRuntime {
...
+                        log::warn!(
+                            target: "nemo_relay.runtime",
+                            event = "adaptive_telemetry_drain_timeout";
+                            "Adaptive runtime deregistration timed out while draining telemetry; aborting remaining work"
+                        );
+                        handle.abort();
+                    }
+                });
+            } else {
+                log::warn!(
```
bbednarski9: depending on how you want to address the CR comment above this would be the dead code

bbednarski9 (changes requested):

crates/adaptive/src/runtime/features.rs
willkill07: done

crates/adaptive/src/runtime/features.rs
willkill07: done

bbednarski9 (approved):

PR #981, opened 2026-09-02
fix(cli): reuse MCP gateways across configuration changes

willkill07:
#### Overview

Reuse an existing compatible MCP gateway when the caller resolves a different persistent configuration.

#### Details

- Stop requiring the caller's bootstrap fingerprint to match the running gateway's persistent configuration for health checks and the Relay TLS tunnel.
- Continue binding the caller fingerprint into the authenticated bootstrap proof.
- Add coverage for a different client fingerprint and MCP heartbeat reuse.

Start with `crates/cli/src/server/mod.rs`, where the health and TLS bootstrap compatibility checks now prioritize MCP liveness.

- Relates to: none

## Summary by CodeRabbit

* **Bug Fixes**
  * Bootstrap tunnel requests now require valid client authentication while accepting valid caller fingerprints without requiring an exact match with the gateway’s configured fingerprint.
  * Health checks accept valid fingerprints when accompanied by the required nonce and challenge credentials, returning a successful status instead of an incompatibility error.
  * Gateway heartbeat behavior remains healthy when gateway and client fingerprints differ.
  * Verified gateway connections now include the required bootstrap authentication token, preventing unauthenticated tunnel payloads.

### DIFF
--- a/crates/cli/src/gateway/client.rs
+++ b/crates/cli/src/gateway/client.rs
@@ -126,7 +126,7 @@ pub(crate) fn post_verified(
     let authority = loopback_authority(&host, port);
     let client_token = key.client_token();
     let tunnel = format!(
-        "GET /bootstrap/tunnel HTTP/1.1\r\nHost: {authority}\r\nX-NeMo-Relay-Bootstrap-Fingerprint: {bootstrap_fingerprint}\r\nX-NeMo-Relay-Bootstrap-Nonce: {nonce}\r\nX-NeMo-Relay-Bootstrap-Token: {client_token}\r\nConnection: upgrade\r\nUpgrade: nemo-relay-tls\r\n\r\n"
+        "GET /bootstrap/tunnel HTTP/1.1\r\nHost: {authority}\r\nX-NeMo-Relay-Bootstrap-Fingerprint: {bootstrap_fingerprint}\r\nX-NeMo-Relay-Bootstrap-Nonce: {nonce}\r\n{BOOTSTRAP_CLIENT_TOKEN_HEADER}: {client_token}\r\nConnection: upgrade\r\nUpgrade: nemo-relay-tls\r\n\r\n"
     );
     stream.write_all(tunnel.as_bytes()).map_err(|error| {
         VerifiedHttpError::before_payload(format!(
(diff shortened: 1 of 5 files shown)


bbednarski9 (approved):

crates/cli/src/server/mod.rs
```diff
@@ -668,25 +669,25 @@ async fn bootstrap_tls_tunnel(
...
-        state.bootstrap_challenge_key.as_ref(),
-        state.bootstrap_tls.clone(),
-        state.local_address,
-    ) else {
+    let Some(key) = state.bootstrap_challenge_key.as_ref() else {
         return StatusCode::NOT_FOUND.into_response();
     };
-    if !fingerprint_matches
+    let token_is_valid = headers
+        .get(BOOTSTRAP_CLIENT_TOKEN_HEADER)
```
bbednarski9, line 676: Code now send X-NeMo-Relay-Bootstrap-Token but server validates x-nemo-relay-client-token per resolution of const

crates/cli/src/gateway/client.rs
```diff
@@ -124,8 +124,9 @@ pub(crate) fn post_verified(
         .map(|byte| format!("{byte:02x}"))
         .collect::<String>();
     let authority = loopback_authority(&host, port);
+    let client_token = key.client_token();
     let tunnel = format!(
-        "GET /bootstrap/tunnel HTTP/1.1\r\nHost: {authority}\r\nX-NeMo-Relay-Bootstrap-Fingerprint: {bootstrap_fingerprint}\r\nX-NeMo-Relay-Bootstrap-Nonce: {nonce}\r\nConnection: upgrade\r\nUpgrade: nemo-relay-tls\r\n\r\n"
+        "GET /bootstrap/tunnel HTTP/1.1\r\nHost: {authority}\r\nX-NeMo-Relay-Bootstrap-Fingerprint: {bootstrap_fingerprint}\r\nX-NeMo-Relay-Bootstrap-Nonce: {nonce}\r\nX-NeMo-Relay-Bootstrap-Token: {client_token}\r\nConnection: upgrade\r\nUpgrade: nemo-relay-tls\r\n\r\n"
```
bbednarski9, line 129: So i see how this PR makes the MCP fingerprint acceptance tolerate changes in the config. But now, whats the point of the fingerprint if we are just going to accept that it's different. Do we want to restart the gateway so that the fingerprint being accepted is indicative of the current config?

willkill07: The TL;DR is that there isn't much purpose right now.

I'd like to realign this with the daemon/broker mode.

bbednarski9 (approved):

PR #983, opened 2026-09-03
enhancement: add an experimental, reversible Codex thread-history migration

bbednarski9:
#### Overview

Codex records the provider that produced each thread in `threads.model_provider` and filters its resume picker by the provider that is currently active. Installing the Relay integration switches Codex to the `nemo-relay-openai` provider, so threads recorded under the built-in `openai` provider stop appearing in the picker. The threads stay on disk and remain resumable by id; only discovery is affected.

This adds an experimental, reversible migration behind `nemo-relay install codex --migrate-history` that records existing threads under the Relay provider so they stay visible. `nemo-relay uninstall codex` reverses it automatically, with no flag needed at uninstall.

This is a stopgap for the Relay side only. It does not change the upstream situation tracked in [openai/codex#27381](https://github.com/openai/codex/issues/27381), where the built-in `openai` provider cannot be pointed at a local gateway without also opting into a WebSocket transport Relay does not support.

#### Details

**New flags**

| Flag | Command | Behavior |
| --- | --- | --- |
| `--migrate-history` | `install codex` | Moves threads from `openai` to `nemo-relay-openai` after a successful install |
| `--history-database <PATH>` | `install codex`, `uninstall codex` | Names the thread database when Codex has moved past `state_5.sqlite` |
| `--skip-history-migration` | `uninstall codex` | Leaves thread history on the Relay provider |

`--dry-run` works with all of them and reports counts without touching the database or the journal.

**Reversibility.** A migration writes a journal (owner-only) to the user configuration directory recording the database path, backup path, timestamp, direction, and migrated thread ids. Uninstall reads that journal and reverses the migration without being asked, then clears it. `integrations refresh` does not re-run the migration; a recorded migration survives reinstalls and upgrades.

Reversal moves every thread still recorded under `nemo-relay-openai`, not only the ids the migration captured. Uninstall removes that provider from `config.toml`, and Codex resolves a thread's recorded provider against `config.toml` on resume, so a thread left pointing at a removed provider fails to resume with ``Model provider `nemo-relay-openai` not found`` — the same defect the migration exists to prevent, mirrored.

**Safety.** Before rewriting anything, Relay checkpoints the WAL and copies the database plus its sidecars to a timestamped `nemo-relay-history-backup-*` directory in the Codex home. `BEGIN IMMEDIATE` takes the write lock up front, so a running Codex produces a clean `database is locked` failure naming the remedy rather than partial state. Because the database path can come from a flag, the schema is validated before any copy or write: a database with no `threads` table or no `model_provider` column is rejected.

**No new dependency.** The migration shells out to `sqlite3` rather than adding a SQLite crate to the workspace. That makes `sqlite3` a runtime requirement for this one feature — present by default on macOS and most Linux distributions, but not on Windows or in minimal container images. This is documented in the Codex page, the installation page, and the flag's `--help`. No other Relay command depends on it.

**Explicit over discovery.** `state_5.sqlite` is pinned rather than discovered. When Codex bumps the schema generation, migration stops and names the path it expected instead of guessing at an untested schema; `--history-database` is the stopgap and the installer should be updated in a release.

**Failure semantics.** A migration failure is reported as a nonzero exit status with a structured log event, not as `CliError::Install`. Returning an install error made `run_agent_operations` report "failed to install one or more integrations" when the integration was installed and working, which sent the caller to the wrong remedy.

**Docs and skills.** New "Migrate Codex Thread History" section in the Codex page; cross-references from the host-neutral installation page's install and uninstall sections. The `nemo-relay-install` skill previously told agents never to touch Codex SQLite state as a migration workaround, because no supported path existed — that guidance and its generated recovery file are updated to point at the supported command while keeping the prohibition on hand-editing.

`crates/cli/src/agents/codex/history.rs` — the whole feature is one module, and its doc comment states the constraint that shapes every decision in it: Codex owns the schema and offers no supported API for changing a thread's provider, so a direct database edit is the only available mechanism.

The design decision most worth challenging is the reversal scope, documented on `restore_from_relay`: reversal deliberately covers threads created while Relay was installed, not just the migrated ones. `restore_also_moves_threads_created_while_relay_was_installed` pins that behavior.

Tests are in `crates/cli/tests/coverage/agents/codex_history_tests.rs` (17 cases: round trip, journal contents, backup contents, both dry runs, the relay-era reversal rule, database override in both bare-name and full-path forms, both schema rejections, and a locked-database case that holds a competing `BEGIN IMMEDIATE` from a second process). They drive a real `sqlite3` against a temporary Codex home and skip when `sqlite3` is absent.

**Validation.** `just test-rust` (4477 passed), `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo fmt --all`, `uv run pre-commit run --all-files` passing, `just docs` and `just docs-linkcheck` clean. Exercised end-to-end against a real 1361-thread Codex database in both directions: migration moved 1361, reversal moved 1376 (including 15 Relay-era threads), the journal was cleared, and `config.toml` was left with no reference to the Relay provider.

**Not yet verified.** Windows behavior — Rust CI covers `windows-2022` and `windows-11-arm`, but GitHub's Windows images do not ship `sqlite3`, so these tests will skip there and the run will still look green. Windows testing is in progress. Also unverified: whether a resumed migrated thread routes through Relay in a live session, and the stale `model_provider` left in rollout JSONL session headers, which appears cosmetic because resume reads the provider from the database.

- Relates to: [openai/codex#27381](https://github.com/openai/codex/issues/27381) — upstream request that would remove the need for this workaround
- Relates to: [openai/codex#24648](https://github.com/openai/codex/issues/24648) — the Codex Desktop provider-filter behavior this mitigates

### DIFF
--- a/crates/cli/src/agents/codex/history.rs
+++ b/crates/cli/src/agents/codex/history.rs
@@ -146,6 +146,7 @@ pub(crate) fn restore_from_relay(
     // An explicit override wins, then the database the migration pinned, then
     // the default. The journal matters most here: it names the database that
     // was actually rewritten, even if the default has since moved on.
+    let overridden = database.is_some();
     let database = match database {
         Some(database) => resolve_database(Some(database))?,
         None => match journal_database(&journal) {
@@ -154,6 +155,16 @@ pub(crate) fn restore_from_relay(
         },
     };
     if !database.exists() {
+        // Only a missing *recorded* database says the migration went stale. An
+        // override that does not resolve is a caller mistake, so keep the
+        // journal: discarding it would strand the threads on the Relay
+        // provider with no record of how to move them back.
+        if overridden {
+            return Err(format!(
+                "Codex thread database {} does not exist; kept the migration journal",
+                database.display()
+            ));
+        }
         clear_journal(dry_run)?;
         return Err(format!(
             "recorded Codex thread database {} no longer exists; discarded the migration journal",
--- a/crates/cli/src/agents/codex/history.rs
+++ b/crates/cli/src/agents/codex/history.rs
@@ -103,8 +103,18 @@ pub(crate) fn migrate_to_relay(
         return Ok(Some(outcome));
     }
     let backup = back_up_database(&database)?;
-    update_provider(&database, OPENAI_PROVIDER, RELAY_PROVIDER)?;
+    // Record the journal before the database changes. Writing it afterwards
+    // leaves no way back when the write fails: the threads have already moved,
+    // and uninstall infers reversal from the journal, so it would silently
+    // decline to reverse anything.
     write_journal(&database, &backup, &outcome)?;
+    if let Err(error) = update_provider(&database, OPENAI_PROVIDER, RELAY_PROVIDER) {
+        // The update is a single transaction, so a failure changed nothing and
+        // the journal describes a migration that never happened. Drop it, but
+        // do not let cleanup mask the original failure.
+        let _ = clear_journal(/*dry_run*/ false);
+        return Err(error);
+    }
     println!("moved {} in {}", outcome.describe(), database.display());
     println!(
         "backed up the previous thread database to {}",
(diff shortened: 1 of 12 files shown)


bbednarski9:
closing this out in leu of #1000

PR #984, opened 2026-09-03
fix: preserve LangChain message boundaries

ericevans-nv:
#### Overview

Preserve LangChain message boundaries when LLM requests pass through Relay. A multi-block LangChain message now remains one logical message instead of being expanded into multiple provider-facing messages.

#### Details

Previously, `LangChainCodec` flattened each content block into a separate annotated message. For Deep Agents, one four-block system message therefore reached the provider as four system messages, which could change how the composed system prompt was interpreted.

This change:

- Preserves one annotated message for each original LangChain message.
- Keeps ordered text and provider-native content blocks inside that message.
- Preserves message-level tool-call information.
- Allows middleware to modify individual blocks without mutating the original request.
- Adds synchronous and asynchronous LangChain and Deep Agents regression coverage.

Validation:

- `just test-python-langchain`: 103 tests passed.
- Real Deep Agents, Relay, and NVIDIA provider smoke: passed.
  - The provider received one four-block system message and one human message.
  - The managed tool emitted Relay start and end events.
  - The tool ran once and returned the expected result.
- `uv run pre-commit run --all-files`: passed.
- `just test-python`: 79 of 80 native tests passed. The remaining environment-sensitive Guardrails coverage test failed with `Invalid config path ./rails`; it does not exercise the changed LangChain serialization path.

This change does not alter public APIs or non-Python bindings.

Start with `python/nemo_relay/integrations/langchain/_serialization.py`, particularly the conversion between LangChain messages and Relay's annotated request representation.

- Relates to: none

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved LangChain message handling to preserve multi-block text, native content, system messages, and tool-call details.
  - Prevented edits to one system-message block from unintentionally affecting other messages or the original request.
  - Improved compatibility with Deep Agents middleware when processing messages with multiple content blocks.

- **Tests**
  - Added coverage for synchronous and asynchronous middleware flows, content preservation, tool calls, native image blocks, and targeted message edits.

### DIFF
--- a/python/nemo_relay/integrations/langchain/_serialization.py
+++ b/python/nemo_relay/integrations/langchain/_serialization.py
@@ -28,6 +28,7 @@ if TYPE_CHECKING:
 
 LANGCHAIN_MODEL_RESPONSE_KEY = "__nemo_relay_integrations_langchain_model_response"
 _LANGCHAIN_MODELED_REQUEST_KEYS = {"messages", "model", "tool_choice", "tools"}
+_LANGCHAIN_PROVIDER = "langchain"
 _LC_TO_RELAY_MESSAGE_ROLE = {
     "human": "user",
     "ai": "assistant",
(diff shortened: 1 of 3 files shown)


python/nemo_relay/integrations/langchain/_serialization.py
```diff
@@ -132,41 +133,79 @@ def _annotated_tool_calls_to_provider(cls, tool_calls: Any) -> list[dict[str, An
         ]
 
     @classmethod
-    def _langchain_message_to_annotated(cls, message: BaseMessage) -> list[dict[str, Any]]:
-        content = message.content
+    def _langchain_content_to_annotated(cls, content: Any) -> str | list[dict[str, Any]]:
```
bbednarski9, line 136: nit, non blocking: Bare string items inside a message's content list get irreversibly promoted to {"type": "text", "text": ...} dicts on decode, and the reverse conversion (_annotated_content_to_langchain) never converts them back

bbednarski9 (approved):

mnajafian-nv (approved):
LGTM+ coderabbit comment

python/nemo_relay/integrations/langchain/_serialization.py
ericevans-nv: Good catch, thanks. I updated the round trip to preserve bare-string items when the message content is unchanged, while intentionally using the normalized dictionary form when middleware modifies the content. I also added regression coverage for both cases.

ericevans-nv: Good catch, thanks. I updated the round trip to preserve bare-string items when the message content is unchanged, while intentionally using the normalized dictionary form when middleware modifies the content. I also added regression coverage for both cases.

PR #987, opened 2026-09-03
fix(cli): preserve Pi tool-result text

afourniernv:
#### Overview

Pi returns normal tool output as an ordered `content` array. The Relay extension only kept scalar strings, so Read and Bash spans were recorded as `Tool completed.` or `Tool failed.` instead of the text Pi produced.

This change keeps the text from those results while preserving the existing 2,000-character bound.

#### Details

- Join Pi text parts in order before applying the existing result limit.
- Omit image and unknown parts instead of serializing them into observability events.
- Keep the existing scalar-result and fallback behavior.
- Add Pi-shaped Read and Bash success/error tests, plus mixed-content and truncation coverage.
- Clarify the result projection in the Pi README and user guide.

There is no Rust, gateway, binding, event-schema, or public API change.

PR #986 refactors the same helper but does not add array handling. This branch is based directly on `main`, not on #986. It is staying in draft while that overlap is resolved; if #986 lands first, this branch will take the small conflict resolution before review.

Validation:

- `just test-pi` — 113 passed.
- `cargo build -p nemo-relay-cli` — passed.
- Changed-file pre-commit checks — passed.
- Real Pi 0.84.4 through the rebuilt Relay CLI and a deterministic local OpenAI-compatible endpoint — passed. Pi executed Read, and the ATOF tool-end event contained the exact canary text instead of `Tool completed.`

```json
{"name":"read","status":"ok","content":"CANARY-PI-TOOL-RESULT: Ada Lovelace has an account at ada.lovelace@example.test.\nThis line exists to verify the real Pi AgentToolResult shape reaches Relay rather than a generic completion marker.\n"}
```

Start with `toolResultText()` and `summarize()` in `crates/cli/assets/pi-extension/index.ts`, then the real result shapes in `integrations/pi/test/tool-result.test.mjs`.

- Relates to: none

## Summary by CodeRabbit

* **Bug Fixes**
  * Structured tool results now preserve text blocks in order while omitting images and other non-text content.
  * Large results are truncated to 2,000 characters with an indication of omitted content.
  * Truncation preserves Unicode characters correctly.
  * Results without text use a completion-status fallback, including clearer failed-command diagnostics.

* **Documentation**
  * Updated guidance to explain structured-result formatting, omitted content, and truncation behavior.

### DIFF
--- a/crates/cli/assets/pi-extension/index.ts
+++ b/crates/cli/assets/pi-extension/index.ts
@@ -729,15 +729,40 @@ function summarize(result: unknown, isError: boolean): unknown {
   if (typeof result === 'object') {
     const record = result as Record<string, unknown>;
     const content = record.content ?? record.output ?? record.text;
+    const text = toolResultText(content);
     return {
-      content:
-        typeof content === 'string'
-          ? truncate(content)
-          : `Tool ${isError ? 'failed' : 'completed'}.`,
+      content: text === null ? `Tool ${isError ? 'failed' : 'completed'}.` : text,
       result_keys: Object.keys(record).slice(0, 20),
     };
   }
   return { content: truncate(String(result)) };
+}
+
+/** Extract and bound Pi's ordered text blocks without forwarding image or unknown parts. */
+function toolResultText(content: unknown): string | null {
+  if (typeof content === 'string') return truncate(content);
+  if (!Array.isArray(content)) return null;
+
+  let text = '';
+  let omittedChars = 0;
+  let foundText = false;
+
+  const append = (value: string): void => {
+    const kept = value.slice(0, Math.max(0, MAX_CONTENT_CHARS - text.length));
+    text += kept;
+    omittedChars += value.length - kept.length;
+  };
+
+  for (const part of content) {
+    if (typeof part !== 'object' || part === null || Array.isArray(part)) continue;
+    const block = part as Record<string, unknown>;
+    if (block.type !== 'text' || typeof block.text !== 'string') continue;
+    if (foundText) append('\n');
+    append(block.text);
+    foundText = true;
+  }
+  if (!foundText) return null;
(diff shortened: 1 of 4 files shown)


willkill07 (approved):

willkill07:
/ok to test [7ac414b](https://github.com/NVIDIA/NeMo-Relay/pull/987/commits/7ac414b88d3e54370dee5722ba2768d714af4ffe)

crates/cli/assets/pi-extension/index.ts
```diff
@@ -729,17 +729,42 @@ function summarize(result: unknown, isError: boolean): unknown {
...
+function toolResultText(content: unknown): string | null {
+  if (typeof content === 'string') return truncate(content);
+  if (!Array.isArray(content)) return null;
+
+  let text = '';
+  let omittedChars = 0;
+  let foundText = false;
+
+  const append = (value: string): void => {
+    const kept = value.slice(0, Math.max(0, MAX_CONTENT_CHARS - text.length));
```
yczhang-nv: `value.slice()` cuts on UTF-16 code units, so a cut inside a surrogate pair emits a lone surrogate and the gateway rejects the whole POST. I'll suggest to execute `toWellFormed()` on the result, or trim a trailing unpaired high surrogate.

yczhang-nv (approved):
Approved with one comment

crates/cli/assets/pi-extension/index.ts
afourniernv: Good catch—confirmed. The 2,000-unit slice produced a trailing high surrogate, which could make Relay reject the tool-result hook as invalid JSON. Fixed in b7c0e123 by cutting only at a UTF-16 code-point boundary in both the structured and scalar truncation paths. The regression covers 1,999 ASCII characters followed by an emoji through both paths; the typecheck and all 112 Pi integration tests pass.

willkill07 (approved):

PR #991, opened 2026-09-04
fix: harden trajectory context redaction

bbednarski9:
#### Overview

Make the `trajectory_context` PII-redaction preset contract-first: LLM observability payloads are generated solely from sanitized normalized annotations, rather than by overlaying sanitized fields onto original provider JSON. This fails closed when a provider surface cannot be safely decoded or rendered.

#### Details

- Add a private trajectory projection module that accepts only sanitized `AnnotatedLlmRequest` and `AnnotatedLlmResponse` values; it cannot inspect the original provider payload.
- Render minimal, decodeable request and response shapes for OpenAI Chat, OpenAI Responses, Anthropic Messages, OCI GenAI, and Gemini `generateContent`.
- Keep only approved normalized analytics and safe identifiers, including model and tool names, sanitized content, request tuning, usage/cache accounting, and validated cost values.
- Omit source-only provider data by default: unknown fields and extensions, headers, raw metadata, opaque API-specific values, logprobs, reasoning, tool arguments, timestamps, extra choices/candidates, and additional output items. Responses emit one normalized representation only.
- Return `{}` for unsupported, runtime, opaque, malformed, undecodable, or unrenderable request/response payloads.
- Rebuild LLM event data from the sanitized annotation so the same boundary applies to subscribers, ATIF, OpenTelemetry, GenAI OpenTelemetry, OpenInference, and downstream backends. Real provider request and response values are unchanged.
- Retain existing overlay behavior only for non-`trajectory_context` normalized-path policies.
- Update the PII-redaction README and configuration guide to describe generated minimal projections, typed metric retention, and fail-closed behavior.
- Add provider projection matrices, sentinel-leakage coverage, failure-path coverage, multi-choice/candidate coverage, and a live Phoenix/inferencehub end-to-end harness.

Validation:

- `cargo test -p nemo-relay-pii-redaction` (168 passed)
- `just test-rust` (4,628 passed, plus plugin example suites)
- `just docs` (passed; Fern reported one unauthenticated redirect-check warning)
- `uv run pre-commit run --all-files` was run; it remains blocked only by pre-existing Python test `ty` unused-ignore diagnostics outside this change.
- Live end-to-end checks delivered redacted OpenAI Chat, OpenAI Responses, Anthropic Messages, and Gemini `generateContent` projections to Phoenix while preserving the caller-facing provider values. The available inferencehub gateway did not expose an OCI GenAI model or route, so OCI live verification is not included.

Start with `crates/pii-redaction/src/trajectory_projection.rs`, which makes the source-payload boundary structural. Then review `crates/pii-redaction/src/trajectory.rs` and `crates/pii-redaction/src/builtin.rs` for the sanitizer and event-pipeline integration, followed by the provider and exporter coverage in `crates/pii-redaction/tests/unit/component_tests.rs`.

- Relates to: none

## Summary by CodeRabbit

* **New Features**
  * Added fail-closed sanitization for trajectory data across supported LLM providers.
  * Preserves approved typed fields, including identifiers, usage, metrics, severity, and provider metadata, while removing unsupported or opaque content.
  * Added sanitized request and response projections for OpenAI, Anthropic, OCI GenAI, and Gemini.
  * Added configurable custom-mark handling and exact metric string-array allowlists; unapproved, numeric, and boolean attributes are dropped.

* **Documentation**
  * Documented the `trajectory_context` preset, sanitization behavior, supported fields, and configuration options.


crates/pii-redaction/src/trajectory.rs
```diff
@@ -449,265 +866,101 @@ fn redact_custom_category_profile(
...
+                .metadata()
+                .and_then(Json::as_object)
+                .and_then(|value| find_string(value, &keys))
+        })
+        .or_else(|| {
+            event
+                .data()
+                .and_then(Json::as_object)
+                .and_then(|value| find_string(value, &keys))
+        })
```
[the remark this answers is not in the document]
bbednarski9: 951c762c

scripts/test-trajectory-context-e2e.py
```diff
@@ -0,0 +1,398 @@
...
+    except HTTPError as error:
+        if error.code in {301, 302, 307, 308} and redirects_remaining:
+            location = error.headers.get("Location")
+            if location:
+                return post_json(
+                    urljoin(url, location),
+                    headers,
+                    payload,
+                    redirects_remaining=redirects_remaining - 1,
+                )
```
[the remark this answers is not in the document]
mnajafian-nv: +1. I reproduced this locally with a synthetic key: a 307 redirect to a different origin received the original Authorization header. Please restrict or disable redirects before forwarding provider credentials.

mnajafian-nv (approved):
LGTM, Conditional approval upon addressing the redirect handling. CodeRabbit has a few additional nits worth considering if you have time.

crates/pii-redaction/src/trajectory_projection.rs
```diff
@@ -0,0 +1,605 @@
...
+/// trajectory contract does not model.
+pub(super) fn render_request(
+    surface: ProviderSurface,
+    request: &AnnotatedLlmRequest,
+) -> Option<LlmRequest> {
+    let template = LlmRequest {
+        headers: Map::new(),
+        content: request_template(surface),
+    };
+    let mut rendered = request_codec(surface).encode(request, &template).ok()?;
```
willkill07: OCI request projections can become empty or change to the wrong format.

Preconditions:
1. The `trajectory_context` preset is enabled.
2. The active codec is `oci_genai`.
3. The request is a normal ChatDetails envelope with a model in `servingMode`, or it uses COHERE or COHEREV2.
4. The sanitized annotation is rendered from this fresh GENERIC template.

The decoder keeps the envelope model and OCI API format in the annotation. The trajectory sanitizer removes `api_specific`, and this renderer uses a bare GENERIC template. For an enveloped request, the OCI encoder sees the retained model as an unsupported change and returns an error, so the caller emits `{}`. A COHERE or COHEREV2 request can also be rendered as GENERIC instead of its original request shape. The current test uses only a bare GENERIC body, so it does not cover these cases.

Possible fixes:
- Build a sanitized template for each OCI format and keep the minimum safe envelope fields needed by the encoder.
- Add a dedicated OCI trajectory renderer that handles GENERIC, COHERE, and COHEREV2 without copying unknown fields.
- If these formats are intentionally unsupported, reject or document them clearly and add tests for the empty result.

crates/pii-redaction/src/builtin.rs
```diff
@@ -889,7 +935,21 @@ pub(super) fn llm_sanitize_request_callback(
                     .as_object()
                     .cloned()
                     .unwrap_or_default();
-                request.content = trajectory.sanitize_provider_payload(request.content);
+                let sanitized = backend
+                    .selected_surface(context.codec())
```
willkill07: Active runtime and opaque request codecs are skipped.

Preconditions:
1. The `trajectory_context` preset is enabled.
2. A managed call provides a resolved request codec.
3. Its identity is `Runtime` or `Opaque`.
4. The callback reaches `selected_surface(context.codec())`.

`selected_surface` returns `None` for both identities. This path never calls `context.resolve_codec()`, even though the normal request-sanitizing path below does. The request therefore falls through to an empty `{}` payload even when Relay has a codec that can decode it.

Possible fixes:
- Use the resolved codec to decode the request and add a safe projection contract for runtime codecs.
- Let runtime codecs declare a supported built-in projection surface.
- If custom codecs cannot be projected safely, document that limit and test the fail-closed behavior explicitly.

crates/pii-redaction/src/builtin.rs
```diff
@@ -638,6 +650,26 @@ impl CompiledBuiltinBackend {
...
+        request: &LlmRequest,
+    ) -> Option<LlmRequest> {
+        let provider_hint = match surface {
+            ProviderSurface::AnthropicMessages => Some("anthropic.messages"),
+            ProviderSurface::OCIGenAI => Some("oci.genai"),
+            ProviderSurface::OpenAIChat
+            | ProviderSurface::OpenAIResponses
+            | ProviderSurface::GeminiGenerateContent => None,
+        };
+        (detect_request_surface_with_hint(&request.content, provider_hint) == Some(surface))
```
willkill07: This surface check may reject requests that the active codec can handle.

Preconditions:
1. Relay has already selected an active request codec.
2. The request shape is ambiguous, or the codec identity is `Runtime` or `Opaque`.
3. This function tries to detect the surface again from the raw payload.

At this point, the active codec is a stronger source of truth than payload heuristics. The extra detection step needs provider hints only because it is trying to rediscover information Relay already has. It also cannot represent runtime or opaque codecs.

Possible fixes:
- Trust the active built-in surface and let that codec decode the request. A decode error can still fail closed.
- Pass the resolved codec into this function, use it for decoding, and add a safe projection contract for runtime or opaque codecs.
- If runtime projections are not supported, separate that case explicitly instead of treating it as a surface-detection failure.

willkill07 (changes requested):
Some observations with Codec usage and application.

crates/pii-redaction/src/component.rs
```diff
@@ -1292,7 +1293,7 @@ fn register_builtin_backend(
             sanitizer,
         )?;
     }
-    if config.input {
+    if config.input && !compiled.is_trajectory() {
```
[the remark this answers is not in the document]
bbednarski9: [87938ce](https://github.com/NVIDIA/NeMo-Relay/pull/991/commits/87938ce45fbed53cd5de85f4fca443a30d79b135)

scripts/test-trajectory-context-e2e.py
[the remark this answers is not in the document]
bbednarski9: removed this entire example

crates/pii-redaction/src/builtin.rs
bbednarski9: already addressed in related refactor. line 949 now documents fail closed.

crates/pii-redaction/src/builtin.rs
bbednarski9: Thanks i got this one in 5dcdc5be. Recognized built-in codec identity is now authoritative

crates/pii-redaction/src/builtin.rs
```diff
@@ -922,16 +922,40 @@ pub(super) fn llm_sanitize_request_callback(
...
+pub(super) fn trajectory_llm_request_projection_callback(
+    backend: CompiledBuiltinBackend,
+) -> LlmSanitizeRequestFn {
+    let backend = Arc::new(backend);
+    Arc::new(move |request: LlmRequest, context| {
+        let backend = Arc::clone(&backend);
+        Box::pin(async move {
+            if matches!(context.codec(), LlmCodecIdentity::None) {
+                return Ok(backend.legacy_surface.is_some().then_some(request));
+            }
```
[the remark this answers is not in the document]
bbednarski9: NA

crates/pii-redaction/src/trajectory_projection.rs
```diff
@@ -0,0 +1,658 @@
...
+    Some(rendered)
+}
+
+fn request_template(surface: ProviderSurface, request: &AnnotatedLlmRequest) -> Json {
+    match surface {
+        ProviderSurface::OpenAIChat | ProviderSurface::AnthropicMessages => {
+            json!({"messages": []})
+        }
+        ProviderSurface::OpenAIResponses => json!({"input": []}),
+        ProviderSurface::OCIGenAI => {
```
mnajafian-nv: [P1] I reproduced this with standard OCI `ChatDetails` requests for both GENERIC and COHERE. Because the projection starts from a bare `chatRequest`, the OCI codec sees no `servingMode` model in its template and rejects the sanitized annotation. `render_request()` then returns `None`, so the event emits `{}`. Could we build a minimal sanitized `ChatDetails` envelope here and add coverage for enveloped OCI requests?

bbednarski9:
@coderabbitai review

crates/pii-redaction/src/trajectory_projection.rs
bbednarski9: 4b421de5 fix: project OCI trajectory request envelopes

willkill07 (approved):

mnajafian-nv (approved):
LGTM, thanks

PR #1009, opened 2026-09-08
feat(observability): add endpoint session filters

bbednarski9:
## Summary

Adds Relay-native, endpoint-attached OTLP session filtering. A protected trace or log destination can block the matching tool event and all later events for that `session_id`, without disabling unrelated sessions or unprotected destinations.

This feature is intentionally independent from Relay's public conditional-middleware guardrail API: guardrails operate on runtime registrations, while `session_filter` makes a destination-specific decision for each trace/log event.

## Delivery model

```mermaid
flowchart LR
    Events[Relay events] --> Fanout[OpenTelemetry endpoint fan-out]
    Fanout --> Protected[Protected trace/log endpoint]
    Fanout --> Debug[Unprotected debug endpoint]
    Fanout --> Metrics[Metric endpoint]

    Protected --> Resolve[Resolve session_id\nmetadata then scope ancestry]
    Resolve --> Match{Matching tool scope?}
    Match -->|yes| Block[Record blocked session\nand drop current event]
    Match -->|no| Session{Session blocked?}
    Block --> Future[Drop later trace/log events\nfor that session]
    Session -->|yes| Future
    Session -->|no| Export[Export to protected destination]

    Debug --> DebugExport[Export full event stream]
    Metrics --> MetricExport[Export metrics unchanged]
```

After the first tool match at a protected endpoint, events that cannot be associated with a session are also dropped there (`block_after_match`). This is intentionally fail-closed. Blocked-session state lives for the Relay process lifetime; already exported telemetry is not retractable.

## Configuration

```toml
[[components.config.opentelemetry.endpoints]]
endpoint = "https://observability.example/v1/traces"
transport = "http_binary"
type = "gen_ai"

[components.config.opentelemetry.endpoints.session_filter]
type = "block_after_tool_match"
session_metadata_key = "session_id"
tool_name_patterns = [
  "(?i)(?:^|[^a-z0-9])(?:e-?mail|mail|gmail|outlook|inbox|imap|smtp|mailbox)(?:$|[^a-z0-9])",
]
unattributed_events = "block_after_match"
```

- Trace endpoints apply the policy to trace delivery and derived log delivery.
- Explicit protected log endpoints can declare the same `session_filter`.
- Metrics are intentionally not filtered, including measurements produced by a blocked session.
- Endpoints without the policy continue receiving the complete stream.

## Implementation

- Extend trace and log endpoint configuration with `session_filter`.
- Validate nonblank session metadata keys, nonempty pattern lists, and regex syntax.
- Reject `session_filter` on metric endpoints.
- Maintain independent endpoint-local blocked-session and active-scope ancestry state.
- Apply the filter immediately before each trace/log callback in the existing OTLP fan-out path.

## Verification

- `cargo test -p nemo-relay --lib --no-fail-fast` — 1577 passing tests.
- `cargo check -p nemo-relay --features schema`
- Added coverage for metadata identity, scope-ancestry identity, current/future session blocking, fail-closed unattributed events, and protected-versus-unprotected endpoint delivery.

## Companion integration

A separate downstream integration can opt selected endpoint configurations into this policy and retain any field sanitizers as defense in depth.

## Summary by CodeRabbit

- **New Features**
  - Added configurable OpenTelemetry session filtering for trace and explicit log endpoints.
  - Sessions can be blocked after matching tool names, with configurable session metadata and handling for unattributed events.
  - Filtering is supported across available API clients and derived log endpoints.
  - Added diagnostics distinguishing filtered events from queue drops.

- **Bug Fixes**
  - Filtered events are excluded from delivery, including events queued before a matching tool event.
  - Metric endpoints reject unsupported session-filter configurations.

- **Documentation**
  - Documented configuration, behavior, persistence, and endpoint support.


crates/core/src/observability/plugin_component.rs
```diff
@@ -4191,6 +4420,42 @@ fn validate_opentelemetry_signal_endpoint_values(
...
+    if filter.tool_name_patterns.is_empty() {
+        invalid.push("tool_name_patterns must contain at least one pattern".to_string());
+    }
+    for (index, pattern) in filter.tool_name_patterns.iter().enumerate() {
+        if let Err(error) = Regex::new(pattern) {
+            invalid.push(format!(
+                "tool_name_patterns[{index}] is not a valid regex: {error}"
+            ));
+        }
+    }
```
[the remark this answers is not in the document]
bbednarski9: f80ef4a5

bbednarski9:
@coderabbit review

crates/core/src/observability/otel_logs.rs
```diff
@@ -485,6 +499,18 @@ impl LogDeliveryDiagnostics {
...
     }
 
+    fn record_policy_filtered(&self, count: u64) {
+        self.policy_filtered.fetch_add(count, Ordering::Relaxed);
+        self.runtime_diagnostics.record(
+            "otel.logs_session_filtered",
+            format!(
+                "OpenTelemetry session policy filtered {count} logs before export to endpoint {}",
+                self.endpoint
+            ),
```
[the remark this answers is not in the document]
mnajafian-nv: +1. I reproduced this on head `36c72dd2` with URL userinfo and a query token; both appeared in `otel.logs_session_filtered`. Reusing the sanitized endpoint identity here would keep credentials out of diagnostics.

crates/core/src/observability/otel_session_filter.rs
```diff
@@ -0,0 +1,133 @@
...
+                state.scope_sessions.remove(&event.uuid());
+            }
+            None => {}
+        }
+    }
+
+    /// Whether an observed Relay event must not emit a log record.
+    pub(crate) fn blocks_event(&self, event: &Event) -> bool {
+        let state = self.state.lock().unwrap_or_else(|error| error.into_inner());
+        if state.blocked_trace_ids.contains(&event_trace_id(event)) {
```
mnajafian-nv, line 91: **[P1]** I reproduced this on head `36c72dd2` with two string session IDs under the same propagation root. Once session A matches, session B is also blocked because the trace ID check runs before session resolution. That conflicts with the PR’s contract that unrelated sessions remain unaffected. Please preserve session identity through filtering instead of promoting the decision to the entire propagation-root trace, or enforce and document one session per propagation root.

crates/core/src/observability/otel_session_filter.rs
```diff
@@ -0,0 +1,133 @@
...
+
+/// Compiled policy and mutable state for one OpenTelemetry destination.
+#[derive(Debug)]
+pub(crate) struct EndpointSessionFilter {
+    session_metadata_key: String,
+    tool_name_patterns: Vec<Regex>,
+    state: Mutex<EndpointSessionFilterState>,
+}
+
+#[derive(Debug, Default)]
```
mnajafian-nv, line 26: **[P3 follow-up]** The documented process-lifetime retention is sufficient for this PR. For long-running processes with high session churn, this state can still grow continuously. A follow-up should define lifecycle-based cleanup or another bounded design that preserves fail-closed behavior, since simple eviction could allow blocked sessions to resume exporting.

crates/core/src/observability/otel_session_filter.rs
```diff
@@ -0,0 +1,133 @@
...
+            })
+            .or_else(|| state.scope_sessions.get(&event.uuid()).cloned())
+    }
+}
+
+fn event_trace_id(event: &Event) -> TraceId {
+    relay_trace_id(event.propagation_root_uuid().unwrap_or(event.uuid()))
+}
+
+fn json_session_value(value: &Json) -> Option<String> {
```
mnajafian-nv, line 128: **[P2, nonblocking]** I reproduced this on head `36c72dd2`: after `session_id: 1` matches, an unrelated event with `session_id: 2` is also treated as unattributed and blocked. Since `session_metadata_key` is configurable and the contract does not restrict its value to strings, please either normalize scalar identifiers or document and validate a nonblank-string-only contract.

PR #1013, opened 2026-09-09
enhancement: expose tool_call_id to tool execution intercepts

### LINKED ISSUE #1006: [Bug]: Tool execution intercepts cannot access the managed tool_call_id
## Affected area

- Rust core runtime
- Language bindings
- Middleware
- Plugins

## Current behavior

Managed tool execution accepts an optional `tool_call_id`. Relay preserves this
value on the managed tool handle and lifecycle events, but does not pass it to
tool execution intercepts.

The current execution-intercept callback receives only:

```text
tool_name, arguments, next_call
```

The runtime already has the `tool_call_id` when it constructs the execution
chain, but the value is omitted from the callback invocation.

As a result, a plugin cannot inspect or preserve the originating tool-call
identity when it returns a result without invoking `next_call`. The value cannot
be recovered through the existing callback arguments.

Relevant source:

- `crates/core/src/api/tool.rs`
- `crates/core/src/api/runtime/state.rs`
- `crates/core/src/api/runtime/callbacks.rs`

## Expected behavior

Tool execution intercepts should receive the optional `tool_call_id` supplied
to managed execution.

## Impact

Plugins do not have access to the complete identity of the tool execution they
are intercepting.

This prevents a plugin from reliably associating a result with the originating
tool call when it completes the execution itself instead of invoking the
remaining execution chain.

There is no equivalent value available through the existing callback arguments,
and invoking `next_call` is not a valid way to recover information that should
already be part of the interception context.

## API design options

Both viable approaches change the existing tool execution-intercept callback
contract and are breaking API changes.

### Option A: Introduce `ToolExecutionContext`

Replace:

```python
async def intercept(tool_name, arguments, next_call):
    ...
```

with:

```python
async def intercept(
    context: ToolExecutionContext,
    next_call,
) -> ToolExecutionInterceptOutcome:
    ...
```

The initial context contains:

- `tool_name`
- `arguments`
- optional `tool_call_id`

This requires existing callbacks to migrate, but future optional context fields
can be added without changing the callback signature again.

### Option B: Add `tool_call_id` as a callback parameter

Replace:

```python
async def intercept(tool_name, arguments, next_call):
    ...
```

with:

```python
async def intercept(
    tool_name,
    arguments,
    tool_call_id,
    next_call,
) -> ToolExecutionInterceptOutcome:
    ...
```

This is the smaller change, but another required execution value would require
another callback-signature migration.

## Recommended direction

Prefer Option A because Relay maintains this callback contract across multiple
language bindings and plugin transports. A typed execution context requires a
larger initial migration but establishes a stable interface for future optional
execution data.

Whichever option is selected:

- Document the breaking callback migration.
- Update primary bindings together.
- Version native-plugin ABI changes instead of modifying an existing ABI table
  in place.
- Update worker protocol and SDK compatibility requirements.
- Reject incompatible plugin/runtime combinations with a clear version error.

## Acceptance criteria

- An execution intercept receives the exact `tool_call_id` supplied to managed
  tool execution.
- An omitted `tool_call_id` remains absent.
- An intercept can return a result without invoking `next_call` while retaining
  the originating tool-call identity.
- The selected callback contract is implemented consistently across Rust,
  Python, and Node.js.
- Experimental and dynamic-plugin surfaces preserve their documented
  compatibility boundaries.
- Native and worker plugin boundaries reject incompatible versions safely.
- Documentation explains the selected API and migration path.

yuvic-glitch:
#### Overview

Tool execution intercepts received `(tool_name, arguments, next)` and could not observe the provider-issued `tool_call_id`, even though the runtime already records it on the managed tool handle and on lifecycle events. A plugin that completes execution itself, without invoking the remaining chain, had no way to correlate its result with the originating tool call.

This implements Option A from #1006: a structured `ToolExecutionContext` carrying `tool_name`, `arguments`, and `tool_call_id`, delivered to a two-parameter intercept callback. A struct rather than a bare extra parameter lets future optional fields land without another signature change.

**The change is additive.** Every existing registration function keeps the legacy callback shape untouched, and each gains a `_v2` sibling taking the context shape. This was required rather than merely convenient: Node's legacy callback is `(args, next)` — the same arity as `(context, next)` — so runtime arity detection cannot disambiguate the two. Both shapes share one registry, so they order together by `priority` and the existing deregistration removes either. No `_v2` deregistration exists.

#### Details

**Core (`crates/core`)**

- Added `ToolExecutionContext` (private fields + accessors, manual `Debug` with `finish_non_exhaustive`, mirroring `LlmSanitizeRequestContext`) and `ToolExecutionContextFn` in `api/runtime/callbacks.rs`, plus `tool_execution_fn_with_context` to adapt a legacy `ToolExecutionFn`.
- Registries now store a single callable shape — the context form. The legacy `register_tool_execution_intercept` / `scope_register_tool_execution_intercept` are hand-written adapters over the `_v2` macro-generated functions. This keeps the chain builder single-shape and preserves cross-shape priority ordering in one sorted registry.
- `tool_build_execution_chain` takes a context template instead of `name: &str`; each link rebuilds the context with the arguments entering *that* link, so `context.arguments()` matches legacy per-hop behavior exactly.
- `tool_call_execute` reads `handle.tool_call_id` (already in scope at the chain-build site — no threading through intermediate layers was needed).

**Bindings** — `_v2` / `V2` registration added for global, scope-local, and plugin-context surfaces in Python (PyO3 + pure-Python wrappers + all three stub files), Node.js (napi, regenerated `index.d.ts`/`index.js`, hand-written `plugin.d.ts`), Go (cgo trampoline + typedefs), and the C FFI (`nemo_relay.h` regenerated via cbindgen).

Note that neither Node.js nor the C FFI ever passed the tool name to the argument-only shape, so `_v2` is a strict capability gain on both.

**gRPC worker** — additive `optional string tool_call_id = 3` on `ToolInvocation`. Old workers ignore it, old hosts never set it, so no `grpc-v1` version change. Both the Rust and Python worker SDKs gained `register_tool_execution_intercept_v2`.

**Native plugins** — the async/typed path carries the id as an additive JSON envelope field requiring **no ABI change**. The raw `NemoRelayNativeToolExecutionCb` still cannot see it; that needs an ABI 4→5 bump with a frozen v4 table, which would force every native plugin to rebuild. Deliberately left for follow-up and documented as a known limitation in the native ABI reference.

**Docs** — `docs/reference/tool-execution-intercept-outcomes.mdx` gains a "Tool Execution Context" section (canonical field list + per-binding registration table); plugin-context, worker middleware, `grpc-v1` protocol, native ABI reference, and the middleware guide are updated.

`crates/core/src/api/runtime/callbacks.rs` for the context type and the legacy adapter, then `crates/core/src/api/runtime/state.rs` (`tool_build_execution_chain`) for the per-hop context construction — that is the one place where the argument semantics could have drifted from the legacy shape.

The key design decision to scrutinize is **the single-registry-plus-adapter approach**: legacy callbacks are wrapped into the context form at registration time rather than stored as a second variant. This is what makes mixed-shape priority ordering work with one deregistration, and it is verified by `test_execution_intercept_context_exposes_tool_call_id` in `crates/core/tests/integration/middleware_tests.rs`, which registers a legacy intercept at priority 1 and a context intercept at priority 2 and asserts they interleave correctly.

**Non-breakage is demonstrated by omission:** every pre-existing tool-intercept test across all bindings is byte-identical in this PR and still passes. The only pre-existing test modified is `crates/node/tests/llm_tests.mjs`, where a declaration-count assertion legitimately moves from 2 to 4 because two new declarations exist.

#### Validation

All run locally and passing:

- `just test-rust`, `just test-python` (692 + 77 + 24), `just test-node` (394), `just test-go`, `just test-python-plugin` (144)
- `just test-plugin-examples` (26), `just docs` (0 errors), `just check-python-worker-proto`
- `uv run pre-commit run` — fully green, including `go fmt`, `go vet`, `cargo clippy`, and `ffi header sync`

Not run: `just test-all` and the opt-in e2e suites, which are unaffected by this change.

One pre-existing, unrelated failure: `crates/node/tests/dynamic_plugin_tests.mjs` fails to parse on Node 22.23.1 (`await using` syntax). Verified identical on a clean baseline with these changes stashed; the 394 Node passes exclude that file.

- Closes #1006

## Summary by CodeRabbit

- **New Features**
  - Added context-aware tool execution interceptors across Rust, Python, Node.js, Go, C, and worker APIs.
  - Callbacks can access the tool name, arguments, and optional provider `tool_call_id`.
  - Added global, plugin-level, and scope-local registration options.
  - Tool-call correlation IDs are preserved through worker and middleware execution.

- **Compatibility**
  - Existing interceptors remain supported and share priority ordering and deregistration behavior with the new callbacks.

- **Documentation**
  - Added usage guidance, API details, examples, and compatibility notes for context-aware interception.


willkill07:
> **The change is additive.** Every existing registration function keeps the legacy callback shape untouched, and each gains a _v2 sibling taking the context shape. This was required rather than merely convenient: Node's legacy callback is (args, next) — the same arity as (context, next) — so runtime arity detection cannot disambiguate the two. Both shapes share one registry, so they order together by priority and the existing deregistration removes either. No _v2 deregistration exists.

I would rather this be a breaking change. We are not yet in API stability and this change is necessary for stabilization.

willkill07:
@yuvic-glitch can you please update your PR to make it a full breaking change rather than additive?

bbednarski9:
@yuvic-glitch  please drop `go/nemo_relay/nemo-relay-events-2026-09-09-13.54.34.jsonl` from tracked files

crates/core/src/api/registry.rs
```diff
@@ -815,18 +816,47 @@ global_intercept_registry_api!(
...
-    /// Register a global tool execution intercept.
+    /// Register a global tool execution intercept that receives the full
+    /// [`ToolExecutionContext`], including the managed `tool_call_id`.
     /// Execution intercepts can wrap or replace the tool callback. Each
     /// callback returns a canonical tool execution outcome, while its
     /// continuation resolves to the downstream
     /// [`ToolExecutionResult`](crate::api::tool::ToolExecutionResult).
-    register_tool_execution_intercept,
-    /// Deregister a global tool execution intercept.
+    register_tool_execution_intercept_v2,
```
ericevans-nv, line 825: Please make this a breaking change by replacing the existing callback shape with the new `ToolExecutionContext` shape currently implemented through the `_v2` APIs. The `_v2` implementations should use the existing registration names: `register_tool_execution_intercept` in Rust, `register_tool_execution` in high-level Python, `registerToolExecutionIntercept` in Node.js, and `RegisterToolExecutionIntercept` in Go.

Please apply this replacement consistently across the scope and plugin APIs, C FFI, PyO3, Rust and Python worker SDKs, generated declarations, tests, and documentation. Remove the previous argument-only implementations and adapters while keeping the existing registration and deregistration names. Document the breaking callback migration to `(context, next)`. Keep `ToolExecutionContext`, `grpc-v1`, the protobuf package and field, and the result schemas unchanged.

ericevans-nv:
@willkill07

> I would rather this be a breaking change. We are not yet in API stability and this change is necessary for stabilization.

Agreed. This is implemented as a full breaking replacement in [PR #1029](https://github.com/NVIDIA/NeMo-Relay/pull/1029). The existing registration names now accept `(ToolExecutionContext, next)` across the global, scope-local, plugin, C FFI, Python, Node.js, Go, and worker surfaces. The legacy adapters and parallel `_v2` APIs were removed, and native plugins now use ABI v5 with frozen compatibility tables for earlier ABIs.

ericevans-nv:
@willkill07

> @yuvic-glitch can you please update your PR to make it a full breaking change rather than additive?

Addressed in [PR #1029](https://github.com/NVIDIA/NeMo-Relay/pull/1029). The continuation preserves Yuvic’s original signed-off commit and adds the full breaking-contract migration, compatibility boundaries, affected regression tests, and documentation in a follow-up commit.

ericevans-nv:
@bbednarski9

> @yuvic-glitch please drop `go/nemo_relay/nemo-relay-events-2026-09-09-13.54.34.jsonl` from tracked files

Addressed in [PR #1029](https://github.com/NVIDIA/NeMo-Relay/pull/1029). `go/nemo_relay/nemo-relay-events-2026-09-09-13.54.34.jsonl` has been removed from the tracked files.

ericevans-nv:
This work is continued in [PR #1029](https://github.com/NVIDIA/NeMo-Relay/pull/1029), which preserves the original signed-off commit and incorporates the requested breaking-contract migration and follow-up fixes. Closing this PR in favor of the continuation.

PR #1044, opened 2026-09-11
fix(plugin): defer worker registration until static components activate

bbednarski9:
#### Overview

Fix dynamic gRPC worker activation when a worker needs to discover or gate registrations installed by an earlier static component. In the 0.9 unified plugin host, worker startup performed the gRPC Register RPC before static components initialized. A worker that queries for the OpenTelemetry subscriber during registration therefore received an empty result and failed activation.

This restores the intended lifecycle relationship: validate and prepare the worker before host activation, but invoke its Register RPC only when the worker component activates after the configured static components.

##### Worker lifecycle

- Split worker activation into preparation and registration. Preparation still starts the worker, authenticates it, performs the handshake, and validates its manifest, compatibility, and plugin configuration. It deliberately does not call Register.
- Defer the gRPC Register RPC until the worker adapter is initialized as a configured component. Static components are initialized first, and dynamic components are appended afterwards, so worker registration can reliably discover static subscribers and middleware.
- Cache the registration plan returned by the first successful Register RPC, validate it, and install it only after registration succeeds. Activation errors still roll back host-owned registrations and release the worker so a later activation can succeed.
- Keep the public direct worker-loader API eager. This preserves its existing behavior for callers that intentionally load and register workers outside the dynamic plugin host.

##### Native dynamic plugins

Rust native dynamic plugins already defer their register callback until component initialization. No native-loader behavior changes in this PR. The added regression explicitly verifies that a native plugin can discover the static OpenTelemetry subscriber during registration, protecting the shared lifecycle guarantee across both plugin lanes.

##### Validation

- Reproduced the original worker failure without the fix: the registration callback found zero OpenTelemetry subscribers.
- Added worker coverage for static OpenTelemetry discovery, export gating, rollback after registration failure, and successful reactivation.
- Added native dynamic-plugin coverage for OpenTelemetry discovery during the native register callback.
- Passed the focused worker/core suite: 1,690 tests.
- Passed the native dynamic-plugin integration suite: 38 tests.
- Passed repository pre-commit hooks.
- Manually verified the patched CLI starts the configured Claude worker plugin that previously failed with found 0.

Start with crates/core/src/plugin/dynamic/worker.rs, specifically prepare_worker_plugins and WorkerPluginAdapter::register. The key design choice is to retain early process and protocol validation while moving only the registration RPC behind static component activation. Then review crates/core/src/plugin/dynamic/host.rs for the host ordering, followed by the worker and native integration regressions.

- Relates to: none

## Summary by CodeRabbit

* **Improvements**
  * Worker plugins are now validated and authenticated before component initialization.
  * Worker registration is coordinated during activation, improving support for plugins with multiple components.
  * Observability configuration can be discovered and validated for native and worker plugins.
  * Invalid observability registrations now prevent activation with clear validation errors.
  * Successful observability setup applies the appropriate tracing safeguards during runtime.

* **Tests**
  * Added coverage for worker registration reuse, activation failures, cleanup, and observability behavior.

### DIFF
--- a/crates/core/src/plugin/dynamic/host.rs
+++ b/crates/core/src/plugin/dynamic/host.rs
@@ -29,13 +29,16 @@ use super::{
 };
 
 #[cfg(feature = "worker-grpc")]
-use super::{WorkerPluginActivation, WorkerPluginLoadSpec, load_worker_plugins};
+use super::{WorkerPluginActivation, WorkerPluginLoadSpec, prepare_worker_plugins};
 
 /// Initializes the process-wide static and dynamic plugin host.
 ///
 /// The returned handle must remain alive while any plugin-provided callback can
 /// run. Closing or dropping it unregisters components before unloading dynamic
 /// runtimes.
+/// Workers are authenticated and validated before component initialization.
+/// Their `Register` RPC runs when the worker component activates, after earlier
+/// static components have installed their runtime registrations.
 pub async fn initialize(
     config: PluginConfig,
     additional_plugins_toml: Option<PathBuf>,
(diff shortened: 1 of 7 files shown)


willkill07 (commented):
splitting into stages makes a lot of sense.

Coderabbit feedback looks addressable

willkill07 (approved):

PR #1045, opened 2026-09-11
perf: improve daemon worker and gateway concurrency

willkill07:
#### Overview

Improve daemon, worker, and gateway concurrency by isolating session work and reducing control-plane, request-processing, and streaming overhead.

- [x] I confirm this contribution is my own work, or I have the right to submit it under this project's license.
- [x] I searched existing issues and open pull requests, and this does not duplicate existing work.

#### Details

- Serialize hook and gateway work per session while keeping the shared session directory and authenticated-owner locks out of middleware and subscriber waits.
- Keep daemon ACK and challenge control traffic local, add stable-route read-lock decisions, and preserve recovery transitions.
- Build worker middleware requirements from one registry snapshot, move prepared gateway JSON instead of cloning it, and reduce SSE and ATOF serialization allocations.
- Add regressions for independent-session progress, ownership/recovery correctness, SSE recovery, and existing request and cancellation behavior.

#### Where should the reviewer start?

- `crates/cli/src/sessions/mod.rs` and `crates/cli/src/sessions/routing.rs` for the per-session synchronization and ownership lifecycle.
- `crates/cli/tests/coverage/shared/session_tests.rs` for the delayed-tool-guardrail concurrency regression.

#### Related Issues: (use one of the action keywords Closes / Fixes / Resolves / Relates to)

- Relates to: none

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved session event handling to prevent cross-session blocking and reduce races during shutdown, idle checks, and concurrent activity.
  - Improved authenticated session ownership and cleanup during concurrent or partially failed operations.
  - Improved recovery of worker routes so recovered workers can resume correctly.
  - Corrected control-command handling to avoid unnecessary peer updates and connection notifications.
  - Improved middleware compatibility checks and rejected unsupported raw delivery scenarios.
  - Improved resilience when processing malformed or fragmented streaming responses.

- **Performance**
  - Reduced unnecessary data copying and processing for gateway requests, streaming events, and file-only observability output.

PR #1072, opened 2026-09-15
fix: preserve session aliases on hook cancellation

mnajafian-nv:
#### Overview

Complete the child-session alias cancellation handling deferred from #1059, following the per-session concurrency work introduced in #1045.

Relay now preserves an alias when a hook request is cancelled before its event is applied. If cancellation occurs after successful event application, Relay completes the corresponding alias cleanup before releasing the session gate or allowing shutdown to finish.

This change does not modify public APIs or normal event-routing behavior.

#### Details

- Split alias routing into preparation and commit phases so terminal events do not remove routing state before session application succeeds.
- Retain the session gate and activity marker until committed route cleanup finishes, including after request cancellation.
- Compare cleanup snapshots with current routing state so delayed cleanup cannot remove a replacement alias or pending child route.
- Recheck alias and pending-route state after concurrent ownership and routing waits.
- Add regression coverage for cancellation, replacement routes, concurrent alias removal, and shutdown coordination.
- Validate the change with the full Rust test suite, focused cancellation tests, formatting, Clippy, and Cargo checks.

Start with `AppliedRouteCleanup` and `route_hook_event` in `crates/cli/src/sessions/mod.rs`. Then review `SessionAlignmentState::prepare_route` and `commit_route` in `crates/cli/src/agents/shared/alignment.rs`. The cancellation and shutdown regressions are in `crates/cli/tests/coverage/shared/session_tests.rs`.

- Relates to #1059
- Relates to #1045

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session event routing during concurrent activity, ownership changes, and alias updates.
  * Preserved newly created session routes while older cleanup completes.
  * Improved cleanup of ended sessions and pending child sessions without affecting replacement routes.
  * Fixed routing behavior after cancelled operations, including terminal events and shutdown scenarios.
  * Improved promotion and fallback handling for pending child sessions.

* **Tests**
  * Added coverage for concurrent routing, cancellation, cleanup, promotion, and alias changes.


crates/cli/src/sessions/mod.rs
```diff
@@ -1268,64 +1375,91 @@ impl SessionManager {
...
+                            pending.parent_session_id()
+                        )));
+                    }
+                    Some(_) => {}
+                    None => {
+                        owners.insert(pending.parent_session_id().to_string(), owner.to_string());
+                    }
                 }
             }
+            alignment_state.remove_pending(&session_id);
```
willkill07: Cancellation can still lose a pending child route here. `remove_pending` runs before the route is promoted. If the request is cancelled while waiting for promotion, the pending route is gone and a retry treats the child as a separate session. Please keep it until promotion finishes, or restore it on cancellation.

crates/cli/src/agents/shared/alignment.rs
```diff
@@ -329,19 +334,66 @@ impl SessionAlignmentState {
...
         }
-        if matches!(&event, NormalizedEvent::AgentEnded(_)) {
-            self.clear_for_ended_agent(&session_id);
+        for (child_session_id, ended_alias) in &cleanup.ended_aliases {
+            if self.aliases.get(child_session_id) == Some(ended_alias) {
+                self.aliases.remove(child_session_id);
+            }
+        }
+        for (child_session_id, ended_pending) in &cleanup.ended_pending_subagents {
+            if self.pending_subagents.get(child_session_id) == Some(ended_pending) {
```
willkill07: These pending-route checks use `PartialEq`, so a new entry with the same contents can look like the snapshot. If a child starts again before cleanup, this removes the new route and the parent cannot promote it later. Please track a per-entry token and remove only the entry that was originally prepared.

willkill07 (changes requested):

willkill07 (approved):

Issue #1087, opened 2026-09-16
Dynamic plugin layering rejects identical user and system declarations across CLI and SDK hosts


bbednarski9:
## Affected area

Dynamic plugin configuration discovery and file-layering across CLI and SDK
hosts.

## Observed behavior

Relay resolves user `plugins.toml` followed by `/etc/nemo-relay/plugins.toml`.
If both files declare the same manifest-backed dynamic plugin, configuration
resolution fails before the plugin is initialized.

Using the publicly available [Switchyard Relay plugin](https://github.com/NVIDIA-NeMo/Switchyard/tree/main/crates/switchyard-nemo-relay-plugin),
whose manifest ID is `nvidia.switchyard`, the CLI resolver returns:

```text
duplicate dynamic plugin id 'nvidia.switchyard' in /etc/nemo-relay/plugins.toml across plugins.toml sources
```

The core SDK resolver has the same invariant and reports:

```text
duplicate dynamic plugin id 'nvidia.switchyard' in resolved plugins.toml layers
```

This behavior is present in `0.9.0-rc.2`.

## Minimal reproduction

1. Obtain a packaged Switchyard Relay plugin bundle and keep its
   `relay-plugin.toml` at
   `/opt/nemo-relay/plugins/switchyard/relay-plugin.toml`.
2. Create both `$XDG_CONFIG_HOME/nemo-relay/plugins.toml` and
   `/etc/nemo-relay/plugins.toml` with the same record:

   ```toml
   [[plugins.dynamic]]
   manifest = "/opt/nemo-relay/plugins/switchyard/relay-plugin.toml"
   ```

3. Start any affected CLI surface, for example:

   ```bash
   nemo-relay daemon --bind 127.0.0.1 --port 47632 --pass-through
   ```

The configuration resolver rejects the duplicate ID instead of producing an
effective configuration with one Switchyard plugin.

## Affected surfaces

- Normal CLI gateway startup and `nemo-relay gateway start`.
- Transparent wrapper runs, including `nemo-relay run` and the Claude Code and
  Codex wrappers. Their configuration resolution happens before the child agent
  starts.
- Persistent installed/MCP gateway bootstrap, which resolves the same user and
  system scopes.
- The multi-user daemon broker.
- Rust, Python, Node.js, Go, and C SDK hosts that call the normal
  `initialize(config, path)` API. Default SDK discovery is user -> system; an
  explicit path replaces the user layer but does not remove the system layer.

[...]

PR #1090, opened 2026-09-16
fix: layer dynamic plugin declarations by ID

willkill07:
#### Overview

Resolve RELAY-883 by treating dynamic plugin declarations in layered `plugins.toml` files as keyed by manifest plugin ID. A higher-precedence declaration now replaces the lower declaration in full.

#### Details

- Resolve cross-file dynamic plugin declarations by ID in both CLI and core SDK configuration.
- Preserve same-file duplicate rejection; select the higher manifest, lifecycle scope, and configuration on a cross-file match.
- Add CLI/core regression coverage and document the behavior, migration impact, and release-note entry.

Review `crates/core/src/plugin/dynamic/configuration.rs`, where layered dynamic declarations become one effective record before validation and activation. The CLI resolver mirrors that behavior in `crates/cli/src/configuration/mod.rs`.

- Closes #1087
- Closes RELAY-883: https://linear.app/nvidia/issue/RELAY-883/dynamic-plugin-layering-rejects-identical-user-and-system

## Summary by CodeRabbit

* **Changed**
  * Dynamic plugin declarations with the same manifest ID across configuration layers now use the higher-priority record.
  * The higher-priority record fully replaces the lower-priority plugin manifest, lifecycle settings, source, and configuration.
  * Duplicate plugin IDs within a single configuration file remain invalid.
  * Plugin records are now hydrated in their effective configuration scope.
  * Lifecycle operations now resolve and update plugins within the applicable configuration scope.

* **Documentation**
  * Added guidance on precedence, replacement behavior, required settings, and migration for layered plugin configurations.


bbednarski9 (approved):
Thanks for taking this on. Good consistency with keying by manifest ID but preserving the higher layer's configuration

crates/cli/src/configuration/mod.rs
```diff
@@ -1504,14 +1504,22 @@ where
...
             || resolved_plugins.dynamic_plugin_policy != DynamicPluginHostPolicy::default()
         {
             contributing_sources.push(path.clone());
         }
-        dynamic_plugins.extend(resolved_plugins.dynamic_plugins);
+        for plugin in resolved_plugins.dynamic_plugins {
+            if let Some(index) = dynamic_plugin_indices.get(&plugin.plugin_id) {
+                // Dynamic plugin declarations layer by manifest ID. The later source owns the
+                // effective manifest, lifecycle state, and opaque configuration.
+                dynamic_plugins[*index] = plugin;
```
mnajafian-nv, line 1517: I reproduced this with an existing user-scoped record and a same-ID system declaration. Configuration resolution selects the system record, but lifecycle hydration still finds the user record and then attempts to update the empty system registry, causing startup resolution to fail with `dynamic plugin 'fixture.layered' is not registered`. Please make lifecycle lookup honor the effective source and add regression coverage for this user-to-system override path.

crates/cli/src/plugins/lifecycle/mod.rs
```diff
@@ -2114,7 +2122,11 @@ fn load_and_hydrate_scopes_with_updates(
         let policy =
```
mnajafian-nv, line 2139: This now hydrates the system record correctly, but the existing user record remains live. As a result, lifecycle commands that look up by plugin ID see two records and fail with “configured in multiple lifecycle scopes.” I hit this with `plugins inspect <id>` after a user-to-system override; validate, enable/disable, and remove follow the same lookup path. Could these commands select the effective declaration’s scope instead, with regression coverage for that layered case?

mnajafian-nv (approved):
lgtm upon addressing ^^^

mnajafian-nv (approved):
LGTM!Thanks

mnajafian-nv (commented):
Can we address the three reproducible dynamic-plugin lifecycle test failures on Linux ARM64? The CodeRabbit comments are valid hardening, but they are not blockers.

mnajafian-nv (approved):
LGTM

PR #1099, opened 2026-09-17
enhancement: add CLI hook and upstream operational logging

willkill07:
#### Overview

Add private operational logging for CLI hook delivery and upstream LLM calls.

#### Details

- Add opaque operation and session correlation fields without logging hook or request details.
- Record normal hook and upstream lifecycle events at debug level.
- Record delays, stalls, limits, transport failures, and runtime failures at warn or error level.
- Cover local, daemon, and managed-worker paths with timing and JSONL privacy tests.
- Document event names, thresholds, and redaction rules in plain language.

Start with `crates/cli/src/operational.rs`, then review its use from `crates/cli/src/gateway/mod.rs` and `crates/cli/src/daemon/worker/managed.rs`.

- Relates to: none

#### Validation

- `just test-rust`
- Commit hooks, including `cargo clippy`

## Summary by CodeRabbit

- **New Features**
  - Added privacy-conscious operational logging for hook delivery and upstream requests.
  - Logs capture lifecycle events, failures, response statuses, payload limits, cancellations, and streaming delays.
  - Added operation and optional session correlation while filtering sensitive request and credential data.
  - Streaming monitoring detects delayed headers, delayed first events, and stalled streams without cancelling requests.

- **Documentation**
  - Added reference documentation covering operational log fields, thresholds, lifecycle events, and privacy protections.

- **Bug Fixes**
  - Internal operation identifiers are no longer forwarded to upstream providers.


crates/cli/src/operational.rs
```diff
@@ -0,0 +1,367 @@
...
+            .unwrap_or_else(|| Uuid::now_v7().to_string());
+        Self {
+            operation_id,
+            session_id: None,
+            started_at: Instant::now(),
+        }
+    }
+
+    pub(crate) fn with_session(mut self, session_id: impl Into<String>) -> Self {
+        self.session_id = Some(session_id.into());
```
bbednarski9: **Session correlation values are not necessarily opaque.**

`with_session` stores the supplied string verbatim. The hook adapters accept session IDs directly from payloads or headers (`agent_session_id` in `crates/cli/src/agents/shared/adapters.rs`), and the new hook handlers attach those IDs and emit `hook_started` before session validation. The log formatter also preserves field values verbatim. Consequently, private text supplied as a session identifier can be written unchanged to operational logs, including failure records, despite the intended privacy-preserving correlation contract.

Could we derive a stable, privacy-safe correlation token from the native session identifier instead of logging the raw value? Please also extend the JSONL privacy coverage through a real hook handler with sensitive sentinels in both the payload session ID and the supported session headers. The current privacy test exercises the forwarder against a mock server, so it does not cover these handler-side session fields.

crates/cli/src/daemon/hook/mod.rs
```diff
@@ -117,11 +149,15 @@ async fn forward(
         .post(endpoint)
         .header(CONTENT_TYPE, "application/json")
         .header(CLIENT_TOKEN_HEADER, token)
+        .header(
+            crate::operational::OPERATION_ID_HEADER,
+            operational.operation_id(),
+        )
```
bbednarski9, line 155: **The daemon broker strips this correlation header before it reaches the worker.**

`public_proxy_inner` calls `strip_public_relay_headers` in `crates/cli/src/daemon/broker/server.rs` before forwarding the request. That helper removes all `x-nemo-relay-*` headers except the client token and, for provider routes, the named-upstream header. It therefore removes `x-nemo-relay-operation-id` from this hook request. The managed worker's `OperationalContext::take_from_headers` then generates a new ID, so the forwarder and handler records cannot be correlated using the promised shared operation ID.

Could we preserve a validated operation ID specifically for authenticated hook forwarding through the broker, while continuing to strip caller-supplied correlation headers from provider requests? An end-to-end forwarder → broker → managed-worker test asserting that both hook boundaries log the same operation ID would cover the missing hop.

bbednarski9 (approved):

PR #1110, opened 2026-09-17
fix(cli): preserve Linux venv lib64 aliases

afourniernv:
#### Overview

Linux virtual environments normally contain a root `lib64 -> lib` symlink. Dynamic Python plugin attestation currently traverses that alias as a second copy and can reject a physically sub-512 MiB environment as oversized. Activation snapshots also expand the alias instead of preserving it.

#### Details

- Charge file bytes once for the exact top-level `lib64 -> lib` venv alias while continuing to hash both logical paths.
- Keep all entry/depth limits and normal byte charging for every other symlink.
- Preserve the standard alias in Unix activation snapshots when `pyvenv.cfg` identifies the venv root.
- Add regression coverage for budget enforcement, digest compatibility, snapshot verification, and tamper detection.

Start with `python_environment_byte_budget_counts_internal_directory_alias_once` and `python_activation_snapshot_preserves_internal_directory_alias` in `crates/cli/tests/coverage/shared/plugins_lifecycle_tests.rs`.

Validation:

- 92 plugin lifecycle tests passed

#### Related Issues

- Relates to NVIDIA/NeMo-Relay-Plugins#6


afourniernv:
The underlying issue is duplicate accounting of the standard Linux virtual-environment layout:

```text
lib/
lib64 -> lib
```

Relay currently follows both paths during attestation, so the same installed files are charged twice against the 512 MiB environment limit. Snapshot creation also expands the alias into a second physical copy instead of preserving the symlink.

This change charges those bytes once and preserves the standard alias in the activation snapshot. It still hashes both logical paths for digest compatibility and tamper detection, and it leaves arbitrary symlinks and the existing physical limits unchanged.

The Guardrails dependency pruning may mean that integration no longer strictly requires this fix, but the Relay bug remains relevant for future larger Linux Python workers.

crates/cli/src/plugins/lifecycle/environment.rs
```diff
@@ -350,18 +350,33 @@ pub(super) fn environment_tree_digest(environment: &Path) -> Result<String, Stri
...
     max_entries: usize,
+) -> Result<String, String> {
+    environment_tree_digest_with_limits(
+        environment,
+        max_entries,
+        crate::filesystem::bounded::MAX_BOUNDED_FILE_BYTES,
+    )
+}
+
+fn environment_tree_digest_with_limits(
```
willkill07: The name of this function is way too similar to `environment_tree_digest_with_limit`

crates/cli/src/plugins/lifecycle/environment.rs
```diff
@@ -393,9 +424,8 @@ fn digest_environment_directory(
     relative_directory: &Path,
     ancestors: &mut Vec<PathBuf>,
     digest: &mut Sha256,
-    total: &mut u64,
-    entries: &mut usize,
-    max_entries: usize,
+    budget: &mut EnvironmentDigestBudget,
```
willkill07: Just a thought... the new struct could be `EnvironmentDigest` which combines the Sha with the data. This would probably clean things up a bit :)

crates/cli/src/plugins/lifecycle/environment.rs
```diff
@@ -482,17 +513,33 @@ fn digest_environment_entry(
...
-            crate::filesystem::bounded::MAX_BOUNDED_FILE_BYTES
-        ));
+    if charge_bytes {
+        budget.total_bytes = budget.total_bytes.saturating_add(bytes.len() as u64);
+        if budget.total_bytes > budget.max_bytes {
+            return Err(format!(
+                "managed Python environment exceeds the {}-byte attestation budget",
+                budget.max_bytes
+            ));
+        }
```
willkill07: suggestion: have an `impl` for `EnvironmentDigest`/`EnvironmentDigestBudget` that does the charge that returns `Result<(), Error>` so we can simplify the logic here to:
```rust
if charge_bytes {
    budget.charge(bytes.len())?
}
```

Additionally, `update_tree_digest` could also be an `impl` function for `EnvironmentDigest`

willkill07 (changes requested):

crates/cli/src/plugins/lifecycle/environment.rs
afourniernv: Addressed in fab370c7: the private helpers are now environment_tree_digest_with_entry_limit and environment_tree_digest_with_budget, so their purposes are explicit rather than distinguished by singular/plural naming.

crates/cli/src/plugins/lifecycle/environment.rs
afourniernv: Done in fab370c7. EnvironmentDigest now owns the SHA-256 state, entry and byte counters, and their limits; the recursive traversal passes one state object.

crates/cli/src/plugins/lifecycle/environment.rs
afourniernv: Done in fab370c7. Charging and digest updates are now EnvironmentDigest methods (charge_entry, charge_bytes, and update), which removes the inline budget bookkeeping while preserving separate hashing and byte charging for the lib64-to-lib alias.

crates/cli/src/plugins/lifecycle/environment.rs
```diff
@@ -482,17 +513,33 @@ fn digest_environment_entry(
...
     }
     update_tree_digest(digest, b'f', &relative, &bytes);
     Ok(())
 }
 
+fn is_python_lib64_alias(path: &Path, relative: &Path) -> Result<bool, String> {
+    if relative != Path::new("lib64") {
```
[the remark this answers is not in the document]
afourniernv: Confirmed and fixed in 2fcdf693. Digest byte charging now requires the same pyvenv.cfg marker as snapshot alias preservation, so a non-venv lib64-to-lib symlink remains fully charged.

crates/cli/tests/coverage/shared/plugins_lifecycle_tests.rs
```diff
@@ -1914,6 +1914,93 @@ fn python_environment_entry_budget_counts_skipped_cache_entries() {
...
+    .expect("lib64 -> lib must not charge installed files twice");
+
+    std::fs::write(&installed, b"changed package!").unwrap();
+    let changed = environment::test_environment_tree_digest_with_limits(
+        &environment_path,
+        16,
+        payload.len() as u64,
+    )
+    .unwrap();
+    assert_ne!(original, changed, "aliased content must remain attested");
```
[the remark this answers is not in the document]
afourniernv: Added in 2fcdf693. The regression now compares the aliased venv digest with an equivalent physical lib and lib64 tree, while retaining the mutation and byte-budget assertions.

crates/cli/src/plugins/lifecycle/environment.rs
```diff
@@ -482,17 +532,31 @@ fn digest_environment_entry(
...
+    if charge_bytes {
+        digest.charge_bytes(bytes.len())?;
     }
-    update_tree_digest(digest, b'f', &relative, &bytes);
+    digest.update(b'f', &relative, &bytes);
     Ok(())
 }
 
+fn is_python_lib64_alias(path: &Path, relative: &Path) -> Result<bool, String> {
+    if relative != Path::new("lib64")
```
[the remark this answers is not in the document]
afourniernv: Confirmed and fixed in 0b642da5. The byte exemption is now Unix-only, matching the platform where snapshotting preserves the venv alias; non-Unix hosts continue charging the expanded tree.

willkill07 (approved):

PR #1118, opened 2026-09-18
feat: add OTLP environment autoconfiguration

willkill07:
#### Overview

Adds automatic OTLP exporter creation from OpenTelemetry environment configuration.

#### Details

- Delegate OTLP endpoint, transport, timeout, compression, headers, and resource handling to the Rust OpenTelemetry SDK where applicable.
- Preserve Relay endpoint isolation and add strict automatic-export activation and opt-out behavior.
- Document supported OTEL variables in a sticky searchable table and add regression coverage.

Start with `crates/core/src/observability/plugin_component.rs` for automatic activation and `docs/configure-plugins/observability/opentelemetry.mdx` for the public contract.

- Closes RELAY-888

## Summary by CodeRabbit

- **New Features**
  - Added automatic OpenTelemetry setup through standard OTLP environment variables for traces, logs, and metrics.
  - Supports signal-specific endpoints, headers, HTTP/gRPC transports, compression, and SDK resource detection.
  - Honors per-signal settings and SDK-wide telemetry disablement.
  - Explicit configuration retains transport, batching, resource, and timing controls.
  - Prevents redirects for Relay-configured OTLP/HTTP exporters.

- **Documentation**
  - Documented automatic configuration, precedence, defaults, and limitations.
  - Updated observability configuration requirements to version 4; version 3 configurations are no longer supported.


Salonijain27 (approved):

ericevans-nv (approved):

crates/core/src/observability/plugin_component.rs
```diff
@@ -1196,36 +1196,51 @@ fn register_observability(
     config: ObservabilityConfig,
     ctx: &mut PluginRegistrationContext,
 ) -> PluginResult<()> {
-    if !matches!(config.version, 3 | 4) {
+    if config.version != 4 {
```
mnajafian-nv, line 1199: This changes the observability version gate from accepting v3 and v4 to rejecting v3 outright. That breaks existing consumers that do not use the new automatic OTLP behavior, including the OpenClaw live smoke and Rust/FFI/CLI coverage, which now fail with `observability config version 3 is unsupported; use version 4`. Could we preserve the existing v3 trace-only contract and require v4 only when logs or metrics are configured? That keeps this feature additive and avoids a breaking migration for unrelated ATOF/ATIF and trace-only users.

willkill07: We are intentionally dropping v3 in 0.10 in preparation for stability/unification.

mnajafian-nv (approved):

